Issue #6, opened 2021-09-22
Settings


joaopat98:
- [ ] Parse Settings File
- [x] Retrieve Settings on request with required defaults
- [x] Update Settings values

RicardoEPRodrigues:
Hello, I just wanted to share some thoughts.

What are these settings for exactly? Perhaps a concrete example might help reduce implementation efforts.

Is it for engine-specific properties? Or can it be a generic system for engine plus game? Should the settings be split?

Hope these questions help define the issue better.

ratuspro:
I think we can close this issue without the parsing from file.
We should assume that something external to the Settings singleton will populate it.
Potentially calling `setString` will be enought right now

PR #33, opened 2021-09-30
Bootstrap settings

ratuspro:
Add the interface and the implementation to start solving #6.

Currently, it is not using templates for the getters and setters.
It just accepts strings. This should change in the future.

Also, it already has a couple of unit tests to test the basic functionality.
Bear in mind that using Singletons will make testing a bit hard. 
We should discuss if it is not preferable to use dependency injection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ option(BUILD_SAMPLES "Build samples" OFF)
 set(CUBOS_SOURCE
     "src/entry.cpp"
     "src/log.cpp"
-    
+
     "src/io/glfw_window.cpp"
 
     "src/gl/ogl_render_device.cpp"
@@ -26,6 +26,8 @@ set(CUBOS_SOURCE
 set(CUBOS_INCLUDE
     "include/cubos/log.hpp"
 
+    "include/cubos/settings.hpp"
+    
     "include/cubos/io/window.hpp"
 
     "include/cubos/gl/render_device.hpp"
@@ -99,7 +101,7 @@ include(CPack)
 
 # Enable Google Test
 set(CUBOS_TESTS_SOURCE
-    "tests/test_example.cpp"
+    "tests/test_settings.cpp"
 )
 
 add_executable(
@@ -109,6 +111,7 @@ add_executable(
 
 target_link_libraries(
     cubos_tests
+    cubos
     gtest_main
 )
 
--- a/include/cubos/settings.hpp
+++ b/include/cubos/settings.hpp
@@ -1,8 +1,8 @@
 #ifndef CUBOS_SETTINGS_HPP
 #define CUBOS_SETTINGS_HPP
+#include <memory>
 #include <string>
-#include <map>
-#include <any>
+#include <unordered_map>
 
 namespace cubos
 {
@@ -10,7 +10,9 @@ namespace cubos
     class Settings
     {
     protected:
-        Settings();
+        Settings()
+        {
+        }
         static Settings* _instance;
 
     public:
@@ -18,14 +20,18 @@ namespace cubos
         void operator=(const Settings&) = delete;
         static Settings* GetInstance();
 
-        template <typename T> void set(const std::string& key, T value);
-        template <typename T> T get(const std::string& key, T defaultValue);
+        void clear();
+
+        void setString(const std::string& key, const std::string& value);
+        std::string getString(const std::string& key, const std::string& defaultValue);
 
     private:
-        std::map<std::string, std::any> _values;
+        std::unordered_map<std::string, std::string> _values;
     };
 
 } // namespace cubos
+
+cubos::Settings* cubos::Settings::_instance = nullptr;
 
 cubos::Settings* cubos::Settings::GetInstance()
 {
@@ -36,4 +42,22 @@ cubos::Settings* cubos::Settings::GetInstance()
     return _instance;
 }
 
-#endif
+void cubos::Settings::clear()
+{
+    _values.clear();
+}
+
+void cubos::Settings::setString(const std::string& key, const std::string& value)
+{
+    _values[key] = value;
+}
+
+std::string cubos::Settings::getString(const std::string& key, const std::string& defaultValue)
+{
+    if (_values.count(key) == 0)
+        return defaultValue;
+
+    return _values[key];
+}
+
+#endif // CUBOS_SETTINGS_HPP
--- a/include/cubos/settings.hpp
+++ b/include/cubos/settings.hpp
@@ -1,0 +1,39 @@
+#ifndef CUBOS_SETTINGS_HPP
+#define CUBOS_SETTINGS_HPP
+#include <string>
+#include <map>
+#include <any>
+
+namespace cubos
+{
+
+    class Settings
+    {
+    protected:
+        Settings();
+        static Settings* _instance;
+
+    public:
+        Settings(Settings& settings) = delete;
+        void operator=(const Settings&) = delete;
+        static Settings* GetInstance();
+
+        template <typename T> void set(const std::string& key, T value);
+        template <typename T> T get(const std::string& key, T defaultValue);
+
+    private:
+        std::map<std::string, std::any> _values;
+    };
+
+} // namespace cubos
+
+cubos::Settings* cubos::Settings::GetInstance()
+{
+    if (_instance == nullptr)
+    {
+        _instance = new Settings();
+    }
+    return _instance;
+}
+
+#endif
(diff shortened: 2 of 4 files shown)


include/cubos/settings.hpp
```diff
@@ -0,0 +1,63 @@
...
+        void setString(const std::string& key, const std::string& value);
+        std::string getString(const std::string& key, const std::string& defaultValue);
+
+    private:
+        std::unordered_map<std::string, std::string> _values;
+    };
+
+} // namespace cubos
+
+cubos::Settings* cubos::Settings::_instance = nullptr;
```
RiscadoA, line 34: This will cause a compilation error when the header is included in more than one compilation unit, since there will be a duplicate definition. This can moving be easily solved by moving the definitions to a `src/settings.cpp` file.

include/cubos/settings.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+cubos::Settings* cubos::Settings::GetInstance()
```
RiscadoA, line 36: Read previous comment (same issue)

include/cubos/settings.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+    if (_instance == nullptr)
+    {
+        _instance = new Settings();
+    }
+    return _instance;
+}
+
+void cubos::Settings::clear()
```
RiscadoA, line 45: Read previous comment (same issue)

include/cubos/settings.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+    _values.clear();
+}
+
+void cubos::Settings::setString(const std::string& key, const std::string& value)
```
RiscadoA, line 50: Read previous comment (same issue)

include/cubos/settings.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+    _values[key] = value;
+}
+
+std::string cubos::Settings::getString(const std::string& key, const std::string& defaultValue)
```
RiscadoA, line 55: Read previous comment (same issue)

RiscadoA (changes requested):

RiscadoA:
I also think that we should avoid singletons and instead should the pass the settings class by reference to where it will be needed.

include/cubos/settings.hpp
ratuspro: Added to settings.cpp

include/cubos/settings.hpp
ratuspro: Added to settings.cpp

include/cubos/settings.hpp
ratuspro: Added to settings.cpp

include/cubos/settings.hpp
ratuspro: Added to settings.cpp

RiscadoA:
Doxygen documentation is still missing.

RiscadoA:
I think that instead of `Settings::GetInstance()->setString(...)`, we should use `Settings::setString(...)` and make getInstance a private method/define the instance as a static member of the class.

ratuspro:
@RiscadoA should I add doxygen documentation to all the methods of Settings?

RiscadoA:
> @RiscadoA should I add doxygen documentation to all the methods of Settings?

I think so. The methods documentation should also have some sample code.

RiscadoA:
Okay I hadn't seen the two previous commits. I think its fine this way but you should try to use the same documentation style as we had decided before (it was alre!dy used in render_device.hpp, you can a look there). Also, shouldn't GetSettings be getSettings?

CMakeLists.txt
```diff
@@ -98,7 +100,7 @@ include(CPack)
 
 # Enable Google Test
 set(CUBOS_TESTS_SOURCE
```
RiscadoA, line 102: Maybe move the tests cmake configuration to the tests folder (like the samples) and add it with add_subdirectory)

include/cubos/settings.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <string>
+#include <unordered_map>
+
+namespace cubos
+{
+
+    class Settings
+    {
+    protected:
+        Settings()
```
RiscadoA, line 13: Should be changed to Settings() = default; This way it becomes shorter and there's no need to move it to the source file.

include/cubos/settings.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace cubos
+{
+
+    class Settings
+    {
+    protected:
```
RiscadoA, line 12: Why protected instead of private?

src/settings.cpp
```diff
@@ -0,0 +1,33 @@
+#include <cubos/settings.hpp>
+
+cubos::Settings* cubos::Settings::_instance = nullptr;
```
RiscadoA, line 3: Add `using namespace cubos;` and use auto for variable definitions to make it less verbose.

CMakeLists.txt
ratuspro: Yes, I agree.
The problem is related with #34 and it will be fixed soon.

include/cubos/settings.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <unordered_map>
+
+namespace cubos
+{
+
+    class Settings
+    {
+    private:
+        Settings() = default;
+        static Settings* _instance;
```
RiscadoA, line 14: Why aren't this methods and fields in line 36? Also, are we using underscore in private fields? I wasn't following that convention in the rest of the code, since I personally don't like doing that in C++: names starting with _ in C++ are reserved.

I also think we should use std::unique_ptr<Settings> instead of a raw pointer, since the current code leads to a memory leak.

include/cubos/settings.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+        /// Retrieves the string settings with a specific key.
+        /// If no setting exists with such key, returns the default value.
+        /// @param key the setting's key
+        /// @param defaultValue the value returned when not setting with the provided key exists
+        /// @return std::string the setting's value
+        static std::string getString(const std::string& key, const std::string& defaultValue);
+
+    private:
+        std::unordered_map<std::string, std::string> _values;
```
RiscadoA, line 37: Underscore again; above private constructor and field should be moved here too

src/settings.cpp
```diff
@@ -0,0 +1,35 @@
+#include <cubos/settings.hpp>
+
+using namespace cubos;
+
+Settings* Settings::_instance = nullptr;
```
RiscadoA, line 5: Use std::unique_ptr

RiscadoA (changes requested):

include/cubos/settings.hpp
ratuspro: Regarding the pointer issue,
If we use a unique_ptr, then I would not be able to return it from GetInstance. Or am I missing something?

To avoid memory leaks wouldn't it be enough to define the destructor of Settings?

RicardoEPRodrigues:
> I think that instead of `Settings::GetInstance()->setString(...)`, we should use `Settings::setString(...)` and make getInstance a private method/define the instance as a static member of the class.

Hello, this is a similar implementation to what Unity uses, while it works well I would suggest allowing for a more flexible implementation, providing a structure that holds several key-value pairs would allow less calls to the instance. This suggestion is similar to what Unreal Engine uses (take a look at https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/SaveGame/).

Hope it helps.

src/settings.cpp
```diff
@@ -2,10 +2,10 @@
 
 using namespace cubos;
 
-Settings* Settings::instance = nullptr;
-
 Settings* Settings::getInstance()
 {
+    static Settings* instance = nullptr;
```
RiscadoA, line 7: Why a pointer? This will lead to a memory leak anyway.
Since we're doing it this way we can declare it the normal way and just return a reference to it, like
```cpp
Settings& Settings::getInstance()
{
    static Settings instance;
    return instance;
}

File: include/cubos/settings.hpp [hunk 14]
```diff
@@ -31,7 +31,6 @@ namespace cubos
 
     private:
         Settings() = default;
-        static Settings* instance;
         std::unordered_map<std::string, std::string> values;
         static Settings* getInstance();
```
RiscadoA, line 35: Very minor edit: put private methods together, methods first members last (tldr just put values after getInstance, with a new line between).

Issue #7, opened 2021-09-22
Window/Input Manager


joaopat98:
- [x] Create windows
- [x] Input Bindings
Hardware that is connected to specific actions (WASD, Spacebar, left joystick, etc.)
- [x] Input Actions
Specific actions that result from a player's input (move, jump, etc.)

ratuspro:
Can this be closed @joaopat98 @DannyIsYog?

RiscadoA:
No, we still have no input manager, just window events.

DannyIsYog:
Lambda functions don't work when using them as callback on the register binding functions, let it be given by us or the programmer. Need to find a work around that.

PR #39, opened 2021-10-09
Add events to the Window class

RiscadoA:
Depend on #38
Depend on #50 
Work on #7 

The following public members (`Event<...>` instances) should be added to the `Window` class:

- [x] Framebuffer resized
- [x] Key down
- [x] Key up
- [x] Mouse down
- [x] Mouse up
- [x] Mouse moved
- [x] Mouse wheel scrolled

In the source file, callbacks to GLFW event should be added which fire the events above.


src/io/glfw_window.cpp
```diff
@@ -119,11 +118,28 @@ bool GLFWWindow::shouldClose() const
     return glfwWindowShouldClose(this->handle);
 }
 
-CubosKey glfwToCubosKey(int key)
+MouseButton glfwToCubosMouseButton(int button)
```
RiscadoA, line 121: Add static before return type

src/io/glfw_window.cpp
```diff
@@ -119,11 +118,28 @@ bool GLFWWindow::shouldClose() const
...
+{
+#define MAP_BUTTON(glfw, cubos)                                                                                        \
+    case GLFW_MOUSE_BUTTON_##glfw:                                                                                     \
+        return MouseButton::cubos;
+    switch (button)
+    {
+        MAP_BUTTON(LEFT, Left)
+        MAP_BUTTON(RIGHT, Right)
+        MAP_BUTTON(MIDDLE, Middle)
+        // Not sure what the mapping in GLFW is for the Exta1 and Extra2 buttons
```
RiscadoA, line 131: Extra 1 = GLFW_MOUSE_BUTTON_4
Extra 2 = GLFW_MOUSE_BUTTON_5

src/io/glfw_window.cpp
```diff
@@ -119,11 +118,28 @@ bool GLFWWindow::shouldClose() const
...
+    default:
+        return MouseButton::Invalid;
+    }
+#undef MAP_BUTTON
+}
+
+Keyboard glfwToCubosKey(int key)
```
RiscadoA, line 138: Add static before return type

RiscadoA:
>     * The mouse scoll event  is not only called on mouse wheel scroll but is also called on touchpad scroll, don't know if it has any implications.
I don't think we have to worry about this.

> So should it be implemented with Sticky Mouse Buttons?

This also won't be a problem since that part of the documentation refers to `glfwGetMouseButton`, which we aren't using.

include/cubos/io/keyboard.hpp
```diff
@@ -0,0 +1,108 @@
+namespace cubos::io
```
RiscadoA, line 1: Include guard missing (#ifndef CUBOS_IO_KEYBOARD_HPP).
Maybe also add `/// Keyboard key codes enum` before `enum class Keyboard`.

src/entry.cpp
```diff
@@ -11,8 +12,8 @@ using namespace cubos;
 int main(void)
 {
     initializeLogger();
-    auto window = io::Window();
-    auto& renderDevice = window.getRenderDevice();
+    io::Window* window = new io::GLFWWindow();
```
RiscadoA, line 15: Window must be deleted before returning from main

src/io/glfw_window.cpp
```diff
@@ -1,20 +1,26 @@
-#include <cubos/io/window.hpp>
+#include <cubos/io/glfw_window.hpp>
 
-#ifdef WITH_GLFW
```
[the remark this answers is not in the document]
RiscadoA: This still needs to be solved.

src/entry.cpp
```diff
@@ -153,6 +153,6 @@ int main(void)
 
         t += 0.01f;
     }
-
+    window->~Window();
```
RiscadoA, line 156: Don't call the destructor, this doesn't actually free the memory, and it shouldn't be called directly usually.
Instead use `delete window;` (its like a free() that calls the destructor).

include/cubos/io/window.hpp
```diff
@@ -47,11 +47,11 @@ namespace cubos::io
 
         /// Invoked when a keyboard key is pressed
         /// Invoked with a Keyboard
```
RiscadoA, line 49: Invoked with a Key code*

include/cubos/io/window.hpp
```diff
@@ -47,11 +47,11 @@ namespace cubos::io
...
-        Event<Keyboard> onKeyDown;
+        Event<Key> onKeyDown;
 
         /// Invoked when a keyboard key is released
         /// Invoked with a Keyboard
```
RiscadoA, line 53: Invoked with a Key code*

RiscadoA:
Good work, I think this is ready, we just need to merge #38 and #50 first.

Issue #35, opened 2021-09-30
CMake "include googletest" fails


RiscadoA:
CMake fails with:
```
include could not find load file:

googletest
```
Even if googletest was found (or added as a submodule).

RiscadoA:
Maybe instead of using submodules use [this](https://cmake.org/cmake/help/latest/module/GoogleTest.html)

Issue #34, opened 2021-09-30
Add cmake alternatives for game engine and testing


ratuspro:
Use an option to determine the executable being generated.
By default the engine should be run as the game engine: invoking entry.cpp's main function.
In the other cases, it should run the test executable.

ratuspro:
Solved on #40

PR #40, opened 2021-10-10
Fix google test configuration

ratuspro:
This PR focuses on solving the issues related with generating and running the test executables that rely on googletest #35 and #34.

The cmake source for generating tests was moved to a CMakeLists.txt inside `tests` directory.

The root CMakeLists.txt file defines an option `BUILD_ENGINE_TESTS` to determine if the tests should be build. 
If so, the entry.cpp is removed from the source files - to remove the entry.cpp:main() - and the `tests` directory is added as a subdirectory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,15 +40,15 @@ set(CUBOS_ENTRYPOINT
 )
 
 # Create Cubos Library
-add_library(cubos_lib ${CUBOS_SOURCE} ${CUBOS_INCLUDE})
-target_include_directories(cubos_lib PUBLIC "include")
-set_property(TARGET cubos_lib PROPERTY CXX_STANDARD 20)
+add_library(cubos-core ${CUBOS_SOURCE} ${CUBOS_INCLUDE})
+target_include_directories(cubos-core PUBLIC "include")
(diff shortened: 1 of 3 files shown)


ratuspro:
@RiscadoA can you please test this to see if #35 was fixed?

RiscadoA:
Also, instead of removing the `entry.cpp` from the main `cubos` target, which breaks the samples, we should create an additional `cubos-test` target which doesn't contain the `entry.cpp` file. This way we can run samples and tests at the same time.

ratuspro:
The current build pipeline generates:
- _cubos_lib (library)_: with cubos core components
- _cubos (library)_: with the entrypoint to start the engine and _cubos_lib_
- _samples (executable)_: executable that links _cubos_ and game assets
- _cubos_tests (executable)_: executable to test _cubos_lib_

Options BUILD_SAMPLES and BUILD_ENGINE_TESTS can be used to determine if _samples_ and _cubos_tests_ are generated, respectively.

Commit #4d704177a6c16d4f08a61c3aca3ad14b4fbbcf8c also solves #40
 
Down the road, we should consider restructuring the repo to solve #41 and include a cubos editor executable in the cmake too.
Additionally, SCons might be an interesting approach to ease the engine generation as proposed in #42

PR #61, opened 2021-10-28
Restructure project to accommodate systems

### LINKED ISSUE #60: Restructure directories to accommodate systems
Since we'll start working on the engine systems, we now need to separate the core sources from the engine sources.
The current `src` and `include` folders should be moved to a `core` folder.
A new `engine` folder should be created which contains the engine sources.

With this change it makes sense to also move the `samples` and `tests` folders to the `core` folder.

### LINKED ISSUE #65: Move OGLRenderDevice to `src/`
Just like the `GLFWWindow` was moved to the `src` folder, the `OGLRenderDevice` should too.

RiscadoA:
Closes #60.
Closes #65.

Moved all core sources, dependencies, samples, tests, and configuration to the folder `core`.
Moved the old entry.cpp to the core samples directory and renamed it to `render_device.cpp` (render device sample).
Added a `engine` folder which will contain the engine entry point and other sources, such as the systems.
Added a basic `entry.cpp` to the `engine` source folder, which just creates an empty window.
Moved `images` to `docs/images`.


RiscadoA:
#65 should be fixed in this PR too.

joaopat98 (approved):

PR #71, opened 2021-11-03
Debug rendering

joaopat98:
Depends on #72


core/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ message( "# Building core tests: " ${BUILD_CORE_TESTS})
 # Set core source files
 
 set(CUBOS_CORE_SOURCE
+    "src/cubos/debug_render.cpp"
```
RiscadoA, line 23: Why here? Shouldn't it be under `src/cubos/gl/`?

core/CMakeLists.txt
```diff
@@ -32,6 +33,7 @@ set(CUBOS_CORE_SOURCE
 )
 
 set(CUBOS_CORE_INCLUDE
+    "include/cubos/debug_render.hpp"
```
RiscadoA, line 36: Why here? Shouldn't it be under `include/cubos/gl/`?

core/include/cubos/debug_render.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef CUBOS_DEBUG_RENDER_HPP
+#define CUBOS_DEBUG_RENDER_HPP
+
+#include <cubos/io/window.hpp>
+#include <glm/glm.hpp>
+
+namespace cubos
+{
+    class debug
```
RiscadoA, line 9: Fix case (`class Debug`)

core/include/cubos/debug_render.hpp
```diff
@@ -0,0 +1,21 @@
...
+    {
+    public:
+        static void init(cubos::io::Window* window);
```
RiscadoA, line 12: Maybe instead of passing the window pass just the render device? I I don't think the debug renderer needs to know the window exists.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include <list>
+
+#include <glm/gtc/matrix_transform.hpp>
+#include <glm/gtx/transform.hpp>
+
+using namespace cubos;
+using namespace cubos::gl;
+
+static RenderDevice* debugRenderDevice;
```
RiscadoA, line 13: Maybe instead of having a bunch of static globals, we should make them static fields of the class, just to keep it more clean.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,183 @@
...
+static ConstantBuffer mvpBuffer;
+static ShaderBindingPoint mvpBindingPoint, colorBindingPoint;
+static ShaderPipeline pipeline;
+static io::Window* debugWindow;
+
+static RasterState FillRasterState, WireframeRasterState;
+
+struct DebugDrawObject
```
RiscadoA, line 21: We can also define theses structs in the class, in the `private:` section.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,183 @@
...
+    VertexArray va;
+    IndexBuffer ib;
+    int numIndices;
+};
+
+static DebugDrawObject objCube, objSphere;
+
+struct DebugDrawRequest
+{
+    DebugDrawObject obj;
```
RiscadoA, line 32: Shouldn't this be a pointer/reference? Right now you're copying the whole DebugDrawObject which is unnecessarily expensive.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+void debug::drawWireSphere(glm::vec3 center, float radius, float time, glm::vec3 color)
+{
+    requests.push_back(DebugDrawRequest{objSphere, WireframeRasterState,
+                                        glm::translate(center) * glm::scale(glm::vec3(radius)), time, color});
+}
+
+void debug::flush(glm::mat4 vp, double deltaT)
+{
+    debugRenderDevice->setShaderPipeline(pipeline);
```
RiscadoA, line 157: This should be made thread safe: add a static std::mutex to the class

core/include/cubos/debug_render.hpp
```diff
@@ -0,0 +1,21 @@
...
+        static void drawCube(glm::vec3 center, glm::vec3 size, float time, glm::vec3 color = glm::vec3(1));
+        static void drawWireCube(glm::vec3 center, glm::vec3 size, float time, glm::vec3 color = glm::vec3(1));
+        static void drawSphere(glm::vec3 center, float radius, float time, glm::vec3 color = glm::vec3(1));
+        static void drawWireSphere(glm::vec3 center, float radius, float time, glm::vec3 color = glm::vec3(1));
+        static void flush(glm::mat4 vp, double deltaT);
+    };
```
RiscadoA, line 18: Add documentation to all of those methods

core/include/cubos/debug_render.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <cubos/io/window.hpp>
+#include <glm/glm.hpp>
+
+namespace cubos
+{
+    class debug
+    {
+    public:
+        static void init(cubos::io::Window* window);
+        static void drawCube(glm::vec3 center, glm::vec3 size, float time, glm::vec3 color = glm::vec3(1));
```
RiscadoA, line 13: Shouldn't it be possible to set the cube's rotation? Maybe optionally send a quaternion here.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,183 @@
...
+    auto sz = debugWindow->getFramebufferSize();
```
RiscadoA, line 158: This variable isn't being used.

RiscadoA (changes requested):

core/include/cubos/debug_render.hpp
joaopat98: It needs to know so it can know the current size of the framebuffer

core/src/cubos/debug_render.cpp
joaopat98: which methods should be thread safe exactly? the draw*() methods? I see no reason for the flush function to be called more than once per frame

core/include/cubos/debug_render.hpp
joaopat98: nvm, gonna remove it

core/include/cubos/debug_render.hpp
joaopat98: I don't really think this is needed, would only add more optional parameters, given that a cube with a fixed orientation should suffice for debugging purposes

core/src/cubos/debug_render.cpp
joaopat98: in order for them to be part of the class they would still have to be initialized in the cpp file though

RiscadoA: Yeah of course, its just that for example in the Settings singleton we are declaring the global variables as static class fields, so it would be more consistent to do the same here.

core/src/cubos/debug_render.cpp
RiscadoA: The draw and flush methods should be thread safe: you could have another thread running adding objects to the debug renderer at the same time as the previous frame is being rendered or somethint. If we don't make it thread safe, it should be made explicit in the docs.

core/include/cubos/debug_render.hpp
RiscadoA: I think it could also be really useful, for, for example, showing colliders which are not aligned with the axes. That parameter could just default to the identity quaternion, that way it doesn't get more complicated than needed.

core/src/cubos/debug_render.cpp
```diff
@@ -0,0 +1,484 @@
...
+using namespace cubos;
+using namespace cubos::gl;
+
+static RenderDevice* debugRenderDevice;
+static ConstantBuffer mvpBuffer;
+static ShaderBindingPoint mvpBindingPoint, colorBindingPoint;
+static ShaderPipeline pipeline;
+static io::Window* debugWindow;
+
+static RasterState FillRasterState, WireframeRasterState;
```
RiscadoA, line 19: Use camelCase for variables.

core/include/cubos/debug_render.hpp
joaopat98: is that really what the debug renderer should be used for though? If we want to show something like colliders there should be it's own rendering system for it I believe.

core/include/cubos/debug_render.hpp
```diff
@@ -1,20 +1,60 @@
...
+            DebugDrawObject& obj;
+            gl::RasterState rasterState;
+            glm::mat4 modelMatrix;
+            double timeLeft;
+            glm::vec3 color;
+        };
+
+        static std::list<DebugDrawRequest> requests;
+
+        static std::mutex debug_draw_mutex;
```
RiscadoA, line 54: Use camelCase.

core/include/cubos/debug_render.hpp
```diff
@@ -1,20 +1,60 @@
...
+
+    private:
+        static gl::RenderDevice* renderDevice;
+        static gl::ConstantBuffer mvpBuffer;
+        static gl::ShaderBindingPoint mvpBindingPoint, colorBindingPoint;
+        static gl::ShaderPipeline pipeline;
+
+        static gl::RasterState fillRasterState, wireframeRasterState;
+
+        struct DebugDrawObject
```
RiscadoA, line 34: Maybe reorder these methods, fields and type so that types come first, then methods and then static fields.

RiscadoA (changes requested):

core/CMakeLists.txt
```diff
@@ -22,11 +22,12 @@ message( "# Building core tests: " ${BUILD_CORE_TESTS})
 set(CUBOS_CORE_SOURCE
     "src/cubos/log.cpp"
     "src/cubos/settings.cpp"
-
+
```
RiscadoA, line 25: Fix whitespace issues (here are 4 extra spaces)

core/include/cubos/gl/debug_render.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include <list>
+#include <mutex>
+
+#include <glm/glm.hpp>
+#include <glm/gtc/quaternion.hpp>
+
+#include <cubos/gl/render_device.hpp>
+
+namespace cubos
```
RiscadoA, line 12: Change namespace to cubos::gl

core/include/cubos/gl/debug_render.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    class Debug
```
RiscadoA, line 14: Write docs for the class also, just something brief to describe what this does. I've also just noticed the name of the file and the class don't match up, shouldn't it be named `debug_renderer.hpp`? Or maybe just `debug.hpp` then since that's name of the class. Another option is renaming the class to `DebugRenderer`.

core/samples/debug_renderer.cpp
```diff
@@ -0,0 +1,50 @@
...
+        auto sz = window->getFramebufferSize();
+
+        auto vp = glm::perspective(glm::radians(70.0f), float(sz.x) / float(sz.y), 0.1f, 1000.0f) *
+                  glm::lookAt(glm::vec3{4.0f, 3.0f, -5.0f}, glm::vec3{0.0f, 0.0f, 0.0f}, glm::vec3{0.0f, 1.0f, 0.0f});
+
+        Debug::flush(vp, deltaT);
+        window->swapBuffers();
+        window->pollEvents();
+    }
+
```
RiscadoA, line 47: I've tested this sample and found that this gives us a segmentation fault: this is happening because the debug rendering GL objects are freed after the render device is freed. What this means is that we need to implement a Debug::terminate() function which clears up the handles (just set them to nullptr)

core/src/cubos/gl/debug_render.cpp
```diff
@@ -0,0 +1,471 @@
...
+    vaDesc.elements[0].buffer.offset = 0;
+    vaDesc.elements[0].buffer.stride = 3 * sizeof(float);
+    vaDesc.buffers[0] = vb;
+    vaDesc.shaderPipeline = pipeline;
+    objCube.va = renderDevice->createVertexArray(vaDesc);
+}
+
+void Debug::initSphere()
+{
+    float verts[] = {0.0f,
```
RiscadoA, line 61: Don't hardcode stuff like this: its too error prone and looks too ugly. With the cube its alright since its easy to understand.
Instead try to write code which generates a sphere. You can implement something like [this](https://github.com/opencv/opencv/blob/4295ed29fa8c2db2fe255467f822190a8e53b24a/modules/highgui/src/viz3d/viz3d.cpp#L364).

core/src/cubos/gl/debug_render.cpp
```diff
@@ -0,0 +1,471 @@
...
+        it->timeLeft -= deltaT;
+
+        auto current = it++;
+        if (current->timeLeft <= 0)
+        {
+            requests.erase(current);
+        }
+    }
+    debugDrawMutex.unlock();
+    renderDevice->setRasterState(nullptr);
```
RiscadoA, line 470: Unnecessary, you can remove this if you want. Other rendering code should be responsible for setting its rasterizer state, so there's no need to have an extra call here.

core/src/cubos/gl/debug_render.cpp
```diff
@@ -0,0 +1,471 @@
...
+        colorBindingPoint->setConstant(it->color);
+
+        renderDevice->setRasterState(it->rasterState);
+        renderDevice->drawTrianglesIndexed(0, it->obj.numIndices);
+
+        it->timeLeft -= deltaT;
+
+        auto current = it++;
+        if (current->timeLeft <= 0)
+        {
```
RiscadoA, line 465: Remove braces since its just one line of code.

RiscadoA (changes requested):

core/include/cubos/gl/debug.hpp
```diff
@@ -87,6 +89,6 @@ namespace cubos
 
         static std::mutex debugDrawMutex;
     };
-} // namespace cubos
+} // namespace cubos::gl
 
 #endif // CUBOS_DEBUG_RENDER_HPP
```
RiscadoA, line 94: Change include guard to CUBOS_GL_DEBUG_HPP

RiscadoA (changes requested):

RiscadoA (approved):

PR #73, opened 2021-11-09
Bootstrap Input System

ratuspro:
The proposed Input System focuses on decoupling the input bindings from the gameplay actions.
Additionally, this input system will allows to serialize the all the actions and bindings.

To create a **fire action** that can be bound to the **mouse's Right Button** or an **xbox gamepad's A Button** or the **keyboard's Space Key**, the gameplay developer should do the following: 

```cpp
ActionMapping actionFire = InputManager.CreateNewAction('fire');
actionFire.registerBinding('mouse_rightButton_down');
actionFire.registerBinding('xbox_AButton_down');
actionFire.registerBinding('keyboard_spaceKey_down');
actionFire.onTriggered += Fire;
...
void Fire(InputContext context) { 
    // E.g. shoot a projectile
}
```

If the action triggered has some value, the gamplay developer can access it through the input context parameter on the callback:
```cpp
ActionMapping actionMove = InputManager.CreateNewAction('player_move');
actionMove.registerBinding(InputCombo('keyboard_AKey_down', 'keyboard_DKey_down', 'keyboard_WKey_down', 'keyboard_SKey_down'));
actionMove.registerBinding('xbox_leftJoystick');
actionMove.onTriggered += MoveCharacter;
...
void MoveCharacter(InputContext context){
    Vector2 translation = context.ReadValue_Vector2();
    // E.g. move character position
}
```

The input mappings will also be able to be dumped and loading using a format similar to

```json
{
  "action": [
    {
      "name": "fire",
      "bindings": [
        "mouse_rightButton_down",
        "xbox_AButton_down",
        "keyboard_spaceKey_down"
      ]
    },
    {
      "name": "move",
      "bindings": [
        {
          "type": "combo",
          "xNegative": "keyboard_AKey_down",
          "xPositive": "keyboard_DKey_down",
          "yNegative": "keyboard_WKey_down",
          "yPositive": "keyboard_SKey_down"
        },
        "xbox_leftJoystick"
      ]
    }
  ]
}
```

Still, the callback functions associated with each action will not be dumped in it. However, they can be registered after loading the mappings file as follows:

```cpp
InputManager.LoadMapping(mappings_file);
ActionMapping actionFire = InputManager.GetAction('fire');
actionFire.onTriggered += Fire; 
```

This PR attempts to solve #7.


core/src/cubos/input/input_manager.cpp
```diff
@@ -8,6 +8,40 @@ using namespace cubos::input;
...
+
+void InputManager::handleKeyDown(Key key)
+{
+    if (InputManager::keyBindings.empty() || !InputManager::keyBindings.contains(key) ||
+        InputManager::keyBindings[key].empty())
+    {
+        return;
+    }
+
+    for (std::vector<std::function<void(void)>>::iterator it = InputManager::keyBindings[key].begin();
```
RiscadoA, line 25: Here you can just use `auto it = ...`, there's no need to specify the whole type.

DannyIsYog:
Lambda functions don't work when using them as callback on the register binding functions, let it be given by us or the programmer. Need to find a work around that.

ratuspro:
We should consider changing how keys and mouse button callbacks are being registered and invoked.

Currently, we are creating a lambda to wrap the function being called 
```
InputManager::registerKeyDownCallback([this]() { this->handleButtonDown(); }, std::get<cubos::io::Key>(this->button));
```
and then just invoke the lambdas when processing the sources:
```
for (auto itCallback = itCallbacksVector->second.begin(); itCallback != itCallbacksVector->second.end(); itCallback++){
    (*itCallback)();
}
```

Instead, we should register the Input Source
```
InputManager::registerKeyDownCallback(this, std::get<cubos::io::Key>(this->button));
```
and then invoke the handleKeyDownBased on the InputSource reference
```
for (auto itCallback = itCallbacksVector->second.begin(); itCallback != itCallbacksVector->second.end(); itCallback++){
    itCallback->handleButtonDown();
}
```

This allows for a greater control over the callback, stored as a instance rather than a lambda. For instance, for unregister operations.

ratuspro:
I would say that the input_manager doesn't need to know about the implementations of the input_sources.
The game developer should include only the ones that are relevant in the game logic source. Otherwise, the input_manager would be always including everything(e.g. ButtonPress, SingleAxis, DoubleAxis) even if they were not used.

ratuspro:
The InputContext can only accept one value from a specific type (for now float or glm::vec2). See e912ee62ddc9aa865122da6dc31e93e462e1c368
We should considering improving the InputContext to easily accommodate multiple values of the same type (or not) that can be accessed by the game developer through the InputContext. (e.g. normalizedDelta, rawDelta)
Perhaps they can be stored in a map<str, T> and each InputSource registers the values that they can produce. Alternatively, we can use a struct for each InputSource with their relevant values that can be added to the InputContext.

core/include/cubos/input/input_sources/single_axis.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace cubos::input
+{
+
+    class SingleAxis : public InputSource
+    {
+
+    public:
+        SingleAxis(cubos::io::MouseAxis axis);
+        SingleAxis(cubos::io::Key negativeKey, cubos::io::Key positiveKey);
```
DannyIsYog, line 17: I was trying to think the structure of the single axis constructor and this seems like the obvious way but should't we allow for other types of inputs? Keyboards, joysticks, controllers, etc...
And if so would it make sense to allow combining them? e.g posAxis is binded to a keyboard Key and the negAxis to a mouse button.
This would probably make a nightmare for constructors, I don't know if we can use templates or variants for this.
Thoughts?

ratuspro: The constructor will always need to be there. Either in the same class SingleAxis or in multiple classes SingleJoystickActions, SingleMouseAxis. 

The only problem might be associated with the number of handle functions that need to be created for supporting multiple controls. However, has the output of the controls stays the same (e.g. button being pressed, axis changing a value) we will be fine. Even if we split the multiple axis supported across multiple implementations, we would always need to do multiple handles.

I prefer this way because it is clear to the game developer what each input source produces regardless of the controls that produced it.

ratuspro:
Solves #7

ratuspro:
@RiscadoA suggestions:
- [ ] change from `cubos/input` to `cubos/io`
- [ ] change from `input_sources` to `sources `
- [ ] extract context and action to other files input_manager.hpp

ratuspro:
Followed @RiscadoA suggestions.

We still need to make the Input Manager not static but we should close this PR first.
Later I'll address the multiple `register****Callbacks` too.

Issue #81, opened 2022-01-11
Compressed Archives


RiscadoA:
Implement compressed Archives. See #16.

PR #82, opened 2022-01-11
FileSystem and STDArchive

### LINKED ISSUE #16: Archives

RiscadoA:
Closes #16.

- adds the `File` class, which represents a `File` in the virtual file system.
- adds the generic `Archive` class, which acts as a mount point from the real file system to the virtual one.
- implements the `STDArchive` class, which is an archive used to mount OS directories or files.
- adds the `FileSystem` class, used as a utility to manage the entire virtual file system.
- added a test for `STDArchive`.

Compressed archives will be left for future implementation since they're not a priority right now. Check #81


RiscadoA:
- [ ] Make destroy() return false instead of calling abort

PR #91, opened 2022-01-12
Defined gl::Material, gl::Palette and gl::Grid

### LINKED ISSUE #83: Define voxel data structures
Properties for each voxel in a model will be defined using a palette.
Palettes can use different index sizes (8-bit, 16, 32) depending on how many materials are needed for the model.
Palettes can be shared between models.

First Implementation of voxel structures:
- Simple 3d grid (Array)
- each slot in the grid contains index of the material within the model's palette (first index (0) will be considered as empty)

(IDEA)
2 Different palette types:
- Gpu Only: data sent to shaders (only properties needed to draw the model)
- Engine: Contains other properties of the voxels, such as physics info, metadata, etc.

RiscadoA:
Closes #83.


RiscadoA:
- [ ] Add assertions to get and set on *Grid*

PR #94, opened 2022-01-16
Add parseQB() function for loading gl::Grid and gl::Palette

RiscadoA:
Closes #94.

- Adds `memory/endianness.hpp` which contains utility functions for converting between different endianness.
- Adds `parseQB()` which parses QB files from a stream, outputting a vector of `QBMatrix`.
- Adds the `clear()` method and move constructor to the `Grid` class.


RiscadoA:
TODO: Rename namespaces

Issue #98, opened 2022-01-21
Github actions for Windows


joaopat98:

ratuspro:
This might bring some troubles in the future - as described in #102 - but it is working for now.

PR #102, opened 2022-01-21
Creating a Github Action to build the engine on windows

ratuspro:
Works on #98


ratuspro:
The compilation uses Visual Studio 16 2019 and github environment with windows-latest which is Windows-Server 2019 (see [here](https://github.com/actions/virtual-environments)) 
It will eventually [move to Windows Server 2022](https://github.com/actions/virtual-environments/issues/4856) which might cause some problems in our build pipeline.

Closes #102

Issue #48, opened 2021-10-19
Settings serialization


RiscadoA:
Depends on #15.
Implement loading and storing settings.

joaopat98:
Add interface for basic types (doubles, strings, etc.), stored as strings. When getting, if not able to convert, return type default.

ratuspro:
> When getting, if not able to convert, return type default.

I did something different:
- When there is a setting with the key and can return the requested type, returns the stored value.
- When there is a setting with the key and cannot return the requested type, raises an exception.
- When there is no setting with the key, returns the default value

It helps the game developer to identify errors easily since it does not hide them by returning the default value.

ratuspro:
Solved on #104

joaopat98:
@ratuspro

PR #104, opened 2022-01-26
Add Global Settings with Serialization

ratuspro:
Solves #48


ratuspro:
The settings does not hold on its state the actual types of each setting.
Whenever possible, it will return the value. Otherwise, raises an exception.

PR #105, opened 2022-01-27
Add object callbacks to Events

ratuspro:
Besides static methods, it should be possible to register methods.


ratuspro:
Needed on #73

ratuspro:
@RiscadoA
- [ ] in the Event class, the method callback definition was added to a new private: section which is above the already existing public: and private: sections. it should be moved to the original private: section, since what should be visible first is the interface with which we will interact
- [ ] in the Event class the new method definitions should be moved below the class like the old ones, so that the declaration is easy to read

ratuspro:
See #106

PR #118, opened 2022-02-19
Add generic Factory class

### LINKED ISSUE #110: Create generic Factory class
Factory class to facilitate the automatic registering of types in the engine's various systems. Needed/useful for ECS and Input Sources.

RiscadoA:
Closes #110


RiscadoA:
TODO: Rename namespaces & check if this will actually be used

PR #119, opened 2022-02-19
Make cubos-engine a library and add empty sample

### LINKED ISSUE #92: Entry point restructuring
Change cubos-engine to be library, "entry point" should change to a cubos::init(...) function

RiscadoA:
Closes #92


joaopat98:
"entry.hpp" doesn't feel quite right, maybe just a self titled file like cubos/cubos.hpp? Or something like engine.hpp...

Otherwise all good I think

joaopat98:
also we might want to separate the init function from the window loop, maybe put that in something like "cubos::run()". Workflow would be cubos::init(), setup objects/resources, cubos::run() for example

PR #129, opened 2022-03-23
Extend Input Manager

joaopat98:


RiscadoA:
TODO: Rename namespaces

ratuspro (approved):

ratuspro:
The way SingleAxis and other sources handle multiple inputs and maintains its state might lead to scalability problems in the future since we are not normalizing the delta of the source (e.g. when handling mousewheels). 
However, it would require some additional effort to refactor. This contribution is sound and the PR should be merged @RiscadoA.

ratuspro:
More extensions to the Input Manager (e.g. new sources or adding features to existent sources) should be done in separate branches.

PR #161, opened 2022-05-31
Enhance Renderer

RiscadoA:
This PR aims to enhance the structure of the renderer and pave the way for a future raytracing renderer implementation.
It will also fix compatibility issues with older OpenGL versions by using textures instead of buffers for storing material palettes.


core/src/cubos/core/gl/ogl_render_device.cpp
```diff
@@ -1079,6 +1063,8 @@ class OGLShaderPipeline : public impl::ShaderPipeline
 
 OGLRenderDevice::OGLRenderDevice()
 {
+    glGetString(GL_VERSION);
```
Ev1lbl0w, line 1066: Danglig code :wink:

core/src/cubos/core/gl/util.cpp
```diff
@@ -3,34 +3,30 @@
...
+    desc.elements[0].buffer.stride = 4 * sizeof(float);
+    desc.elements[1].name = "uv";
+    desc.elements[1].type = gl::Type::Float;
+    desc.elements[1].size = 2;
+    desc.elements[1].buffer.index = 0;
+    desc.elements[1].buffer.offset = 2 * sizeof(float);
+    desc.elements[1].buffer.stride = 4 * sizeof(float);
+    desc.buffers[0] = renderDevice.createVertexBuffer(sizeof(screenVerts), screenVerts, gl::Usage::Static);
+    desc.shaderPipeline = pipeline;
+    va = renderDevice.createVertexArray(desc);
```
Ev1lbl0w, line 31: Why were the indices removed?

engine/src/cubos/engine/gl/deferred/renderer.cpp
```diff
@@ -1,465 +1,569 @@
...
-{
-    if (lights.numDirectionalLights >= CUBOS_DEFERRED_RENDERER_MAX_DIRECTIONAL_LIGHT_COUNT)
-    {
-        logWarning("deferred::Renderer::drawLight(const DirectionalLight& light) failed: number of directional "
-                   "lights to be drawn this frame exceeds CUBOS_DEFERRED_RENDERER_MAX_DIRECTIONAL_LIGHT_COUNT.");
-        return;
-    }
-    lights.directionalLights[lights.numDirectionalLights++] = light;
+    // Update the data to the GPU.
+    this->paletteTex->update(0, 0, 256, 256, &data[0]);
```
Ev1lbl0w, line 382: For `std::vector` you can do `data.data()` instead for a cleaner way:
```suggestion
    this->paletteTex->update(0, 0, 256, 256, data.data());
```

Ev1lbl0w (approved):
Looks good to me! Only have a few nitpicks to point.

core/src/cubos/core/gl/ogl_render_device.cpp
RiscadoA: Oops ahahaah

core/src/cubos/core/gl/util.cpp
RiscadoA: They were unnecessary, for such a small vertex buffer, having two extra vertices won't make any performance difference.
Having an index buffer complicates the process since we need to pass it around.

PR #194, opened 2022-09-08
Implement SSAO effect on renderer

Ev1lbl0w:
Implements SSAO (screen-space ambient occlusion) effect
image

Beyond the effect itself (implemented in defered `renderer.cpp/h`), some other changes had to be made to the engine as well:
- Add short/half-float texture formats (R16FLoat, RGB16Float, etc...)
- Fix invalid enum constant `GL_R` (it should be `GL_RED`, source: https://stackoverflow.com/questions/41396731/in-opengl-whats-the-difference-between-the-gl-red-and-gl-r-color-formats)
- Added function to pas a mat4x4 to a shader binding point
- Fix overflow when reading certain cube models, leading to an infinite loop (`grid.cpp`)


Ev1lbl0w:
PR ready for review :tada:

engine/include/cubos/engine/gl/deferred/renderer.hpp
```diff
@@ -33,13 +33,20 @@ namespace cubos::engine::gl::deferred
         virtual RendererGrid upload(const core::gl::Grid& grid) override;
         virtual void setPalette(const core::gl::Palette& palette) override;
 
+        bool isSSAOEnabled() const;
+        void setSSAOEnabled(bool ssaoEnabled);
```
RiscadoA, line 37: Maybe instead of getting/setting data from here, we should we just fetch it directly from the Settings global object.
This way we avoid dynamic casts later on.

engine/include/cubos/engine/gl/deferred/renderer.hpp
```diff
@@ -33,13 +33,20 @@ namespace cubos::engine::gl::deferred
...
 
         virtual void onResize(glm::uvec2 size) override;
         virtual void onRender(const core::gl::Camera& camera, const Frame& frame,
                               core::gl::Framebuffer target) override;
 
+    private:
+        void createSSAOTextures();
+        void generateSSAONoise();
+
     private:
```
RiscadoA, line 50: Duplicate private section

engine/include/cubos/engine/gl/deferred/renderer.hpp
```diff
@@ -67,13 +74,37 @@ namespace cubos::engine::gl::deferred
...
+        core::gl::ShaderBindingPoint ssaoTexBP;
         core::gl::Sampler sampler;
         core::gl::Texture2D paletteTex;
         core::gl::ConstantBuffer lightsBuffer;
 
         // Screen quad used for the lighting pass.
 
         core::gl::VertexArray screenQuadVA;
+
+    private:
```
RiscadoA, line 87: Duplicate private section

RiscadoA (changes requested):
Good job! Looks great, lemme just point out some nitpicks

engine/include/cubos/engine/gl/deferred/renderer.hpp
RiscadoA: We should recheck the value from the settings onResize

Ev1lbl0w: Fixed. Also onResize was already implemented :relieved:

engine/include/cubos/engine/gl/deferred/renderer.hpp
Ev1lbl0w: Fixed

engine/include/cubos/engine/gl/deferred/renderer.hpp
Ev1lbl0w: Fixed

RiscadoA (approved):

PR #218, opened 2022-11-09
Improve logging in the core and engine

### LINKED ISSUE #201: Add logging macros
Add logging macros that print the current file and line and which may be deactivated at compile time.

RiscadoA:
- [X] Add logging macros.
- [x] Review `core` source code and add/improve logging where possible.
- [x] Review `engine` source code and add/improve logging where possible.

Closes #201 
Depends on #221


core/include/cubos/core/ecs/resource_manager.hpp
```diff
@@ -142,7 +142,7 @@ namespace cubos::core::ecs
         auto it = resources.find(typeid(T));
         if (it != resources.end())
         {
-            logCritical("ecs::ResourceManager::add: Resource of type {} was already added.", typeid(T).name());
+            CUBOS_CRITICAL("Could not registe resource of type {}: already registered", typeid(T).name());
```
rsubtil, line 145: Typo:
```suggestion
            CUBOS_CRITICAL("Could not register resource of type {}: already registered", typeid(T).name());
```

core/src/cubos/core/data/file.cpp
```diff
@@ -68,19 +69,19 @@ void File::mount(std::string_view path, std::shared_ptr<Archive> archive)
...
         {
-            logError("Could not mount archive at path '{}', a file already exists at that path", dir->path);
+            CUBOS_CRITICAL("Could not mount archive at path '{}', a file already exists at that path", dir->path);
             abort();
         }
 
         if (dir->archive != nullptr)
         {
-            logError("Could not mount archive at root, an archive has already been mounted at root");
+            CUBOS_CRITICAL("Could not mount archive at root, an archive has already been mounted at root");
```
rsubtil, line 78: I'd remove the ending `at root` to avoid repetition:
```suggestion
            CUBOS_CRITICAL("Could not mount archive at root, an archive has already been mounted");
```

core/src/cubos/core/data/file.cpp
```diff
@@ -68,19 +69,19 @@ void File::mount(std::string_view path, std::shared_ptr<Archive> archive)
...
             abort();
         }
 
         if (!archive->isDirectory(1))
         {
-            logError("Could not mount archive at root, archive must be a directory to be mounted at root");
+            CUBOS_CRITICAL("Could not mount archive at root, archive must be a directory to be mounted at root");
```
rsubtil, line 84: Ditto
```suggestion
            CUBOS_CRITICAL("Could not mount archive at root, archive must be a directory to be mounted");
```

core/src/cubos/core/data/file.cpp
```diff
@@ -93,8 +94,8 @@ void File::mount(std::string_view path, std::shared_ptr<Archive> archive)
     {
         if (dir->findChild(name) != nullptr)
         {
-            logError("Could not mount archive at path '{}', a file already exists at that path",
-                     dir->path + "/" + std::string(name));
+            CUBOS_CRITICAL("Could not mount archive at path '{}', a file already exists at that path",
```
rsubtil, line 97: Changed from `error` to `critical`

core/src/cubos/core/data/file.cpp
```diff
@@ -444,8 +452,5 @@ File::~File()
 {
     if (this->destroyed && this->archive)
         if (!this->archive->destroy(this->id))
-        {
-            logError("Could not destroy file '{}', internal archive error", this->path);
-            abort();
```
rsubtil, line 449: This `abort()` was removed, intended?

core/src/cubos/core/data/package.cpp
```diff
@@ -494,14 +535,20 @@ impl::Unpackager::Unpackager(const Package& pkg) : pkg(pkg)
     // Do nothing.
 }
 
+#define UNEXPECTED(expected, found)                                                                                    \
+    do                                                                                                                 \
```
rsubtil, line 539: Why do a `do {...} while(false);`? Wouldn't it work with just `{...}`?

core/src/cubos/core/io/glfw_window.cpp
```diff
@@ -23,12 +23,19 @@ static Key glfwToCubosKey(int key);
...
+        CUBOS_CRITICAL("Unsupported when building without GLFW");                                                      \
+        abort();                                                                                                       \
+    } while (0)
+
 GLFWWindow::GLFWWindow()
 {
 #ifdef WITH_GLFW
     if (!glfwInit())
     {
-        logCritical("OGLRenderDevice::OGLRenderDevice() failed: glfwInit() failed");
```
rsubtil, line 31: Can be replaced with the new `UNSUPPORTED` macro

engine/src/cubos/engine/data/meta.cpp
```diff
@@ -77,7 +77,7 @@ void cubos::core::data::deserialize(Deserializer& deserializer, Meta& meta)
     }
     else
     {
-        core::logError("asset::Meta::deserialize() failed: Invalid usage type '{}', defaulting to Static", usage);
+        CUBOS_ERROR("Invalid asset usage type '{}', defaulting to Usage::Static", usage);
```
rsubtil, line 80: Ditto, `error` to `warning`

engine/src/cubos/engine/gl/deferred/renderer.cpp
```diff
@@ -605,7 +605,7 @@ void deferred::Renderer::onRender(const Camera& camera, const Frame& frame, Fram
     {
         if (lightData.numSpotLights >= CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT)
         {
-            logWarning("deferred::Renderer::onRender() failed: number of spot lights to be drawn "
+            CUBOS_WARN("Number of spot lights to be drawn "
                        "this frame exceeds CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT ({}).",
```
rsubtil, line 609: I'd rename this to not include the raw variable name:
```suggestion
                       "this frame exceeds the maximum allowed ({}).",
```

engine/src/cubos/engine/gl/deferred/renderer.cpp
```diff
@@ -626,7 +626,7 @@ void deferred::Renderer::onRender(const Camera& camera, const Frame& frame, Fram
     {
         if (lightData.numDirectionalLights >= CUBOS_DEFERRED_RENDERER_MAX_DIRECTIONAL_LIGHT_COUNT)
         {
-            logWarning("deferred::Renderer::onRender() failed: number of directional lights to be drawn "
+            CUBOS_WARN("Number of directional lights to be drawn "
```
rsubtil, line 629: Ditto

engine/src/cubos/engine/gl/deferred/renderer.cpp
```diff
@@ -643,7 +643,7 @@ void deferred::Renderer::onRender(const Camera& camera, const Frame& frame, Fram
     {
         if (lightData.numPointLights >= CUBOS_DEFERRED_RENDERER_MAX_POINT_LIGHT_COUNT)
         {
-            logWarning("deferred::Renderer::onRender() failed: number of point lights to be drawn "
+            CUBOS_WARN("Number of point lights to be drawn "
```
rsubtil, line 646: Ditto

rsubtil (commented):
Looks good, but there's some things important to confirm before approving.

I've also found many instances where logging levels changed, for example from `error` to `critical`:
```cpp
logError("Could not mount archive at path...");
```
```cpp
CUBOS_CRITICAL("Could not mount archive at path...");
```
There are too many instances so I didn't make a comment on each one, it's just a reminder to ensure these were all intended.

core/src/cubos/core/data/package.cpp
RiscadoA: It's common practice to use `do {...} while(false)` to ensure that the a `;` must follow the macro, as if it was a normal statement.

core/src/cubos/core/data/file.cpp
RiscadoA: Yeah, no need to abort() here, we can just log the error.

core/include/cubos/core/ecs/world.hpp
```diff
@@ -133,14 +135,19 @@ namespace cubos::core::ecs
 
         auto entity = this->entityManager.create(mask);
         ([&](auto& component) { this->componentManager.add(entity.index, component); }(components), ...);
+
+#if CUBOS_LOG_LEVEL <= CUBOS_LOG_LEVEL_DEBUG
```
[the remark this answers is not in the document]
RiscadoA: Intended, it is still controlled at runtime through `CUBOS_DEBUG`.
The only issue is that there's the overhead of creating the strings even if we don't get to use them, but this only happens if we're compiling with debug information, so it should be okay.
What do you think?

core/src/cubos/core/io/glfw_window.cpp
RiscadoA: The `UNSUPPORTED` macro is being used to log the error of trying to use GLFW when you built cubos without it.
You may have built it with GLFW and still glfwInit() fails, so its a different kind of error.

RiscadoA:
> Looks good, but there's some things important to confirm before approving.
> 
> I've also found many instances where logging levels changed, for example from `error` to `critical`:
> 
> ```c++
> logError("Could not mount archive at path...");
> ```
> 
> ```c++
> CUBOS_CRITICAL("Could not mount archive at path...");
> ```
> 
> There are too many instances so I didn't make a comment on each one, it's just a reminder to ensure these were all intended.

Thank you for the review!
They were intended, but you're right about using `CUBOS_WARN` instead of `CUBOS_ERROR` when execution just continues as normal, or when it defaults to some value.
I noticed there were lots of places where `CUBOS_ERROR` was being used even though `abort()` was being called after that, so I switched them to `CUBOS_CRITICAL`.

core/src/cubos/core/io/glfw_window.cpp
rsubtil: Ah, alright then :+1:

core/include/cubos/core/ecs/world.hpp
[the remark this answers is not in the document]
rsubtil: Ah I see what you mean. I don't think it's much of an issue since this is only compiled for a debug build, which is expected to be unoptimized and slower for the sake of debugging purposes. This overhead won't happen on release builds which is what developers should do for final products anyways.

core/src/cubos/core/data/package.cpp
rsubtil: I see, the more you know :slightly_smiling_face:

rsubtil (approved):
Looks good to me! :+1:

PR #223, opened 2022-11-30
Rename Window to BaseWindow

RiscadoA:
Depends on #218.

Changes:
- renames `Window` to `BaseWindow`.
- instead make `Window` an alias to `std::shared_ptr<BaseWindow>`.
- allow setting the size and the title of the window when opening it.


rsubtil (approved):
Looks good to me! :+1:

PR #224, opened 2022-11-30
Remove Settings singleton

### LINKED ISSUE #217: Remove settings singleton
A settings singleton is no longer necessary as it will now be used as a resource.

RiscadoA:
Closes #217

The settings 'global' singleton is no longer necessary since it will now be used as a Resource.


rsubtil (approved):
Looks good to me :+1:

PR #240, opened 2022-12-01
Add audio device 

### LINKED ISSUE #204: Add Audio Device

JoseGraca96256:
Closes #204


core/CMakeLists.txt
```diff
@@ -188,6 +195,13 @@ if (WITH_GLFW)
     target_compile_definitions(cubos-core PRIVATE WITH_GLFW)
 endif ()
 
+if (WITH_OPENAL)
+    add_subdirectory(lib/openal-soft)
```
rsubtil, line 199: OpenAL has some [settings on by default](https://github.com/kcat/openal-soft/blob/master/CMakeLists.txt#L91-L94) that results in extra tools not needed. These should be disabled. I've only found these so far but there might be more stuff that we do not want:
```suggestion
    set(ALSOFT_UTILS OFF CACHE BOOL "" FORCE)
    set(ALSOFT_NO_CONFIG_UTIL OFF CACHE BOOL "" FORCE)
    set(ALSOFT_EXAMPLES OFF CACHE BOOL "" FORCE)
    add_subdirectory(lib/openal-soft)
```

core/include/cubos/core/al/audio_device.hpp
```diff
@@ -0,0 +1,152 @@
...
+        AudioDevice(const AudioDevice&) = delete;
+
+        /// Creates an audio device from a given device specifier.
+        /// @param specifier The device specifier (empty for default).
+        /// @returns The audio device.
+        static std::shared_ptr<AudioDevice> create(std::string specifier = "");
+
+        /// Enumerates the available devices.
+        /// @param devices The vector to fill with the available devices.
+        static void enumerateDevices(std::vector<std::string>& devices);
```
rsubtil, line 45: It's better to let the implementation create the array, instead of asking users to create the object before hand:
```suggestion
        static std::vector<std::string> enumerateDevices();
```

core/include/cubos/core/al/audio_device.hpp
```diff
@@ -0,0 +1,152 @@
...
+            /// @param direction The cone direction of the source.
+            virtual void setConeDirection(const glm::vec3& direction) = 0;
+
+            /// Sets the reference distance of the source.
+            /// @param referenceDistance The reference distance of the source.
+            virtual void setReferenceDistance(float referenceDistance) = 0;
+
+            /// Plays the source.
+            virtual void play() = 0;
+
```
rsubtil, line 144: We should add virtual functions for pausing/stopping too.

core/src/cubos/core/al/oal_audio_device.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    virtual ~OALSource() override
+    {
+        alDeleteSources(1, &this->id);
+    }
+
+    virtual void setBuffer(Buffer buffer) override
+    {
+        auto oalBuffer = std::dynamic_pointer_cast<OALBuffer>(buffer);
+        alSourcei(this->id, AL_BUFFER, oalBuffer->id);
```
rsubtil, line 65: Why is this needed? Can't the id value be supplied directly?
```suggestion
        alSourcei(this->id, AL_BUFFER, buffer.id);
```

core/src/cubos/core/al/oal_audio_device.cpp
```diff
@@ -0,0 +1,202 @@
...
+Buffer OALAudioDevice::createBuffer()
+{
+    ALuint id;
+    alGenBuffers(1, &id);
+
+    return std::make_shared<OALBuffer>(id);
+}
+
+Source OALAudioDevice::createSource()
+{
```
rsubtil, line 185: Can't this be implemented already?

rsubtil (changes requested):
Good job! I've left a few notes I picked.

Regarding CI, I'm having a look to figure out what it might be. So far I got it compiling on Apple by switching to Clang, now to figure out Windows...

(I introduced a commit that was supposed to go to my fork for experiments with the CI, that I have since reverted, sorry)

core/src/cubos/core/al/oal_audio_device.cpp
rsubtil: Seems like this empty function is the reason CI fails on Windows, so it definitely needs to be implemented.

rsubtil (approved):

rsubtil (approved):

Issue #241, opened 2022-12-03
Redesign dispatcher


RiscadoA:
- Its hard to specify dependencies between plugins which don't care about each other - `envSettings`, `fileSettings` before `assetsPlugin`.
- should we tag systems to prevent direct dependencies?
- way to group multiple systems?
- make systems or groups of systems run based on conditional systems, which return a `bool` instead of `void`.

The solution could look like:
```cpp
cubos
        .addSystem(mySystem1, mySystem2, mySystem3)
        .tag("assets")
        .after(myOtherSystem)
        .after(All("settings"));
```

Before we start writing code we should write pseudo code of the cars sample using the solution.

RiscadoA:
So, here is my proposal:

```cpp
void windowPlugin(Cubos& cubos)
{
    cubos
        .addStartupSystem(openWindowSystem)
        .tag("window")
        .after("settings")
        .addSystem(pollEventsSystem)
        .tag("poll")
        .addSystem(swapBuffersSystem)
        .tag("swap-buffers");
} 
```

```cpp
void settingsProviderPlugin(Cubos& cubos)
{
    cubos
        .addStartupSystem(provideSettingsSystem)
        .tag("settings");
}
```

```cpp
// Adds a new condition system with tag 'fixed-step' which runs at a fixed time step.
void fixedStepPlugin(Cubos& cubos)
{
   cubos
       .addConditionSystem(returnTrueWhenShouldStepSystem)
       .tag("fixed-step");
}
```

```cpp
void collisionsPlugin(Cubos& cubos)
{
   cubos
        .addPlugin(fixedStepPlugin)
        .beginGroup()
        .tag("sim-step") // tag all systems within the group
        .addSystem(genCollisionEvents)
        (...)
        .endGroup()
}
```

RiscadoA:
If we follow with this, internally, the dispatcher will need to find an execution order for the systems, and also decide when to commit commands. There are three types of constraints to consider (ordered by their 'strength':
- a system may need to only be run after the commands of another system have been commited: effectivelly a hard synchronization point. When commands are commited, no system can be running.
- a system may need to only be run after another finishes - a pair of systems may be forced to run sequentially, in a specific order.
- two systems may access the same resources in an incompatible way (write and read, write and write), in which case they must be run sequentially.

RiscadoA:
Some systems may need to be run on a specific thread (e.g., event pooling, and stuff which uses the render device)

RiscadoA:
[...]

PR #242, opened 2022-12-06
Implement window plugin

rsubtil:
This implement our first core plugin, `windowPlugin`, which users should add if they want to show output to a window. With this, window generation, polling and swapping buffers is handled internally by the engine.

A few bugs had to be fixed too:
- The dispatcher had a bug when a given element was erased from the stages; this may invalidate the `refIt` iterator, causing a crash.
- Some `Options` objects were not initialized on `cubinhos`, causing crashes.

(Rebasing this was becoming ugly, so I had to make a temporary `cubos_class_window_copy` branch. Please only delete this after this is merged, just in case TM)

### DIFF
--- a/engine/include/cubos/engine/plugins/window.hpp
+++ b/engine/include/cubos/engine/plugins/window.hpp
@@ -9,7 +9,17 @@ using namespace cubos::engine;
 
 namespace cubos::engine::plugins
 {
+    /// Plugin to create and handle the lifecycle of a window. This is required
+    /// for showing any visual output to the user.
+    ///
+    /// Startup Stages:
+    /// - openWindow: Opens the window.
+    ///
+    /// Stages:
+    /// - poll: Polls the window for events and handles quit requests.
+    /// - swapBuffers: Swaps the window's back buffers.
+    /// @param cubos CUBOS. main class
     void windowPlugin(Cubos& cubos);
-};
+}; // namespace cubos::engine::plugins
 
 #endif // CUBOS_ENGINE_PLUGINS_WINDOW_HPP
--- a/engine/include/cubos/engine/plugins/window.hpp
+++ b/engine/include/cubos/engine/plugins/window.hpp
@@ -1,0 +1,15 @@
+#ifndef CUBOS_ENGINE_PLUGINS_WINDOW_HPP
+#define CUBOS_ENGINE_PLUGINS_WINDOW_HPP
+
+#include <cubos/core/io/window.hpp>
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins
+{
+    void windowPlugin(Cubos& cubos);
+};
+
+#endif // CUBOS_ENGINE_PLUGINS_WINDOW_HPP
--- a/engine/src/cubos/engine/plugins/window.cpp
+++ b/engine/src/cubos/engine/plugins/window.cpp
@@ -1,0 +1,42 @@
+#include <cubos/engine/plugins/window.hpp>
+#include <cubos/core/settings.hpp>
+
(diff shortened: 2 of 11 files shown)


engine/src/cubos/engine/plugins/window.cpp
```diff
@@ -0,0 +1,42 @@
...
+        {
+            settings.getInteger("window.width", 800),
+            settings.getInteger("window.height", 600)
+        }
+    );
+}
+
+static void pollSystem(const cubos::core::io::Window& window, ShouldQuit& quit)
+{
+    window->pollEvents();
```
RiscadoA, line 18: Possible mistake here: I don't know if the inner GLFW implementation is thread safe.
Can `pollEvents` really be called at the same time as `swapBuffers`? We should check that, and if not, we should make the methods require non-const access to the Window, which should prevent users from doing unsafe stuff like this.

engine/src/cubos/engine/plugins/window.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+void cubos::engine::plugins::windowPlugin(Cubos& cubos)
+{
+    cubos
+        .addResource<cubos::core::io::Window>()
+
+        .addStartupSystem(startup, "OpenWindow")
+
+        .addSystem(pollSystem, "Poll")
+        .putStageAfter("Main", "Poll")
```
RiscadoA, line 38: I don't think you need to specify 'Main' here, lets just believe that the users of this plugin tie everything together for now. The future dispatcher will solve this problem for us.

engine/include/cubos/engine/plugins/window.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include <cubos/core/io/window.hpp>
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins
+{
+    void windowPlugin(Cubos& cubos);
```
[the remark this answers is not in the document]
rsubtil: You mean code comments explaining this plugin, or a page in the documentation?

engine/src/cubos/engine/plugins/window.cpp
rsubtil: Supposedly swap buffer only happens after polling and after all user code and engine rendering happens.

engine/include/cubos/engine/plugins/window.hpp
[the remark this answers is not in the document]
RiscadoA: In the code, for now at least.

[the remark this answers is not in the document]
rsubtil: Done, let me know if you like how I structured it or if I should change it.

engine/src/cubos/engine/plugins/window.cpp
rsubtil: Fixed

engine/src/cubos/engine/plugins/window.cpp
RiscadoA: Okay we need to take that into account when adding multi threading to the dispatcher.
Maybe we should add a way to force a system to run on a specific thread?

RiscadoA: #241

rsubtil: > Okay we need to take that into account when adding multi threading to the dispatcher. Maybe we should add a way to force a system to run on a specific thread?

Seems like we have no option. [This post](https://discourse.glfw.org/t/multithreading-glfw/573) mentions how "the OS only posts events to the thread where the window was created so polling needs to happen on that thread". So it's another config required for the dispatcher.

Issue #226, opened 2022-11-30
Simplify and modernize engine samples


RiscadoA:
- [ ] Simplify/remove/create samples.
- [x] Restructure the samples so that each one has its folder.

The engine samples are outdated and contain lots of duplicated code.
With the introduction of #197, it will be possible to shorten the engine samples a lot.
Some of them, (e.g., `asset_renderer`) are also not necessary anymore.

The engine samples should also reflect a real project structure. Each should be
contained within a folder, in which there may be, optionally,  an `assets` folder and a
`components` folder.

RiscadoA:
Closed in favor of #459.

PR #244, opened 2022-12-07
Restructure engine samples into directories

RiscadoA:
Part of #226 

- Makes engine samples directories instead of single source files, which allows them to have their own assets/components folders.
- Fix cubinhos generate module bug which prevented two targets in the same binary dir from both defining their own components.
- Modernize the scene sample:
  - use lower case component IDs
  - use component code gen syntax
  - remove obsolete code
  - use debug query to print world debug info
- Add `core::initializeLogger` to `Cubos` constructor to make the game code cleaner.


RiscadoA:
Further simplification on samples like `cars` depends on having plugins like `windowPlugin` and `rendererPlugin` merged, so it will be left to a future PR.

rsubtil (approved):
Looks good to me!

PR #251, opened 2022-12-12
Remove Debug query, allow direct access to World

### LINKED ISSUE #245: Allow systems to access the World
For the PR #221, I implemented a special kind of query called `Debug` which allows us to access all components of an entity.
I now see this was stupid. For the UI we will also need to create new entities, destroys old entities and insert and remove components (#222), so, `Debug` is not enough.

Instead, `Debug` should be removed entirely and it should be possible to create special systems which access the ECS `World` directly:

```cpp
void specialSystem(World& world)
{
  // i can do whatever i want >:)
}
```

This gives us way more flexibility for this kind of use case. It should be documented that its something to avoid and only use when really necessary, since it prevents us from executing other systems in parallel and also forces us to apply commands.

RiscadoA:
Closes #245


core/include/cubos/core/ecs/system.hpp
```diff
@@ -136,16 +135,16 @@ namespace cubos::core::ecs
...
             /// @returns The query.
-            static Type arg(const World& world, Type&& fetched);
+            static Type arg(Type&& fetched);
         };
 
-        template <> struct SystemFetcher<Debug>
+        template <> struct SystemFetcher<World&>
         {
-            using Type = Debug;
+            using Type = World*;
```
rsubtil, line 147: Is `Type` still used?

RiscadoA: Yeah, its used in the fetcher for tuples to determine its `Type`.
```cpp
        template <typename... Args> struct SystemFetcher<std::tuple<Args...>>
        {
            using Type = std::tuple<typename SystemFetcher<Args>::Type...>;
            ...
```

rsubtil (approved):
Looks good to me!

PR #253, opened 2022-12-12
[WIP] environment settings plugin

roby2014:


roby2014:
I feel like `envSettingsPlugin` should have more documentation, like it depends on `Arguments` resource which is only added if you use the `Cubos(int argc, char** argv) ` constructor, but not sure how to "word" this.

In rest, I think it looks OK, but I'm all ears for suggestions.

RiscadoA:
Yeah, you can specify something along the lines of 'Using this plugin requires that the `Arguments` resource has been added. For this to happen, the command line arguments `argc` and `argv` from the `main` function should be provided to `Cubos`'.

engine/include/cubos/engine/cubos.hpp
```diff
@@ -15,12 +15,19 @@ namespace cubos::engine
         bool value;
     };
 
+    /// Resource used for storing program arguments
```
RiscadoA, line 18: Add '.' at the end of the comment.

engine/include/cubos/engine/plugins/env_settings.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef CUBOS_ENGINE_PLUGINS_ENV_SETTINGS_HPP
+#define CUBOS_ENGINE_PLUGINS_ENV_SETTINGS_HPP
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins
+{
+    /// Plugin to handle environment settings, provided via command line.
```
RiscadoA, line 10: Explain that the `Arguments` resource needs to have been added, otherwise this will crash, and that it is added when argc, argv are passed to `Cubos` ctor

RiscadoA (changes requested):

RiscadoA (approved):

RiscadoA:
LGTM!

PR #254, opened 2022-12-12
Add renderer plugin

### LINKED ISSUE #238: Add Renderer plugin
Split from #197

- Add the `transformPlugin`, which registers the transform related components (Position, Rotation, Scale, LocalToWorld) and adds the `transformSystem` (already implemented, see `engine/include/cubos/engine/ecs/transform_system.hpp`).
- Add the `rendererPlugin`, which adds the `transformPlugin`, registers the `Grid` component and the resource `Renderer`, and adds the `drawSystem` (also already implemented at `engine/include/cubos/engine/ecs/draw_system.hpp`).

The existing code should be moved to the respective plugin file's, as was done with the window plugin (e.g., move the `drawSystem` definition to `renderer_plugin.cpp`).

DiogoMendonc-a:
closes #238.


engine/include/cubos/engine/plugins/renderer.hpp
```diff
@@ -0,0 +1,31 @@
...
+    ///
+    /// Stages:
+    /// - draw: Draws component's to the renderer's frame.
+    /// @param cubos CUBOS. main class
+    void rendererPlugin(Cubos& cubos);
+
+    /// Resource to contain the current active camera
+    struct ActiveCamera
+    {
+        cubos::core::ecs::Entity entity;
```
rsubtil, line 27: Shouldn't this be a `Camera` component? I believe holding entity references violates the core principle of ECS, so the code that uses this should be changed to work with a direct `Camera` component instead.

engine/include/cubos/engine/plugins/renderer.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <glm/glm.hpp>
+#include <cubos/core/gl/render_device.hpp>
+#include <cubos/core/ecs/entity_manager.hpp>
+
+namespace cubos::engine::plugins
+{
+    /// Plugin to create and use a renderer.
+    ///
+    /// Startup Stages:
```
rsubtil, line 16: Rename to tags (and fix other cases in `transform.hpp` too)
```suggestion
    /// Startup Tags:
```

engine/samples/scene/main.cpp
```diff
@@ -54,6 +70,51 @@ void printStuff(World& world)
...
+
+    auto floorRendererGrid = renderer->upload(floorGrid);
+    auto floor = assetManager.store<data::Grid>("floor", data::Usage::Static,
+                                                data::Grid{
+                                                    .grid = std::move(floorGrid),
+                                                    .rendererGrid = floorRendererGrid,
+                                                });
+
+    world.create(ecs::Grid{floor, {-128.0f, -1.0f, -128.0f}}, ecs::LocalToWorld{}, ecs::Position{}, ecs::Scale{4.0f});
+} //
```
rsubtil, line 107: Extra //
```suggestion
}
```

engine/src/cubos/engine/plugins/renderer.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <cubos/engine/gl/deferred/renderer.hpp>
+#include <cubos/engine/gl/renderer.hpp>
+#include <cubos/engine/gl/frame.hpp>
+
+#include <cubos/core/gl/camera.hpp>
+
+#include <cubos/core/settings.hpp>
+
+static void startup(gl::Renderer& renderer, const cubos::core::io::Window& window)
```
rsubtil, line 17: This is using a blank settings object. You should use the existing one:
```suggestion
static void startup(gl::Renderer& renderer, const cubos::core::io::Window& window, const cubos::core::Settings& settings)
{
    auto& renderDevice = window->getRenderDevice();
    renderer =
        std::make_shared<gl::deferred::Renderer>(renderDevice, window->getFramebufferSize(), settings);
}
```

rsubtil (changes requested):
Good job! Minus a few nitpicks, there are some things that should be addressed though

engine/include/cubos/engine/plugins/renderer.hpp
RiscadoA: I think this is fine.
The `ActiveCamera` is a resource which identifies the *entity* which is designated as the active camera, with which we should draw to the screen.
It doesn't store the camera state, it only refers to the entity which stores it.
This makes sense because we can have multiple cameras in a scene and switch between them easily, without removing/adding `Camera` components from them.

engine/samples/scene/main.cpp
```diff
@@ -54,6 +70,51 @@ void printStuff(World& world)
...
+
+static void TurnOnLight(gl::Frame& frame)
```
RiscadoA, line 109:
```suggestion
static void turnOnLight(gl::Frame& frame)
```

engine/src/cubos/engine/plugins/transform.cpp
```diff
@@ -0,0 +1,19 @@
+#include <cubos/engine/plugins/transform.hpp>
+#include <cubos/engine/ecs/transform_system.hpp>
```
RiscadoA, line 2: Same as `cubos/engine/ecs/draw_system.hpp` but for `cubos/engine/ecs/transform_system.hpp`.

engine/src/cubos/engine/plugins/transform.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <components/cubos/local_to_world.hpp>
+
+void cubos::engine::plugins::transformPlugin(Cubos& cubos)
+{
+    cubos.addComponent<ecs::Position>()
+        .addComponent<ecs::Rotation>()
+        .addComponent<ecs::Scale>()
+        .addComponent<ecs::LocalToWorld>();
+
+    cubos.system(ecs::transformSystem).tagged("PrepareDraw");
```
RiscadoA, line 18: Change the tag to `"Transform"`, its best to leave this plugin renderer-agnostic.

engine/src/cubos/engine/plugins/renderer.cpp
```diff
@@ -0,0 +1,57 @@
+#include <cubos/engine/plugins/renderer.hpp>
+#include <cubos/engine/plugins/transform.hpp>
+#include <cubos/engine/plugins/window.hpp>
+
+#include <cubos/engine/ecs/draw_system.hpp>
```
RiscadoA, line 5: I think its best to remove `cubos/engine/ecs/draw_system.hpp` entirely and move the system definition here, like you did with the `draw` system (maybe rename `draw` to `render`). This way, the plugin implementation is hidden in its source file.

engine/src/cubos/engine/plugins/renderer.cpp
```diff
@@ -0,0 +1,57 @@
...
+        .addPlugin(windowPlugin)
+        .addComponent<ecs::Grid>()
+        .addComponent<ecs::Camera>()
+        .addResource<gl::Frame>()
+        .addResource<gl::Renderer>()
+        .addResource<ActiveCamera>();
+
+    cubos.startupTag("SetRenderer").afterTag("OpenWindow");
+    cubos.startupSystem(startup).tagged("SetRenderer");
+    cubos.system(ecs::drawSystem).tagged("CreateDrawList");
```
RiscadoA, line 55:
```suggestion
    cubos.system(ecs::drawSystem).tagged("CreateDrawList").afterTag("Transform");
```

engine/samples/scene/main.cpp
```diff
@@ -28,17 +42,19 @@ using namespace core::data;
...
+    assetManager.registerType<data::Palette>();
     assetManager.registerType<data::Scene>();
     assetManager.importMeta(FileSystem::find("/assets/"));
 
     // Load and spawn the scene.
     auto scene = assetManager.load<data::Scene>("scenes/main");
-    auto root = cmds.create().entity();
-    cmds.spawn(scene->blueprint).add("main", Parent{root});
+    // auto root = world.create().entity();
+    // cmds.spawn(scene->blueprint).add("main", Parent{root});
```
RiscadoA, line 57: Why is this commented?

engine/include/cubos/engine/plugins/transform.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins
+{
+    /// Plugin to create and use transforms.
+    ///
+    /// Stages:
```
RiscadoA, line 14: Update this to specify the tags used (`"Transform"`).

RiscadoA (changes requested):

engine/samples/scene/main.cpp
DiogoMendonc-a: Replacing the Commands with World made those two lines not compile. I just deleted them. If they are needed I'll need some help with changing them to use the World.

engine/src/cubos/engine/plugins/renderer.cpp
DiogoMendonc-a: Did that, but left the original `cubos/engine/ecs/draw_system.hpp` and `.cpp` files still in the repository, as there are other samples using them and I didn't want to break them.

engine/src/cubos/engine/plugins/transform.cpp
```diff
@@ -0,0 +1,34 @@
+#include <cubos/engine/plugins/transform.hpp>
+
+#include <cubos/core/settings.hpp>
+
+#include <components/cubos/position.hpp>
+#include <components/cubos/rotation.hpp>
+#include <components/cubos/scale.hpp>
+#include <components/cubos/local_to_world.hpp>
+
+void applyTransform(cubos::core::ecs::Query<cubos::engine::ecs::LocalToWorld&, const cubos::engine::ecs::Position*,
```
RiscadoA, line 10: Make this function static, to prevent linking with other functions with the same name in other files.

RiscadoA (changes requested):

RiscadoA (approved):

rsubtil (approved):

PR #255, opened 2022-12-12
Make AnySystemWrapper generic over return type

### LINKED ISSUE #203: Add system run conditions
Some systems should only run when a certain criteria is satisfied. Examples of this are:
- physics systems;
- rendering systems;
- game code related to specific modes (e.g.: a system called when the main menu is open).

One way to solve this is by defining two new concepts.
- Condition System: systems which return a boolean instead of void.
- Stage Groups: sequences of stages which are only executed if a condition system returns true , which is executed at the start of the group.

The syntax could look like:
```cpp
cubos
  .onGroup("Physics") // from now on all calls are relative to the 'Physics' group.
  .addCondition(systemWhichReturnsABoolean)
  .addSystem(integrate, "Integrate")
  .addSystem(solveCollisions, "Solver")
  .putStageBefore(...)
  .noGroup() // we're back to the global scope
  .addSystem(mySystem, "MyStage")
  .putStageBefore("MyStage", "Physics"); // run 'MyStage' stage before the group 'Physics'.
```

RiscadoA:
Closes #203.

Allows systems to have return types other than `void`.
Now, `AnySystemWrapper` has a template argument to specify the return type of the wrapped systems, so, in the dispatcher, normal systems will be wrapped in `AnySystemWrapper<void>` and condition systems will be wrapped in `AnySystemWrapper<bool>`.
Condition systems will then be implemented for #241.


rsubtil (approved):
Looks good to me!

PR #256, opened 2022-12-12
Add ThreadPool and tests for it

### LINKED ISSUE #248: Add Thread pool
Will be useful to introduce multithreading on #241

RiscadoA:
Closes #248.

### DIFF
--- a/core/src/cubos/core/thread_pool.cpp
+++ b/core/src/cubos/core/thread_pool.cpp
@@ -1,0 +1,63 @@
+#include <cubos/core/thread_pool.hpp>
+
+using namespace cubos::core;
+
+ThreadPool::ThreadPool(size_t numThreads)
+{
+    this->threads.reserve(numThreads);
+    this->numTasks = 0;
+    this->stop = false;
+
+    for (size_t i = 0; i < numThreads; i++)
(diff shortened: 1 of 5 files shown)


core/src/cubos/core/thread_pool.cpp
```diff
@@ -0,0 +1,63 @@
...
+                    this->new_task.wait(lock, [this]() { return this->stop || !this->tasks.empty(); });
+                    if (this->stop && this->tasks.empty())
+                        return; // Thread pool is being destroyed, and there are no more tasks to execute.
+                    task = std::move(this->tasks.front());
+                    this->tasks.pop_front();
+                    this->numTasks += 1; // A task is being executed.
+                }
+
+                task();
+                this->numTasks -= 1; // Task has finished executing.
```
rsubtil, line 31: This doesn't look like it's being protected with a lock.

RiscadoA: I doesn't need to, thats the async part.
If you're referring to the `numTasks` variable, its atomic so its okay to decrement it without the mutex.

rsubtil (approved):
Looks good to me!

PR #257, opened 2022-12-12
Refactor cubos and dispatcher stage system

rsubtil:
Refactor the cubos class and dispatcher to the new syntax proposed in #241. The "stages" system has been replaced with the "tags" system, allowing for more flexibility in system configuration.

The refactor is already taking into account multithreading, and is designed already for users to adapt to this paradigm. Systems execution order is now explicit. If no directive is given to a system/tag, their order is undefined. Right now the order is in reverse due to the topology algorithm, so this already gives an hint to developers when they forget to define execution order. When multithreading is introduced, the order will be arbitrary:
```cpp
/// Before
cubos.addSystem(a, "A").addSystem(b, "B").addSystem(c, "C");
// Systems would implicitely run in order: A -> B -> C

/// Now
cubos.system(a).tagged("A").system(b).tagged("B").system(c).tagged("C");
// Systems will run in reverse due to topological algorithm: C -> B -> A
// With multithreading, this execution order will become further erratic
```

Further work, such as implementing tag/setting inheritance and adding conditions/threading modes to settings will be done in other PRs as to not clump everything together.


RiscadoA:
By the way, we hadn't discussed yet and I'm not sure how you are planning to implement it, but I think a possible good solution to the separation of concerts between `Cubos` and `Dispatcher` is:
- `Dispatcher` provides exactly the same methods we planned for `Cubos` (excluding `startupSystem` of course).
- the only job of `Cubos` is to pass call the methods on the correct `Dispatcher` and to make sure that no tag is shared between startup and non-startup systems.

rsubtil:
> By the way, we hadn't discussed this yet and I'm not sure how you are planning to implement it, but I think a possible good solution to the separation of concerns between `Cubos` and `Dispatcher` is:
> 
>     * `Dispatcher` provides exactly the same methods we planned for `Cubos` (excluding `startupSystem` of course).
> 
>     * the only job of `Cubos` is to call the methods on the correct `Dispatcher` and to make sure that no tag is shared between startup and non-startup systems.
> 
> 
> What do you think?

Sure thing, but should the dispatcher handle tag/dependency resolution as well?

RiscadoA:
Yeah the idea was to move all of the dependency resolution to the dispatcher itself.

rsubtil:
The PR is ready for review

core/include/cubos/core/ecs/dispatcher.hpp
```diff
@@ -9,81 +9,198 @@
...
+
+    template <typename F> void Dispatcher::systemSetAfterSystem(F func)
     {
-        // Register stage if it doesn't exist.
-        if (this->stagesByName.find(stage) == this->stagesByName.end())
+        auto it = std::find_if(pendingSystems.begin(), pendingSystems.end(), [&func](const System& system) {
+            SystemWrapper<F>* wrapper = dynamic_cast<SystemWrapper<F>*>(system.system.get());
+            if (!wrapper)
+                return false;
+            return wrapper->system == func;
```
RiscadoA, line 168: Can we do this? I didn't know we could compare two different function types like this. Does it work with lambdas?
If it doesn't I'm fine with forcing systems to be actual function pointers.

core/include/cubos/core/ecs/dispatcher.hpp
```diff
@@ -9,81 +9,198 @@
...
+        struct SystemSettings;
+        struct System;
+
+        /// Internal class to specify system dependencies
+        struct Dependency
         {
-            Before,
-            After
+            std::vector<std::string> tag;
+            std::vector<System*> system;
```
RiscadoA, line 45: I think maybe its safer to refer to systems by their index on the vector instead of by a pointer.

core/include/cubos/core/ecs/dispatcher.hpp
```diff
@@ -9,81 +9,198 @@
...
+        /// @param world World to call the systems in.
+        /// @param cmds Command buffer.
+        void callSystems(World& world, CommandBuffer& cmds);
+
+        /// Visits a DFSNode to create a topological order.
+        /// This is used internally during call chain compilation.
+        /// @param node The node to visit.
+        /// @param nodes Array of DFSNodes.
+        /// @return True if a cycle was detected, false if otherwise.
+        bool dfsVisit(DFSNode& node, std::vector<DFSNode>& nodes);
```
RiscadoA, line 141: Make this private

engine/include/cubos/engine/cubos.hpp
```diff
@@ -49,44 +50,61 @@ namespace cubos::engine
         /// @return Reference to this object, for chaining.
         template <typename C> Cubos& addComponent();
 
-        /// Adds a new system to the engine, which will be executed every iteration of the main loop.
-        /// If the stage doesn't already exist, it is created in the previously specified default position.
+        /// Sets the current tag for the main dispatcher. If the tag doesn't exist, it will be created.
+        /// Subsequent calls will set this tag's settings.
+        /// @param tag The tag to set.
+        Cubos& tag(const std::string& tag);
```
RiscadoA, line 56: Maybe instead of returning `Cubos` we should return a `TagBuilder` (which keeps a reference to `Cubos`), for which we can then call the property methods (`tagged`, `after`, `before`, ...). This way, we prevent some possible errors at compile time.

The same goes for the systems (e.g, returning a SystemBuilder).

engine/include/cubos/engine/cubos.hpp
```diff
@@ -49,44 +50,61 @@ namespace cubos::engine
...
+
+        /// Sets the current system to run before the tag.
+        /// If the specified tag doesn't exist, it is internally created.
+        /// @param tag The tag to run before.
+        Cubos& beforeTag(const std::string& tag);
+
+        /// Sets the current system to run before a given system.
+        /// The specified system must exist.
+        /// @param system The system to run before.
+        template <typename F> Cubos& before(F system);
```
RiscadoA, line 96: Prolly a good idea to change it to `beforeSystem` (the same goes for other similar methods).

engine/include/cubos/engine/cubos.hpp
```diff
@@ -49,44 +50,61 @@ namespace cubos::engine
...
 
         /// Runs the engine.
         void run();
 
     private:
+        bool isStartup;
+        bool isSystem;
```
RiscadoA, line 103: Won't need this if we use the builders.

engine/src/cubos/engine/cubos.cpp
```diff
@@ -67,6 +144,10 @@ void Cubos::run()
 {
     plugins.clear();
```
RiscadoA, line 145: Just noticed this, why are clearing the plugins?

RiscadoA (changes requested):

engine/src/cubos/engine/cubos.cpp
rsubtil: We only need to keep track of plugins to ensure they're not called multiple times. At this point (run), users cannot add anything more to the engine, so the plugin list is cleared to save space

engine/include/cubos/engine/cubos.hpp
rsubtil: How should users "escape" the scope of a builder? Should we have users save the Cubos object so they can access it later for other operations?

```cpp
Cubos c;
c.tag("A").before("B"); // TagBuilder
c.system(a).tagged("A"); // SystemBuilder
c.run(); // Cubos
```

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: So far it seems to work yeah, although I haven't tested system dependencies extensively. When I get the chance I'll confirm if the current sample works.

As for lambdas I didn't even test them tbh, I'll also try them.

engine/include/cubos/engine/cubos.hpp
RiscadoA: Yeah, I think this is fine. On plugins we already have the object being passed as argument, so its not an issue there.

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: This isn't reliable as systems are moved from the `pendingSystems` to `systems`. Also this dependency class effectively "disappears" once the call chain is built; the dependency order is implicit with the order systems are in the `systems` array, so these values aren't used.

I did forget about this fact and didn't delete them though, so I'll change that.

engine/include/cubos/engine/cubos.hpp
rsubtil: Does this still apply with the builder approach?

RiscadoA: Yeah, what I mean was that it may the names `afterTag` vs `after` may get confusing since `after` takes a templated argument. So, I think it would be more explicit to rename `after` to `afterSystem`.

rsubtil: Ah, yes I agree

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: Yup, I just checked. It even works with lambdas, which is awesome.

rsubtil: I've added system ordering on the example as well to demonstrate.

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: I went ahead and also made the new internal classes private as well.

engine/include/cubos/engine/cubos.hpp
rsubtil: Implemented

engine/include/cubos/engine/cubos.hpp
rsubtil: Implemented

core/include/cubos/core/ecs/dispatcher.hpp
```diff
@@ -9,81 +9,199 @@
...
 namespace cubos::core::ecs
 {
     /// @brief Dispatcher is a class that is used to register systems and call them all at once, where they will be
     /// executed in order of the stages they are in.
     class Dispatcher
     {
-    public:
-        // Direction enum
-        enum class Direction
+    private:
```
RiscadoA, line 36: Move private below public (merge two private sections).

core/include/cubos/core/ecs/dispatcher.hpp
```diff
@@ -9,81 +9,199 @@
 #include <string>
 #include <memory>
 
+#define ENSURE_CURR_SYSTEM()                                                                                           \
```
RiscadoA, line 12: Wrap these macros with `do {...} while(false)` to force using `;` after it.

engine/include/cubos/engine/cubos.hpp
```diff
@@ -21,6 +22,56 @@ namespace cubos::engine
         const std::vector<std::string> value;
     };
 
+    // Used to chain configurations related to tags
```
RiscadoA, line 25: Use triple /

engine/include/cubos/engine/cubos.hpp
```diff
@@ -21,6 +22,56 @@ namespace cubos::engine
...
+
+        /// Sets the current tag to be executed after another tag.
+        /// @param tag The tag to be executed after.
+        TagBuilder& afterTag(const std::string& tag);
+
+    private:
+        core::ecs::Dispatcher& dispatcher;
+    };
+
+    // Used to chain configurations related to systems
```
RiscadoA, line 44: Use triple /

RiscadoA (changes requested):

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: Fixed

core/include/cubos/core/ecs/dispatcher.hpp
rsubtil: Fixed

engine/include/cubos/engine/cubos.hpp

engine/include/cubos/engine/cubos.hpp

PR #279, opened 2022-12-15
Add ImGui plugin

roby2014:


engine/src/cubos/engine/plugins/imgui.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+static void beginImGuiFrame()
+{
+    cubos::core::ui::beginFrame();
+}
+
+static void endImGuiFrame()
+{
+    cubos::core::ui::endFrame();
+}
```
RiscadoA, line 18: You can just pass `cubos::core::ui::initialize`, `cubos::core::ui::beginFrame` and `cubos::core::ui::endFrame()` directly to `addSystem`, since they have a valid system signature.

engine/samples/scene/main.cpp
```diff
@@ -58,5 +67,11 @@ int main(int argc, char** argv)
 
         .addPlugin(cubos::engine::plugins::envSettingsPlugin)
```
RiscadoA, line 68: You can add `using namespace cubos::engine::plugins` to make this less verbose.

RiscadoA (changes requested):

engine/src/cubos/engine/plugins/imgui.cpp
roby2014: Not sure how I could call pass the window resource to `cubos::core::ui::initialize`

RiscadoA: Okay true, that one does need to stay like it is right now, but you can do it for begin and end frame.

roby2014: Not possible because `cubos::core::ui::endFrame` receives a `cubos::core::gl::Framebuffer` argument.
In order to maintain consistency, I'm letting them as they are.

RiscadoA (approved):

PR #280, opened 2022-12-15
Add Settings Inspector plugin

roby2014:
Depends on #279


roby2014:
UI needs polishing, but the "inspect" part already works.

engine/include/cubos/engine/plugins/settings_inspector.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins
+{
+    /// Plugin that allows inspecting the current settings
+    /// @param cubos CUBOS. main class
+    void settingsInspectorPlugin(Cubos& cubos);
```
RiscadoA, line 12: I would just move this into another namespace and directory, maybe `plugins::tools`? Not sure what we would name it.

RiscadoA (changes requested):

RiscadoA (approved):

Issue #259, opened 2022-12-13
Format functions and templates on different lines


roby2014:
At the moment, the `clang-format` settings are one lining templates + function signatures, like this:
```cpp
template <typename T, std::size_t M> const T& EventReader<T, M>::Iterator::operator*() const
{
    return pipe.getEvent(index);
}
```
even tho in some cases it looks good, it does not in others 😄 

IMO, separating would be way more readable and easier to look at code, so using the previous example, it would look like this:
```cpp
template <typename T, std::size_t M>
const T& EventReader<T, M>::Iterator::operator*() const
{
    return pipe.getEvent(index);
}
```

roby2014:
This can be done automatically after #219 is working.

roby2014:
PR #260 (issue #219) already fixes this issue. Closing.

PR #260, opened 2022-12-14
Format code automatically using GitHub Actions

roby2014:
This PR should add a github action that runs on every PR review approve, formatting all the code the user added to the branch.

PR #281, opened 2022-12-16
[WIP] Format function signatures and templates on different lines 

roby2014:
Added `AlwaysBreakTemplateDeclarations` to `.clang-format`
This setting allows to format template definitions on a separate line from function signatures, see #259.

The **whole source code needs to be updated**, not sure if we do that on a separate commit or we finish #219 and then the job is done via the action ... (?)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,5 +6,5 @@ IndentWidth: 4
 AccessModifierOffset: -4
 DerivePointerAlignment: false
 PointerAlignment: Left
-
+AlwaysBreakTemplateDeclarations: Yes
 ...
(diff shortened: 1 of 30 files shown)


RiscadoA:
You can add a commit formatting the code to this PR, I think it will reduce the number of conflicts we would get if we went the #219 way.

roby2014:
For further reference, this is what I used.

```sh
SRC="$(find ./core/include ./core/samples ./core/src ./core/tests ./engine ./tools | grep -E ".*(\.cpp|\.c|\.h|\.hpp)$")"
clang-format -i --style=file $SRC
```

RiscadoA:
On my machine `clang-format` formats the failing line (`asset_manager.hpp:67`) to `private :` instead of `private:`. Turns out the check engine format action also does the same.

Maybe this is a issue with the clang-format version I'm using/the action is using, and you're using a more recent one?

roby2014:
#260 Fullfills this PR job. Closing.

PR #282, opened 2022-12-16
Make system argument commands syntax consistent with query

### LINKED ISSUE #250: Make system argument Commands syntax consistent with Query
Currently, the syntax for systems is like:
```cpp
void mySystem(Query<...> myQuery, T& writeRsc, const U& readRsc)
{
  // ...
}
```

Where each reference identifies access to a resource, and `Query` is passed by value.
This is inconsistent with the `Commands` argument, which is passed as `Commands&`, as if it was a resource.
This may be confusing, so `Commands` should be passed by value.

RiscadoA:
Closes #250


rsubtil (approved):
Looks good to me!

PR #287, opened 2022-12-21
Fixed incompatible vector iterators bug

roby2014:


core/src/cubos/core/ecs/dispatcher.cpp
```diff
@@ -70,10 +70,7 @@ void Dispatcher::putStage(std::string stage, std::string referenceStage, Directi
     {
         this->stagesOrder.erase(it);
         // Erase invalidates next iterators, so refIt may need to be updated.
```
RiscadoA, line 72:
```suggestion
        // Erase invalidates next iterators, so refIt needs to to be updated.
```

RiscadoA (changes requested):

rsubtil (approved):
Looks good to me!

RiscadoA (approved):

Issue #202, opened 2022-11-08
Add EventWriter/EventReader


RiscadoA:
Add the classes `EventWriter<T>` and `EventReader<T, Mask>` which can be used as ECS resources for communication between systems.
The `Mask` will be an optional type parameter for filtering the events received. This could be used, for example, in the collision engine.
The collider components can hold a `mask` field, and then when a collision event is generated, both of the masks are merged.
This way the game developer can easily filter events related to certain areas of the game (player, bullets, etc).

Example syntax:
```cpp
struct Collision
{
  // ...
};

enum MyMask
{
  None = 0,
  Player = 1 << 0,
  Bullet = 1 << 1,
  All = 1 << 2 - 1
};

void genEvents(EventWriter<Collision> ev)
{
  ev.write(Collision {/* ... */}); // unmasked component
  ev.write(Collision {/* ... */}, Player); // masked component
}

void checkEvents(EventReader<Collision, Player> ev)
{
  for (auto& e : ev)
  {
    // process event
  }
}
```

Implementation:
- [x] Define `EventPipe<T>` class.
- [x] Define `EventWriter<T>` and `EventReader<T, Mask>`.
- [x] #298

Internally, both the writer and reader store mutable and const references, respectively, to an event pipe, where the events are stored. The pipe should provide a clear method which destroys all events on the pipe. This method will then be called every iteration of the game loop.

PR #297, opened 2023-01-14
Change window to an event loop instead of callback driven

### LINKED ISSUE #284: Change Window to an event loop instead of callback driven

RiscadoA:
Closes #284.

Merging this PR effectively breaks all `ImGui` plugin samples - this is because #202 still isn't merged and we have no way to communicate data from the window plugin to the `ImGui` plugin.
It also disables all code which touches the `InputManager`, as that part of the code needs major refactoring.
**This includes the car sample** - is it to soon to remove it entirely?


RiscadoA:
Looks like the Windows build is failing due to a MSVC error.
I think it may be related to [this](https://stackoverflow.com/questions/71738552/c3520-parameter-pack-must-be-expanded-incorrect-behaviour-for-variadic-using), and it may have already been fixed.
Should we try to use a more recent compiler version?

core/src/cubos/core/io/glfw_window.cpp
```diff
@@ -268,49 +268,41 @@ static void keyCallback(GLFWwindow* window, int key, int scancode, int action, i
...
-        handler->onMouseUp.fire(cubosButton);
+    if (action == GLFW_PRESS || action == GLFW_RELEASE)
+        handler->pushEvent(MouseButtonEvent{.button = cubosButton, .pressed = action == GLFW_PRESS});
 }
 
 static void scrollCallback(GLFWwindow* window, double xoffset, double yoffset)
 {
     GLFWWindow* handler = (GLFWWindow*)glfwGetWindowUserPointer(window);
-    handler->onMouseScroll.fire(glm::ivec2(xoffset, yoffset));
+    handler->pushEvent(MouseScrollEvent{.offset = glm::vec2(xoffset, yoffset)});
```
rsubtil, line 293: This changed from `ivec2` to `vec2`

rsubtil (approved):
Looks good to me!

PR #299, opened 2023-01-14
Make `EventWriter<T>` and `EventReader<T, Mask>` possible system arguments

### LINKED ISSUE #298: Make `EventWriter<T>` and `EventReader<T, Mask>` possible system arguments.

RiscadoA:
Closes #298.
Depends on #258.

- Allows `EventReader<T>` and `EventWriter<T>` to be used as system arguments.
- Adds the `.addEvent<T>` method to the `Cubos` class, which must be called before any system uses an event type.
- Adds the engine `events` sample which shows how events can be used and how they behave.


core/include/cubos/core/ecs/event_reader.hpp
```diff
@@ -0,0 +1,129 @@
...
+    template <typename T, unsigned int M> std::optional<std::reference_wrapper<const T>> EventReader<T, M>::read()
+    {
+        while (this->index < this->pipe.sentEvents())
+        {
+            std::pair<const T&, unsigned int> p = pipe.get(this->index++);
+            auto& event = p.first;
+            auto mask = p.second;
+            if (matchesMask(mask))
+            {
+                // return std::optional<std::reference_wrapper<const T>>(std::reference_wrapper<const T>(event));
```
rsubtil, line 68: Remove commented line

core/include/cubos/core/ecs/event_reader.hpp
```diff
@@ -0,0 +1,129 @@
...
+                return std::make_optional(std::reference_wrapper<const T>(event));
+            }
+        }
+
+        return std::nullopt;
+    }
+
+    template <typename T, unsigned int M> bool EventReader<T, M>::matchesMask(decltype(M) eventMask) const
+    {
+        return M == ~0 ? true : eventMask & M;
```
rsubtil, line 78: This should just be simplified to `eventMask & M`.
```suggestion
        return eventMask & M;
```

core/include/cubos/core/ecs/event_reader.hpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    template <typename T, unsigned int M> typename EventReader<T, M>::Iterator EventReader<T, M>::begin()
+    {
+        // return a new begin iterator only if we did not read all events yet(?)
+        return (this->index >= this->pipe.sentEvents()) ? this->end() : Iterator(*this, this->read(), false);
```
rsubtil, line 84: This ternary operator is a big difficult to read, I'd refactor into an if/else
```suggestion
        if (this->index >= this->pipe.sentEvents())
                return this->end();
        return Iterator(*this, this->read(), false);
```

core/include/cubos/core/ecs/system.hpp
```diff
@@ -77,108 +84,89 @@ namespace cubos::core::ecs
...
-
-            /// @param world The world to fetch from.
-            /// @param commands The commands object.
-            /// @returns The requested resource write lock.
-            static WriteResource<R> fetch(World& world, CommandBuffer& commands);
-
-            /// @param lock The resource lock to get the reference from.
-            /// @returns The requested resource reference.
+            static State prepare(World& world);
+            static WriteResource<R> fetch(World& world, CommandBuffer& commands, State& state);
```
rsubtil, line 91: Use the `Type` alias set previously on all applicable cases
```suggestion
            static Type fetch(World& world, CommandBuffer& commands, State& state);
```

core/samples/event_system.cpp
```diff
@@ -0,0 +1,85 @@
...
+    writer.push(MyEvent{.data = 4}, MyEvent::Mask::KEY_EVENT);
+    writer.push(MyEvent{.data = 1}, MyEvent::Mask::WHEEL_EVENT);
+    writer.push(MyEvent{.data = 2}, MyEvent::Mask::MOUSE_EVENT);
+    writer.push(MyEvent{.data = 6}, MyEvent::Mask::MOUSE_EVENT);
+    writer.push(MyEvent{.data = 15});
+    writer.push(MyEvent{.data = 11});
+
+    std::size_t index = 0;
+
+    printf("\n\n### mouse events using .read():");
```
rsubtil, line 39: *printf moment* :slightly_smiling_face:

core/samples/event_system.cpp
```diff
@@ -0,0 +1,85 @@
...
+    while (true)
+    {
+        static auto mouseReader = EventReader<MyEvent, MyEvent::Mask::MOUSE_EVENT>(pipe, index);
```
rsubtil, line 42: Can this be moved outside the loop so it's not redefined every time?

core/src/cubos/core/ecs/dispatcher.cpp
```diff
@@ -186,6 +186,15 @@ bool Dispatcher::dfsVisit(DFSNode& node, std::vector<DFSNode>& nodes)
 
 void Dispatcher::callSystems(World& world, CommandBuffer& cmds)
 {
+    // If the systems haven't been prepared yet, do so now.
+    // We can't multi-thread this as the systems require exclusive access to the world to prepare.
+    if (!this->prepared)
```
rsubtil, line 191: Since this runs only one, we could do this in a startup system instead in the future.

core/tests/test_event_system.cpp
```diff
@@ -0,0 +1,124 @@
...
+    pipe.addReader();
+    for (const auto& ev : keyEventReader)
+    {
+        // should run only once
+        EXPECT_EQ(ev.data, 1);
+    }
+    for (const auto& ev : keyEventReader)
+    {
+        // this should NOT run, because all KEY_EVENTS from keyEventReader already got read
+        EXPECT_EQ(false, true);
```
rsubtil, line 121: In this case I think [FAIL](https://google.github.io/googletest/reference/assertions.html#FAIL) is more appropriate
```suggestion
        FAIL();
```

rsubtil (changes requested):
Overall looks good, there's just some details to handle.

core/include/cubos/core/ecs/event_reader.hpp
RiscadoA: Not the same because events with mask 0 should be read by readers with mask ~0.
Still, it's easier to understand if changed to `M == ~0 || (eventMask & M)`

core/samples/event_system.cpp
RiscadoA: I think the point was that you could destruct and it and construct it again and it wouldn't make any difference.
(The static shouldn' t be there then).

core/src/cubos/core/ecs/dispatcher.cpp
RiscadoA: Hm, even startup systems may have state on their arguments, so maybe its better to keep the logic separated.
This way its also easy to enforce an order on the setup of that state.

rsubtil (approved):
Looks good to me!

rsubtil (approved):

PR #302, opened 2023-01-14
Switch macOS CI to use LLVM/clang

### LINKED ISSUE #293: Switch macOS CI compiler to Clang
I propose we switch the compiler of the macOS builds from `gcc` to `clang`. From my experience from CI, `clang` tends to be less painful to deal with. Plus, we prevent further scenarios where our code is only compiling in `gcc` by enforcing at all times it also works for `clang`.

In the case of #240, `gcc` doesn't seem to understand that some source code is actually Objective-C and thus fails to compile. Switching to `clang` fixed [the problem](https://github.com/rsubtil/cubos/actions/runs/3725927281/jobs/6319031966)

rsubtil:
Hopefully it fixes the current issues and generates less problems from now on.

Closes #293

### DIFF
--- a/engine/include/cubos/engine/cubos.hpp
+++ b/engine/include/cubos/engine/cubos.hpp
@@ -13,14 +13,14 @@ namespace cubos::engine
     /// Resource used as a flag to indicate whether the engine should stop running.
     struct ShouldQuit
     {
-        ShouldQuit(bool value) : value(value) {}
+        ShouldQuit(bool value);
         bool value;
     };
 
     /// Resource used for storing program arguments.
     struct Arguments
     {
-        Arguments(const std::vector<std::string>& value) : value(value) {}
+        Arguments(const std::vector<std::string>& value);
         const std::vector<std::string> value;
     };
 
--- a/engine/include/cubos/engine/cubos.hpp
+++ b/engine/include/cubos/engine/cubos.hpp
@@ -13,12 +13,14 @@ namespace cubos::engine
     /// Resource used as a flag to indicate whether the engine should stop running.
     struct ShouldQuit
     {
+        ShouldQuit(bool value) : value(value) {}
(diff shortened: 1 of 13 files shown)


rsubtil:
Had to fix some issues in `cubinhos`, but they were long overdue anyways. I'm still not 100% on this, but it seems that defining the same `Options` struct on different files would confuse the program as to what's their intended destructor (ASAN reported errors because `Options` was constructed on `generate.cpp` and then deleted on `embed.cpp` for example). Giving them unique names per file seems to have fixed it.

engine/include/cubos/engine/cubos.hpp
```diff
@@ -13,12 +13,14 @@ namespace cubos::engine
     /// Resource used as a flag to indicate whether the engine should stop running.
     struct ShouldQuit
     {
+        ShouldQuit(bool value) : value(value) {}
```
RiscadoA, line 16: This needs to be inlined or moved to the `.cpp` file.

engine/include/cubos/engine/cubos.hpp
```diff
@@ -13,12 +13,14 @@ namespace cubos::engine
...
         bool value;
     };
 
     /// Resource used for storing program arguments.
     struct Arguments
     {
+        Arguments(const std::vector<std::string>& value) : value(value) {}
```
RiscadoA, line 23: This needs to be inlined or moved to the `.cpp` file.

RiscadoA (changes requested):

engine/include/cubos/engine/cubos.hpp

engine/include/cubos/engine/cubos.hpp

RiscadoA (dismissed):

RiscadoA (approved):

PR #306, opened 2023-01-22
Fix macOS CI compilation

rsubtil:
macOS image [was updated](https://github.com/actions/runner-images/releases/tag/macOS-12%2F20230117.3) and the default version is now 15, so we install v14 with brew before.

### DIFF
--- a/.github/workflows/build-engine-macos.yml
+++ b/.github/workflows/build-engine-macos.yml
@@ -21,7 +21,7 @@ jobs:
         
     - name: Install dependencies
       run: |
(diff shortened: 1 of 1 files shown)


RiscadoA (approved):

RiscadoA:
Did you add `settings.json` by accident @rsubtil?

rsubtil:
> Did you add `settings.json` by accident @rsubtil?

Ups :sweat_smile:

RiscadoA (approved):

PR #307, opened 2023-01-23
Implement tag inheritance

### LINKED ISSUE #292: Implement tag inheritance in the dispatcher
The dispatcher will support tag inheritance to re-use configurations for similar behaviors:

```cpp
cubos.tag("fixed-step").addCondition(...);
// Tag "fixed-step" runs at a fixed 60 FPS
cubos.tag("physics").inheritsTag("fixed-step").addSystem(...);
// Tag "physics" inherits configs from "fixed-step"
```
Internally, tags are just temporary references to a `SystemSettings` struct, but whether we want inheritance to perform copies or reference existing objects will have an impact on the usage of this feature. With the following setup:
```cpp
cubos.tag("A").addCondition(a);
cubos.tag("B").inheritsTag("A");
cubos.system(b).tagged("B");
```
If the user now does:
```cpp
cubos.tag("A").addSystem(c);
```
Would system `b` implicitly have this new system `c` as well? If yes, this requires having inheritance be a reference to an existing setting. If it was a copy, `b` wouldn't have this new system `c` (although subsequent systems with tag `b` would now have it)

rsubtil:
Closes #292 

Implement proper tag inheritance, which was left to do after the initial dispatcher rework.

Other than that, I found some bugs on the dispatcher along the way that were fixed. I also introduced a warning when users use tags from opposite system types:
```cpp
cubos.tag("A"); // Main tag
cubos.startupSystem(a).tagged("A"); // Startup system, tag mismatch
```

### DIFF
--- a/engine/include/cubos/engine/cubos.hpp
+++ b/engine/include/cubos/engine/cubos.hpp
@@ -30,7 +30,6 @@ namespace cubos::engine
     {
     public:
         TagBuilder(core::ecs::Dispatcher& dispatcher, std::vector<std::string>& tags);
-        ~TagBuilder() = delete;
 
         /// Sets the current tag to be executed before another tag.
         /// @param tag The tag to be executed before.
@@ -50,7 +49,6 @@ namespace cubos::engine
     {
     public:
         SystemBuilder(core::ecs::Dispatcher& dispatcher, std::vector<std::string>& tags);
-        ~SystemBuilder() = delete;
 
         /// Sets the current system's tag.
         /// @param tag The tag to be set.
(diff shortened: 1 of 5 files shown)


engine/include/cubos/engine/cubos.hpp
```diff
@@ -198,7 +201,7 @@ namespace cubos::engine
     SystemBuilder& Cubos::startupSystem(F func)
```
RiscadoA, line 201: This introduces a memory leak. Instead of returning `SystemBuilder&`, return `SystemBuilder`

engine/src/cubos/engine/cubos.cpp
```diff
@@ -69,15 +83,15 @@ Cubos& Cubos::addPlugin(void (*func)(Cubos&))
 TagBuilder& Cubos::tag(const std::string& tag)
 {
     mainDispatcher.addTag(tag);
-    TagBuilder* builder = new TagBuilder(mainDispatcher);
+    TagBuilder* builder = new TagBuilder(mainDispatcher, mainTags);
```
RiscadoA, line 86: Same as above

engine/src/cubos/engine/cubos.cpp
```diff
@@ -69,15 +83,15 @@ Cubos& Cubos::addPlugin(void (*func)(Cubos&))
...
 
     return *builder;
 }
 
 TagBuilder& Cubos::startupTag(const std::string& tag)
 {
     startupDispatcher.addTag(tag);
-    TagBuilder* builder = new TagBuilder(startupDispatcher);
+    TagBuilder* builder = new TagBuilder(startupDispatcher, startupTags);
```
RiscadoA, line 94: Same as above

RiscadoA (changes requested):
Just noticed that we were leaking some memory, but otherwise LGTM

engine/include/cubos/engine/cubos.hpp

engine/src/cubos/engine/cubos.cpp
rsubtil: Fixed

engine/src/cubos/engine/cubos.cpp
rsubtil: Fixed

RiscadoA (approved):

RiscadoA (approved):

PR #315, opened 2023-02-08
Ignore direnv config

luishfonseca:


rsubtil:
Clang is failing because this is a branch from your fork instead of the project. It's OK to ignore it for now until you have permission to develop on the project repo directly.

PR #319, opened 2023-02-22
Add Editor Selected Resource

### LINKED ISSUE #268: Add editor selected resource
A resource is needed to specify the currently selected entity in an ECS world, for debug and editor purposes.
Name TBD.

DiogoMendonc-a:
closes #268


RiscadoA:
I think it would be best to make this less Java/C# like and more like the `ShouldStop` resource.
You can make this resource just a struct with an `Entity` handle. No need to add methods.

Also! `Entity` is already a handle on itself, no need to wrap it with `std::shared_ptr`

engine/include/cubos/engine/data/entity_selector.hpp
RiscadoA, line 1: Since this plugin is so small I think you should define `EntitySelector` in actual plugin header.
We're also moving away from the current structure to a directory structure focused on the actual plugins, so it will make more sense in the future to be other way.

RiscadoA: One other thing: use a `struct` for consistency (that way you don't need the `public:` identifier)

RiscadoA (changes requested):

engine/include/cubos/engine/plugins/tools/entity_selector.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins::tools
+{
+    struct EntitySelector
+    {
+        cubos::core::ecs::Entity selection;
+    };
+    /// Plugin that allows having a singular entity be selected for inspection
```
RiscadoA, line 15: Missing whitespace between the docs and the struct

engine/include/cubos/engine/plugins/tools/entity_selector.hpp
```diff
@@ -0,0 +1,20 @@
...
+    /// @param cubos CUBOS. main class
+    void entitySelectorPlugin(Cubos& cubos);
+}; // namespace cubos::engine::plugins::tools
+
+#endif // CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
```
RiscadoA, line 20: Missing newline at EOF

engine/src/cubos/engine/plugins/tools/entity_selector.cpp
```diff
@@ -0,0 +1,6 @@
+#include <cubos/engine/plugins/tools/entity_selector.hpp>
+
+void cubos::engine::plugins::tools::entitySelectorPlugin(Cubos& cubos)
+{
+    cubos.addResource<cubos::engine::plugins::tools::EntitySelector>();
+}
```
RiscadoA, line 6: Missing newline at EOF

engine/include/cubos/engine/plugins/tools/entity_selector.hpp
```diff
@@ -0,0 +1,20 @@
...
+#define CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
+
+#include <cubos/engine/cubos.hpp>
+#include <cubos/core/ecs/entity_manager.hpp>
+
+using namespace cubos::engine;
+
+namespace cubos::engine::plugins::tools
+{
+    struct EntitySelector
```
RiscadoA, line 11: Missing documentation, maybe something like:
```
/// Resource used to identify the currently selected entity. Used by multiple editor plugins.

File: engine/include/cubos/engine/plugins/tools/entity_selector.hpp [hunk 5]
```diff
@@ -0,0 +1,20 @@
+#ifndef CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
+#define CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
+
+#include <cubos/engine/cubos.hpp>
+#include <cubos/core/ecs/entity_manager.hpp>
+
+using namespace cubos::engine;
```
RiscadoA, line 7: Never use a `using namespace` in a header file: users which include this file will have their namespace polluted otherwise.
Its okay to use them in source files though! (.cpp)

engine/include/cubos/engine/plugins/tools/entity_selector.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
+#define CUBOS_ENGINE_PLUGINS_ENTITY_SELECTOR_HPP
+
+#include <cubos/engine/cubos.hpp>
+#include <cubos/core/ecs/entity_manager.hpp>
```
RiscadoA, line 5: Add a newline between the includes, for consistency.
We usually separate the 'engine' includes from the 'core' includes.

RiscadoA (changes requested):
Almost ready! Just some nitpicks, sorry about that xD

RiscadoA (approved):

PR #321, opened 2023-03-02
Implement conditions on dispatcher

### LINKED ISSUE #290: Add conditions to the Dispatcher
See #241.

rsubtil:
Implement conditions that control system/tag execution.

Closes #290

### DIFF
--- a/core/src/cubos/core/ecs/dispatcher.cpp
+++ b/core/src/cubos/core/ecs/dispatcher.cpp
@@ -10,6 +10,7 @@ void Dispatcher::SystemSettings::copyFrom(const SystemSettings* other)
     std::unique_copy(other->after.tag.begin(), other->after.tag.end(), std::back_inserter(this->after.tag));
     std::unique_copy(other->before.system.begin(), other->before.system.end(), std::back_inserter(this->before.system));
     std::unique_copy(other->after.system.begin(), other->after.system.end(), std::back_inserter(this->after.system));
+    this->conditions |= other->conditions;
 }
 
 Dispatcher::~Dispatcher()
@@ -208,13 +209,52 @@ void Dispatcher::callSystems(World& world, CommandBuffer& cmds)
     if (!this->prepared)
     {
         for (auto& system : systems)
+        {
             system->system->prepare(world);
+        }
+        for (auto& condition : conditions)
+        {
+            condition->prepare(world);
+        }
         this->prepared = true;
     }
 
+    // Clear conditions bitmasks
+    runConditions.reset();
+    retConditions.reset();
+
     for (auto& system : systems)
     {
-        system->system->call(world, cmds);
+        // Query for conditions
+        bool canRun = true;
+        auto conditionsMask = system->settings->conditions;
+        int i = 0;
+        while (conditionsMask.any())
+        {
+            if (conditionsMask.test(0))
+            {
+                // We have a condition, check if it has run already
+                if (!runConditions.test(i))
+                {
+                    runConditions.set(i);
+                    if (conditions[i]->call(world, cmds))
+                    {
+                        retConditions.set(i);
+                    }
+                }
+                // Check if the condition returned true
+                if (!retConditions.test(i))
+                {
+                    canRun = false;
+                    break;
+                }
+            }
+
+            i += 1;
+            conditionsMask >>= 1;
+        }
+        if (canRun)
+            system->system->call(world, cmds);
         // TODO: Check synchronization concerns when this gets multithreaded
         cmds.commit();
     }
--- a/engine/include/cubos/engine/cubos.hpp
+++ b/engine/include/cubos/engine/cubos.hpp
@@ -39,6 +39,13 @@ namespace cubos::engine
         /// @param tag The tag to be executed after.
         TagBuilder& afterTag(const std::string& tag);
 
+        /// Adds a condition to the current tag. If this condition returns
+        /// false, systems with this tag will not be executed. For the tagged
+        /// systems to run, all conditions must return true.
+        /// @param func The condition function.
+        template <typename F>
+        TagBuilder& runIf(F func);
+
     private:
         core::ecs::Dispatcher& dispatcher;
         std::vector<std::string>& tags;
@@ -71,6 +78,13 @@ namespace cubos::engine
         /// @param func The system to be executed after.
         template <typename F>
         SystemBuilder& afterSystem(F func);
+
+        /// Adds a condition to the current system. If this condition returns
+        /// false, the system will not be executed. For a system to run, all
+        /// conditions must return true.
+        /// @param func The condition function.
+        template <typename F>
+        SystemBuilder& runIf(F func);
 
     private:
         core::ecs::Dispatcher& dispatcher;
@@ -151,6 +165,13 @@ namespace cubos::engine
     // Implementation.
 
     template <typename F>
+    TagBuilder& TagBuilder::runIf(F func)
+    {
+        dispatcher.tagAddCondition(func);
+        return *this;
+    }
+
+    template <typename F>
     SystemBuilder& SystemBuilder::beforeSystem(F func)
     {
         dispatcher.systemSetBeforeSystem(func);
@@ -161,6 +182,13 @@ namespace cubos::engine
     SystemBuilder& SystemBuilder::afterSystem(F func)
     {
         dispatcher.systemSetAfterSystem(func);
+        return *this;
+    }
+
+    template <typename F>
+    SystemBuilder& SystemBuilder::runIf(F func)
+    {
+        dispatcher.systemAddCondition(func);
         return *this;
     }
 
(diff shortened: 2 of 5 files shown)


rsubtil:
Ready to review :+1:

core/src/cubos/core/ecs/dispatcher.cpp
```diff
@@ -208,13 +209,27 @@ void Dispatcher::callSystems(World& world, CommandBuffer& cmds)
...
-        system->system->call(world, cmds);
+        bool canRun = true;
+        for (auto& condition : system->settings->conditions)
+        {
+            if (!condition->call(world, cmds))
+            {
+                canRun = false;
+                break;
+            }
+        }
```
RiscadoA, line 230: Instead of calling conditions on all of the systems which depend on it, they should only called once, before the first system is executed. Their result is then stored and used to determine which systems run.

One way I would implement this would be by assigning each condition an ID.
You would also need to determine when the conditions would run.

Then you could use a bitset/mask to store the conditions fulfilled on this invocation of `callSystems`.
When you execute a condition, you either set or clear the bit depending on the result.
When you execute a system, you first compare its 'condition mask' with the mask of fulfilled conditions. If it isn't contained in the other, then the system shouldn't run.

RiscadoA (changes requested):

core/src/cubos/core/ecs/dispatcher.cpp
rsubtil: > before the first system is executed

I'm not sure about this approach; you might have conditions that are dependent of specific systems running and changing variables in the meantime. (for example, system `reloadCurrentAsset` runs if system `assetFileUpdate` ran on this frame). Querying all conditions before the systems run will not permit these kinds of scenarios.

> Then you could use a bitset/mask to store the conditions fulfilled on this invocation of callSystems.

My only concern with this is if we are essentially hardcoding the maximum number of existing conditions to 32/64 bits. This seems like an optimization concern that we are not sure if it is that significant to optimize at all.

RiscadoA: Its not just an optimization concern: it may be confusing for a condition to change mid-frame and some systems running and other not.
You're also assuming that condition systems do not change the state of the application, but they may: we're not preventing the user from writing to resources for example.
Should we?

The example you gave would also work in this system: you could have a resource which stored whether or not the condition should evaluate to true. `assetFileUpdate` would set this flag to true. Before `reloadCurrentAsset` is called, the condition is evaluated, and since the condition executes just before the system is called, `reloadCurrentAsset` will run this frame.

Regarding the 32/64 bits limit, we can either roll out our own dynamically-sized std::bitset, e.g., just a std::vector<uint64_t>, which should be really easy, or use a std::vector<bool> (inneficient since you can't just use bitwise operations)

RiscadoA: We also need to take into account that when you assign a condition to a tag, the user is probably seeing the tag as a group of systems, which either all run or none do.

rsubtil: > You're also assuming that condition systems do not change the state of the application, but they may: we're not preventing the user from writing to resources for example.
Should we?

I'd say we allow this implicitly for now. I can't think of a good example to prove or disprove this, so I think it's best to implement this as is and then see how it behaves in actual usage scenarios.

> The example you gave would also work in this system: you could have a resource which stored whether or not the condition should evaluate to true. assetFileUpdate would set this flag to true. Before reloadCurrentAsset is called, the condition is evaluated, and since the condition executes just before the system is called, reloadCurrentAsset will run this frame.

The issue I see with separating condition checking with system running is that we will lose the multi-threaded functionality. By isolating conditions and querying them all before the systems, we lose the inter-dependence chain that exists, forcing them to be queried one by one in order to ensure deterministic behavior.

However, I don't see how the bitwise check, as well as ensuring tagged systems either all run or none run, can be done in a multi-threaded fashion. So if this multi-threaded behavior ends up not be possible at all, then the proposed changes are better indeed.

RiscadoA: Discussed this in the meeting and it was decided to cache the condition's values, only calculating them right before the first system with that condition is called.

rsubtil: Regarding this, I'm not sure how to do this still as a bitwise check. It shouldn't be a problem to assign bits to specific conditions and build bitmasks for systems, but there's no good way to know whether a given condition has run before doing the bitwise check. It's either 0 for false, or 1 for true; there's no value to indicate the condition needs to be run. Using a `set` to store this information defeats the purpose of the bitwise check, and in that case it would be simpler to just have a `set` that store the return values, and be accessed directly by the condition. What do you think of this issue?

RiscadoA: Why not two bitmasks? One for whether it has already been called, another for the result?

rsubtil: True, that works, but actually I think the problem still exists. I'd imagine the dispatcher doing the following:

- **Checking the condition's bit through a `set` (O(log n)) or `vector` (O(n))**
- Run condition if bit is 0, storing result in cached result bitmask
- Check system condition bitmask with cached result bitmask

So, we still have a data structure that saves the information of what bit a condition is stored in.

RiscadoA: Okay I think I get what you're saying. Maybe the best way is actually 'add an entry' to the call chain which calculates the condition.
The vector containing the order of the systems would need to store a struct with more info, maybe one like
```cpp
struct Entry
{
  int conditionMask; // for conditions, stores the bit to be set if it succeeds, for systems it stores the required conditions
  std::shared_ptr<AnySystemWrapper<bool>> condition;
  std::shared_ptr<AnySystemWrapper<void>> system;
};
```
Basically this way you offload the decision of when to compute the condition to the compile chain method.

One other thing I just thought about:
Considering we probably will never have more than 64 conditions, its okay for it to be 'O(n)'. To be honest, I think we could just set a hard limit for the number of conditions, which could be set through CMake. We're already doing this for components! (I forgot about that: CUBOS_CORE_ECS_MAX_COMPONENTS).

Then you could just use a bitset for the results:
```cpp
std::bitset<CUBOS_CORE_DISPATCHER_MAX_CONDITIONS> result
```

rsubtil: Yeah, I like that solution. I'll implement it then :+1:

As for the 64-bit hard-limit, I think it's fine for now, but even if it becomes an issue, we can make a vector of bitsets so it shouldn't be hard to refactor later on if necessary.

rsubtil: Do we support 32-bits? If yes, we should be aware of this [behavior](https://stackoverflow.com/questions/63797119/does-stdbitset64-work-on-32-bit-machines-for-uint64-t) if problems arise there.

RiscadoA: Hm we shouldn't need to worry about that as long as we don't perform casting between bitsets and integers.
My idea was using them everywhere.

core/src/cubos/core/ecs/dispatcher.cpp
```diff
@@ -10,6 +10,7 @@ void Dispatcher::SystemSettings::copyFrom(const SystemSettings* other)
     std::unique_copy(other->after.tag.begin(), other->after.tag.end(), std::back_inserter(this->after.tag));
     std::unique_copy(other->before.system.begin(), other->before.system.end(), std::back_inserter(this->before.system));
     std::unique_copy(other->after.system.begin(), other->after.system.end(), std::back_inserter(this->after.system));
+    this->conditions = other->conditions;
```
RiscadoA, line 13: Shouldn't this be `|=` ?

engine/include/cubos/engine/cubos.hpp
```diff
@@ -39,6 +39,13 @@ namespace cubos::engine
         /// @param tag The tag to be executed after.
         TagBuilder& afterTag(const std::string& tag);
 
+        /// Adds a condition to the current tag. If this condition returns
+        /// false, systems with this tag will not be executed. For the tagged
+        /// systems to run, all conditions must return true.
+        /// @param func The condition function.
+        template <typename F>
+        TagBuilder& addCondition(F func);
```
RiscadoA, line 47: Rename to `runIf`?

engine/include/cubos/engine/cubos.hpp
```diff
@@ -72,6 +79,13 @@ namespace cubos::engine
         template <typename F>
         SystemBuilder& afterSystem(F func);
 
+        /// Adds a condition to the current system. If this condition returns
+        /// false, the system will not be executed. For a system to run, all
+        /// conditions must return true.
+        /// @param func The condition function.
+        template <typename F>
+        SystemBuilder& addCondition(F func);
```
RiscadoA, line 87: Rename to `runIf`?

RiscadoA (changes requested):

core/src/cubos/core/ecs/dispatcher.cpp
rsubtil: Yeah, you're right :+1:

engine/include/cubos/engine/cubos.hpp
rsubtil: Forgot about this decision in the meeting, thanks :+1:

RiscadoA (approved):

RiscadoA (approved):

PR #329, opened 2023-03-13
Make system parameter syntax more readable

### LINKED ISSUE #320: Make system parameter syntax more readable
Use something else instead of `Position&`, `const Position&` etc.
Maybe `Write<Position>`, `Read<Position>`.

RiscadoA:
Closes #320.

- [X] Add `Read`, `Write`, `OptRead` and `OptWrite` classes.
- [x] Replace `const T&` by `Read<T>`.
- [x] Replace `T&` by `Write<T>`.
- [x] Replace `const T*` by `OptRead<T>`.
- [x] Replace `T*` by `OptWrite<T>`.


core/include/cubos/core/ecs/system.hpp
```diff
@@ -70,6 +70,162 @@ namespace cubos::core::ecs
...
+        inline const T& operator*() const
+        {
+            return this->ref;
+        }
+
+        /// @returns A reference to the resource or component.
+        inline const T& get() const
+        {
+            return this->ref;
+        }
```
luishfonseca: Is it ok to have these two methods of getting the reference? It might lead to inconsistency and make it harder to read the codebase.

core/include/cubos/core/ecs/system.hpp
```diff
@@ -70,6 +70,162 @@ namespace cubos::core::ecs
...
+        {
+            return this->get();
+        }
+
+        /// @returns A reference to the resource or component.
+        inline T& get()
+        {
+            CUBOS_ASSERT(this->ref != nullptr, "Attempted to access a null optional resource or component");
+            return *this->ref;
+        }
```
luishfonseca: If these do the same why is their behaviour different?

luishfonseca: These are outdated because you moved the code in a later commit, the code is still de same in accessors.hpp

core/include/cubos/core/ecs/system.hpp
RiscadoA: True, lets keep only the dereference then.

core/include/cubos/core/ecs/system.hpp
RiscadoA: > If these do the same why is their behaviour different?

Wdym?

luishfonseca: This was made irrelevant by removing the getter, disregard

luishfonseca (approved):

PR #336, opened 2023-03-18
Refactor serialization template code

RiscadoA:
This PR greatly simplifies the serializer/deserializer interface which has been a pain to work with.

Previously, to make a type serializable, you had to declare a new overload of the function `serialize` for that type.
This meant that the declaration had to be put in a header file:

```cpp
// header.hpp
```

I took advantage of a feature of C++ I just found out about:

```
// header.hpp
template <typename T> void func(T);

// usually you would need to define the implementation of the method in the header.
// but! if you want 
```

```

### CODE REVIEW

**Comment by RiscadoA** [comment]
I deactivated all code (in CMake) which depended on the asset manager in some way or another.
This means that this PR shouldn't be merged to master before the assets PR is ready.

**Comment by RiscadoA** [comment]
Will be merged to the new next branch.

File: core/src/cubos/core/gl/palette.cpp [hunk 1]
```diff
@@ -1,5 +1,8 @@
 #include <cubos/core/gl/palette.hpp>
 
+#include <cubos/core/log.hpp>
```
rsubtil, line 3: ditto

core/tests/test_yaml_serialization.cpp
```diff
@@ -3,6 +3,9 @@
 #include <cubos/core/data/yaml_serializer.hpp>
 #include <cubos/core/data/serialization_map.hpp>
 
+#include <glm/glm.hpp>
```
rsubtil, line 6: ditto

rsubtil (approved):
LGTM, with only some small issues to tackle

core/include/cubos/core/data/debug_serializer.hpp
```diff
@@ -67,4 +91,45 @@ namespace cubos::core::data
...
+template <typename T>
+struct fmt::formatter<cubos::core::data::Debug<T>> : formatter<string_view>
+{
+    bool pretty = false; ///< Whether to pretty print the data.
+    bool types = false;  ///< Whether to print the type name.
+
+    inline constexpr auto parse(format_parse_context& ctx) -> decltype(ctx.begin())
+    {
+        auto it = ctx.begin(), end = ctx.end();
+        while (it != end)
```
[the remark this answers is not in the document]
RiscadoA: You need to return the iterator :( That piece of code was basically copied from the fmt docs

core/src/cubos/core/data/binary_serializer.cpp
```diff
@@ -1,6 +1,8 @@
 #include <cubos/core/data/binary_serializer.hpp>
 #include <cubos/core/memory/endianness.hpp>
 
+#include <cassert>
```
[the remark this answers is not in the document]
RiscadoA: Nope, we're using assert in that file. Before it was being included in another header, but with the refactor now we have to include it directly.

core/src/cubos/core/gl/palette.cpp
RiscadoA: Before it was being included in another header, but with the refactor now we have to include it directly.

core/tests/test_yaml_serialization.cpp
RiscadoA: Before it was being included in another header, but with the refactor now we have to include it directly.

PR #350, opened 2023-03-27
Move `cubinhos generate` generated code to source files

RiscadoA:
The only thing in the way of moving generated code to hidden source files was `ComponentStorage<T>` which was used to specify the storage type for each vec. I replaced it by adding a storage param to the existing component `Registry`.

After that I changed `generate` to generate a single file: `${target}-components.cpp`. This source file contains the serialization implementations for all components found and calls to `CUBOS_REGISTER_COMPONENT`. User-facing code is still the same, except that now users can include the header they wrote directly.


engine/include/cubos/engine/ecs/grid.hpp
rsubtil, line 1: Just confirming if this file should have been deleted.

engine/samples/scene/main.cpp
```diff
@@ -23,18 +23,18 @@
...
-#include <components/cubos/position.hpp>
-#include <components/cubos/rotation.hpp>
-#include <components/cubos/scale.hpp>
-#include <components/cubos/local_to_world.hpp>
-#include <components/cubos/grid.hpp>
-#include <components/cubos/camera.hpp>
-
-#include <components/num.hpp>
-#include <components/parent.hpp>
+#include <cubos/engine/ecs/position.hpp
```
rsubtil, line 26: Missing `>`
```suggestion
#include <cubos/engine/ecs/position.hpp>
```

rsubtil (changes requested):
LGTM, there's only a couple of things to check.

engine/include/cubos/engine/ecs/grid.hpp
RiscadoA: Yeah I removed it since it will be rewritten in the renderer refactor in the next PR - necessary since the asset manager changed a lot.

rsubtil (approved):

PR #356, opened 2023-04-21
Add World Inspector

### LINKED ISSUE #267: Add World inspector plugin
The existing inspector code is terrible, and we should rethink the UI.
The world editor is no longer necessary considering we will get #264 and #222.

This plugin will use a special editor resource for identifying the currently selected entity.

DiogoMendonc-a:
closes #267

### DIFF
--- a/engine/include/cubos/engine/tools/world_inspector/plugin.hpp
+++ b/engine/include/cubos/engine/tools/world_inspector/plugin.hpp
@@ -1,0 +1,16 @@
+#pragma once
+
+#include <cubos/engine/cubos.hpp>
+
+namespace cubos::engine::tools
+{
+    /// Plugin that allows inspecting the current world through a ImGui window.
+    ///
+    /// @details This plugin adds one system, which adds a ImGui window with the current world.
+    ///
+    /// Dependencies:
+    /// - `imguiPlugin`
+    //
+    /// @param cubos CUBOS. main class
+    void worldInspectorPlugin(Cubos& cubos);
+}; // namespace cubos::engine::tools
--- a/engine/src/cubos/engine/tools/world_inspector/plugin.cpp
+++ b/engine/src/cubos/engine/tools/world_inspector/plugin.cpp
@@ -1,0 +1,41 @@
+#include <cubos/engine/tools/world_inspector/plugin.hpp>
+#include <imgui.h>
+
+#include <cubos/engine/plugins/imgui.hpp>
+#include <cubos/engine/plugins/tools/entity_selector.hpp>
+
+using cubos::core::ecs::World;
+using cubos::engine::plugins::tools::EntitySelector;
+
+static void inspectWorld(World& world, EntitySelector& selector)
+{
+    ImGui::Begin("World Inspector");
+    if (!ImGui::IsWindowCollapsed())
+    {
+        int n = 0;
+        for (auto entity : world)
+        {
+            ImGui::PushID(n);
+
+            ImGui::BulletText(std::to_string(n).c_str());
+
+            ImGui::SameLine();
+            if (ImGui::Button("Select"))
+            {
+                selector.selection = entity;
+            }
+
+            n++;
+            ImGui::PopID();
+        }
+    }
+    ImGui::End();
+}
+
+void cubos::engine::tools::worldInspectorPlugin(Cubos& cubos)
+{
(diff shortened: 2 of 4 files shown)


engine/include/cubos/engine/plugins/tools/world_inspector.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef CUBOS_ENGINE_PLUGINS_WORLD_INSPECTOR_HPP
+#define CUBOS_ENGINE_PLUGINS_WORLD_INSPECTOR_HPP
+
+#include <cubos/engine/cubos.hpp>
+
+using namespace cubos::engine;
```
RiscadoA, line 6: Never use `using namespace` in header files - and in source files prefer `using`ing specifics types instead of the whole namespace. For example:

```cpp
// in a .cpp file
// instead of
// prefer
using cubos::engine::Cubos; // imports only that type into this namespace


```

engine/src/cubos/engine/plugins/tools/world_inspector.cpp
```diff
@@ -0,0 +1,42 @@
...
+using namespace cubos::engine::plugins::tools;
+using namespace cubos::core::ecs;
+
+static void inspectWorld(World& world, EntitySelector& selector)
+{
+    ImGui::Begin("World Inspector");
+    if (!ImGui::IsWindowCollapsed())
+    {
+        int n = 0;
+        for (World::Iterator i = world.begin(); i != world.end(); ++i)
```
RiscadoA, line 17: I think you can the use the for-each syntax: `for (auto entity : world)`, not sure.

RiscadoA (changes requested):
Overall LGTM, just some minor nitpicks and refactoring related issues:
- we're moving things out of the `plugins` directory, and new plugins should follow those guidelines - the plugin's file should be `cubos/engine/tools/world_inspector/plugin.hpp`
- the namespaces are also being simplified, so you can just use `cubos::engine` instead of `cubos::engine::tools::world_inspector`

engine/src/cubos/engine/plugins/tools/world_inspector.cpp
DiogoMendonc-a: Oh yeah, I always forget there's foreach in c++

engine/include/cubos/engine/plugins/tools/world_inspector.hpp
RiscadoA: Don't forget to change the `using namespace` for `using X`

engine/include/cubos/engine/tools/world_inspector/plugin.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef CUBOS_ENGINE_PLUGINS_WORLD_INSPECTOR_HPP
+#define CUBOS_ENGINE_PLUGINS_WORLD_INSPECTOR_HPP
```
RiscadoA, line 2: Use `#pragma once`

engine/include/cubos/engine/tools/world_inspector/plugin.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include <cubos/engine/cubos.hpp>
+
+namespace cubos::engine
```
RiscadoA, line 6: Change it to `cubos::engine::tools` for consistency with #361

engine/src/cubos/engine/tools/world_inspector/plugin.cpp
```diff
@@ -0,0 +1,42 @@
+#include <cubos/engine/tools/world_inspector/plugin.hpp>
+#include <imgui.h>
+
+#include <cubos/engine/plugins/imgui.hpp>
+#include <cubos/engine/plugins/tools/entity_selector.hpp>
+
+using namespace cubos;
+using namespace cubos::engine::plugins::tools;
+using namespace cubos::core::ecs;
```
RiscadoA, line 9: Change this to something like:

```
using cubos::core::ecs::World;
using namespace cubos::engine; // <- okay because this belongs to the engine module anyway

RiscadoA (changes requested):

RiscadoA (changes requested):
Use imperative on commit messages and squash the fix commits

RiscadoA (approved):

PR #330, opened 2023-03-13
Fix compiler warnings

RiscadoA:
- [X] Treat warnings as errors on `core`
- [X] Fix all warnings on `core`
- [x] Treat warnings as errors on `engine`
- [x] Fix all warnings on `engine`
- [x] Treat warnings as errors on the tools
- [x] Fix all warnings on the tools
- [x] Configure clang-tidy for static analysis

PR #357, opened 2023-04-25
Update renderer plugin and samples to use the asset plugin

RiscadoA:
Depends on #330 

Changes:
- add a CMake option to disable `clang-tidy` (check #364).
- fix multiple errors in `cubinhos generate` (both CMake config and generate code).
- add `[[cubos::ignore]]` to `cubinhos generate` - prevents `generate` from generating erialization code for a particular field.
- reactivated the scene sample and removed all of the bloat.
- fix errors found in `Blueprint` which prevented scenes from being spawned successfully.
- make entity deserialization fail in a proper way - previously it was just crashing instead of setting the error flag.
- add an OpenGL message callback to `OGLRenderDevice` - very useful to debug GL errors (I was desperate).
- make `Asset<T>` serializable - previously only `AnyAsset` supported it.
- disable `Assets` file IO when passed an empty path for the assets folder.
- use saner namespaces in the renderer code - no more `gl`, `gl::pps` and `gl::deferred::Renderer` on the engine.
- add light components - found out we still didn't have them after losing one hour trying to understand why I was getting a black screen.
- delete `bloom` and `ssao` samples - too low level to be here (not using `Cubos` yet), those features should be showcased in future improvements to the `renderer` sample.
- delete `custom-asset` sample - forgot to remove this in the `assetsPlugin` PR, I had already added an equivalent sample.


core/src/cubos/core/gl/ogl_render_device.cpp
```diff
@@ -14,6 +14,24 @@ using namespace cubos::core::gl;
...
+    (void)userParam;
+    (void)id;
+
+    if (severity == GL_DEBUG_SEVERITY_HIGH)
+    {
+        CUBOS_ERROR("OpenGL message (source = 0x{:x}, type = 0x{:x}): {}", source, type, message);
+    }
+    else if (severity == GL_DEBUG_SEVERITY_MEDIUM)
+    {
+        CUBOS_WARN("OpenGL message (source = 0x{:x}, type = 0x{:x}): {}", source, type, message);
```
luishfonseca, line 31: Our errors starting with "OpenGL error:" but errors and warnings emitted by Opengl starting with "OpenGL message" is confusing

engine/include/cubos/engine/assets/plugin.hpp
```diff
@@ -6,6 +6,7 @@
 namespace cubos::engine
 {
     /// Plugin which adds asset management to CUBOS.
+    /// If the `assets.path` setting is set to an empty string, no file IO will be done.
```
luishfonseca, line 9: I don't like this, see other review comment

engine/include/cubos/engine/renderer/plugin.hpp
```diff
@@ -51,6 +51,9 @@ namespace cubos::engine
     /// - `RenderableGrid`: a grid to be rendered. Must be used with `LocalToWorld`.
     /// - `Camera`: holds camera information. The entity pointed to by `ActiveCamera` must have
     ///   this component.
+    /// - `SpotLight`
+    /// - `DirectionalLight`
+    /// - `PointLight`
```
luishfonseca, line 56: Missing explanation

luishfonseca (changes requested):

luishfonseca (approved):

Issue #349, opened 2023-03-27
Update renderer plugin and samples to use the asset plugin


RiscadoA:
Currently the renderer plugin is deactivated on `next` since it depended on the previous asset manager.
Now that the PR for the new assets plugin is ready, it should be adapted to use it (should be ez).
The samples are also still using the old asset manager and initializing it manually, which should be replaced by the new functionality.

RiscadoA:
Follow up from #348

Issue #338, opened 2023-03-18
Review namespaces on the engine


RiscadoA:
All plugins should just use the namespace `cubos::engine`

https://abseil.io/tips/130

RiscadoA:
Should we keep nested namespaces for plugin categories? E.g.: `cubos::engine::tools`

PR #361, opened 2023-05-06
Restructure plugin directories

luishfonseca:
The ecs and gl folders look out of place now, as they are the only that are not plugins. We should think about that before merging this


luishfonseca:
I've moved most of the ecs folder to transform, the remaining code in ecs and gl will be moved to the renderer in #349

rsubtil (approved):

engine/include/cubos/engine/tools/entity_selector/plugin.hpp
```diff
@@ -4,7 +4,7 @@
 
 #include <cubos/engine/cubos.hpp>
 
-namespace cubos::engine::plugins::tools
```
RiscadoA, line 7: Should we keep the `tools` namespace? What do you think=

engine/samples/scene/main.cpp
```diff
@@ -13,18 +13,18 @@
...
-#include <cubos/engine/plugins/transform.hpp>
-#include <cubos/engine/plugins/window.hpp>
+#include <cubos/engine/imgui/plugin.hpp>
+#include <cubos/engine/renderer/plugin.hpp>
+#include <cubos/engine/tools/settings_inspector/plugin.hpp>
+#include <cubos/engine/transform/local_to_world.hpp>
+#include <cubos/engine/transform/plugin.hpp>
+#include <cubos/engine/transform/position.hpp>
+#include <cubos/engine/transform/rotation.hpp>
+#include <cubos/engine/transform/scale.hpp>
```
RiscadoA, line 26: Having to write this basically every time you use the plugin is a nuissance, I think we should either include those headers inside `transform/plugin.hpp` or define the components in that file instead of on separate headers.

engine/src/cubos/engine/transform/plugin.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <cubos/engine/transform/local_to_world.hpp>
+#include <cubos/engine/transform/plugin.hpp>
+#include <cubos/engine/transform/position.hpp>
+#include <cubos/engine/transform/rotation.hpp>
+#include <cubos/engine/transform/scale.hpp>
+
+using namespace cubos;
+
+static void applyTransform(
+    core::ecs::Query<engine::LocalToWorld&, const engine::Position*, const engine::Rotation*, const engine::Scale*>
```
RiscadoA, line 12: Unrelated to the PR but:
```suggestion
using cubos::core::ecs::Query;
using namespace cubos::engine;

static void applyTransform(
    Query<LocalToWorld&, const Position*, const Rotation*, const Scale*>
```

engine/src/cubos/engine/window/plugin.cpp
```diff
@@ -1,8 +1,9 @@
 #include <cubos/core/settings.hpp>
 
-#include <cubos/engine/plugins/window.hpp>
+#include <cubos/engine/window/plugin.hpp>
 
 using namespace cubos::core;
```
RiscadoA, line 5: Replace with `using cubos::core::io::Window` etc. The `using namespace cubos::engine` is okay since this file belongs to the engine anyway.

tools/tesserato/src/main.cpp
```diff
@@ -4,8 +4,8 @@
 #include <cubos/core/ui/serialization.hpp>
 
 #include <cubos/engine/cubos.hpp>
-#include <cubos/engine/plugins/env_settings.hpp>
-#include <cubos/engine/plugins/window.hpp>
+#include <cubos/engine/env_settings/plugin.hpp>
+#include <cubos/engine/window/plugin.hpp>
```
RiscadoA, line 9: Use `using namespace cubos::engine;`

RiscadoA (changes requested):

engine/include/cubos/engine/tools/entity_selector/plugin.hpp
luishfonseca: If we remove it that should be done in #338 imo

engine/samples/scene/main.cpp
luishfonseca: I'll move the headers to transform/plugin.hpp

RiscadoA (approved):

PR #373, opened 2023-05-11
Add InputBindings asset

luishfonseca:
Depends on #344

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true


engine/include/cubos/engine/assets/serializer_bridge.hpp
```diff
@@ -0,0 +1,77 @@
...
+            S ser{*stream};
+            core::data::Serializer* serializer = &ser;
+
+            // Read the data from the asset.
+            auto data = assets.read<T>(handle);
+            serializer->write(data, nullptr);
+
+            if (serializer->failed())
+            {
+                CUBOS_ERROR("Could not binary serialize asset to file '{}'", path);
```
RiscadoA, line 70: Remove `binary` from the message

engine/include/cubos/engine/assets/serializer_bridge.hpp
```diff
@@ -0,0 +1,77 @@
...
+            auto path = assets.readMeta(handle)->get("path").value();
+            auto stream = core::data::FileSystem::open(path, core::data::File::OpenMode::Write);
+            if (stream == nullptr)
+            {
+                CUBOS_ERROR("Could not open file '{}'", path);
+                return false;
+            }
+
+            S ser{*stream};
+            core::data::Serializer* serializer = &ser;
```
RiscadoA, line 62: You can just use `ser.write` directly

engine/include/cubos/engine/assets/serializer_bridge.hpp
```diff
@@ -0,0 +1,77 @@
...
+            {
+                CUBOS_ERROR("Could not open file '{}'", path);
+                return false;
+            }
+
+            T data{};
+
+            D des{*stream};
+            core::data::Deserializer* deserializer = &des;
+            deserializer->read(data);
```
RiscadoA, line 37: You can just use `des.read` directly

RiscadoA (changes requested):

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,34 @@
+/// @file
+/// @brief
+
+#pragma once
+
+#include <cubos/core/data/serializer.hpp>
+
+using cubos::core::data::Deserializer;
+using cubos::core::data::Serializer;
```
RiscadoA, line 9: Never do this in headers

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,34 @@
...
+    ser.endObject();
+}
+
+template <>
+void cubos::core::data::deserialize<cubos::engine::Bindings>(Deserializer& des, cubos::engine::Bindings& obj)
+{
+    des.beginObject();
+    des.read(obj.an_integer);
+    des.endObject();
+}
```
RiscadoA, line 34: This should be in the source file.

RiscadoA (changes requested):

engine/include/cubos/engine/input/bindings.hpp
luishfonseca: Posso manter o `using cubos::core::io::Key;`?

core/include/cubos/core/io/keyboard.hpp
```diff
@@ -142,4 +144,57 @@ namespace cubos::core::io
...
+            return Key::Up;
+        else if (str == "Down")
+            return Key::Down;
+        else if (str == "Left")
+            return Key::Left;
+        else if (str == "Right")
+            return Key::Right;
+        else
+            return Key::Invalid;
+    }
```
RiscadoA, line 198: I would just declare the functions here and define them on a `keyboard.cpp`, as not to clutter the header with massive implementations.

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,115 @@
+/// @file
+/// @brief
```
RiscadoA, line 2: Don't forget to describe what is in the file here. Also, we need to have a convention for these commits, should we write:
```cpp
/// @file
/// Bla bla bla
```
or
```cpp
/// @file
/// @brief Bla bla bla

File: engine/include/cubos/engine/input/bindings.hpp [hunk 8]
```diff
@@ -0,0 +1,115 @@
...
+#include <vector>
+
+#include <cubos/core/io/keyboard.hpp>
+
+namespace cubos::engine
+{
+    class KeyWithModifier final
+    {
+    public:
+        KeyWithModifier() = default;
```
RiscadoA, line 18: In this case it's not very clear what would the default value be - it is probably 0, which is `Key::A`, when it should be `Key::Invalid` - I would prefer you either define the constructor and the key to `Key::Invalid`, or add a field initializer down in the private section, like
```cpp
core::io::Key mKey = core::io::Key::Invalid;
```

engine/include/cubos/engine/input/bindings.hpp
RiscadoA, line 1: You can drop the `cubos::` in most of the code, and just write `core::io::`

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,115 @@
...
+        ~Axis() = default;
+
+        const std::vector<KeyWithModifier>& getPositive() const;
+        const std::vector<KeyWithModifier>& getNegative() const;
+
+        std::vector<KeyWithModifier>& getPositive();
+        std::vector<KeyWithModifier>& getNegative();
+
+        float getValue() const;
+        void setValue(float value);
```
RiscadoA, line 59: Use `value()` and `value(float value)` for consistency (the same goes for other methods).

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        std::string toString() const;
+
+    private:
+        std::vector<KeyWithModifier> mPositive;
+        std::vector<KeyWithModifier> mNegative;
+
+        // Not serialized
```
RiscadoA, line 67: Triple slash comment + final period.

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,44 @@
...
+        void clear();
+
+        /// Clears all bindings for a specific player
+        /// @param player The player whose bindings will be cleared
+        void clear(int player);
+
+        /// Sets the bindings for a specific player
+        /// @param player The player whose bindings will be set
+        /// @param bindings The bindings to set
+        void setBindings(int player, const Bindings& bindings);
```
RiscadoA, line 23: Maybe `bind`?

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+        /// Gets an action state for a specific player
+        /// @param player The player whose action state will be retrieved
+        /// @param actionName The name of the action
+        /// @return True if the action exists and is pressed, false otherwise.
+        bool isPressed(int player, const std::string& actionName) const;
```
RiscadoA, line 29: `pressed`?

engine/samples/input/assets/sample.bind
```diff
@@ -0,0 +1,52 @@
+{
+    "action": {
```
RiscadoA, line 2: I spent some time confused does to this not being plural, maybe change it to `actions`

engine/samples/input/assets/sample.bind
```diff
@@ -0,0 +1,52 @@
...
+            "M-a"
+        ],
+        "ctrl-a": [
+            "C-a"
+        ],
+        "ctrl-shift-a": [
+            "C-S-a"
+        ]
+    },
+    "axis": {
```
RiscadoA, line 20: And here `axes`

engine/include/cubos/engine/input/plugin.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <cubos/engine/cubos.hpp>
+
```
RiscadoA, line 4: I would include at least the `Input` class file here - when the user includes this header its basically guaranteed they also want to access the resource.

engine/src/cubos/engine/input/bindings.cpp
```diff
@@ -0,0 +1,314 @@
...
+        ss << "+Alt";
+    }
+
+    if ((mModifiers & Modifiers::System) != Modifiers::None)
+    {
+        ss << "+System";
+    }
+
+    return ss.str();
+}
```
RiscadoA, line 63: This method is unnececessary since the type is serializable anyway! I would make both `io::Key` and `io::Modifiers` serializable, and then, when you need debug printing, just use the `Debug` utility to print serializable types.

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+    // TODO: Deregister from all the events for this player
+    mBindings.erase(player);
+    CUBOS_DEBUG("Input cleared for player {}", player);
+}
+
+void Input::setBindings(int player, const Bindings& bindings)
+{
+    // TODO: Register to all the events for this player
+    mBindings[player] = std::move(bindings);
```
RiscadoA, line 24: The move doesn't do anything here you can't move a type behind a `const` reference - you can remove it.

RiscadoA (changes requested):

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,119 @@
...
+        result += "C-";
+    if ((modifiers & Modifiers::Shift) != Modifiers::None)
+        result += "S-";
+    if ((modifiers & Modifiers::Alt) != Modifiers::None)
+        result += "M-";
+    if ((modifiers & Modifiers::System) != Modifiers::None)
+        result += "D-";
+
+    return result;
+}
```
RiscadoA, line 25: So modifiers always end in `-`? Wouldn't it make more sense to add the trailing '-' in the `KeyAndModifier`?

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+std::string io::keyToString(Key key)
+{
+    char key_char = static_cast<char>(key);
+
+    if (key_char >= 0 && key_char <= 25)
+    {
+        return std::string(1, key_char + 'a');
+    }
```
RiscadoA, line 64: I think its best not to do this, just write a `switch` for every key - way less error prone and if the enum changes for some reason it won't break .

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,115 @@
...
+        void pressed(bool pressed);
+
+        std::string toString() const;
+
+    private:
+        std::vector<KeyWithModifiers> mKeys;
+
+        /// Not serialized.
+        bool mPressed;
+    };
```
RiscadoA, line 95: I would move this to another file `action.hpp`

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+        std::string toString() const;
+
+    private:
+        std::vector<KeyWithModifiers> mPositive;
+        std::vector<KeyWithModifiers> mNegative;
+
+        /// Not serialized.
+        float mValue;
+    };
```
RiscadoA, line 69: I would move this to another file `axis.hpp`

engine/src/cubos/engine/input/bindings.cpp
```diff
@@ -0,0 +1,202 @@
...
+    ser.endDictionary();
+    ser.endObject();
+}
+
+template <>
+void cubos::core::data::deserialize<InputBindings>(Deserializer& des, InputBindings& obj)
+{
+    des.beginObject();
+
+    size_t actions_sz = des.beginDictionary();
```
RiscadoA, line 140: Always prefix `size_t` with `std::` - the standard does not require `size_t` to be defined in the global scope, we had problems with this in the past.

engine/src/cubos/engine/input/bindings.cpp
```diff
@@ -0,0 +1,202 @@
...
+        CUBOS_ASSERT(tmp == "pos", "Expected 'pos' in axis dictionary");
+        std::vector<KeyWithModifiers> pos;
+        des.read(pos);
+
+        des.read(tmp);
+        CUBOS_ASSERT(tmp == "neg", "Expected 'neg' in axis dictionary");
+        std::vector<KeyWithModifiers> neg;
+        des.read(neg);
+
+        des.endDictionary();
```
RiscadoA, line 172: I assume you forgot to change this, you should also use `beginObject` here.

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,117 @@
...
+        for (const auto& pos : axis.second.positive())
+        {
+            mBoundAxes[pos.key()].push_back(BindingIndex{axis.first, player, false});
+        }
+
+        for (const auto& neg : axis.second.negative())
+        {
+            mBoundAxes[neg.key()].push_back(BindingIndex{axis.first, player, true});
+        }
+    }
```
RiscadoA, line 70: Shouldn't your first clear the bindings for that player?

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,117 @@
...
+    mBindings[player] = bindings;
+    CUBOS_DEBUG("Input bindings set for player {}", player);
+}
+
+bool Input::pressed(const std::string& actionName, int player) const
+{
+    auto pIt = mBindings.find(player);
+    if (pIt == mBindings.end())
+    {
+        CUBOS_WARN("Player {} does not exist", player);
```
RiscadoA, line 81:
```suggestion
        CUBOS_WARN("Player {} does not have any associated input bindings", player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return false;
+    }
+
+    auto aIt = pIt->second.actions().find(actionName);
+    if (aIt == pIt->second.actions().end())
+    {
+        CUBOS_WARN("Action {} does not exist", actionName);
```
RiscadoA, line 88:
```suggestion
        CUBOS_WARN("Action {} is not bound to any input for player {}", actionName, player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    return aIt->second.pressed();
+}
+
+float Input::axis(const std::string& axisName, int player) const
+{
+    auto pIt = mBindings.find(player);
+    if (pIt == mBindings.end())
+    {
+        CUBOS_WARN("Player {} does not exist", player);
```
RiscadoA, line 100: Same as above

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return 0.0f;
+    }
+
+    auto aIt = pIt->second.axes().find(axisName);
+    if (aIt == pIt->second.axes().end())
+    {
+        CUBOS_WARN("Axis {} does not exist", axisName);
```
RiscadoA, line 107: Same as above

RiscadoA (changes requested):

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+    if ((modifiers & Modifiers::Alt) != Modifiers::None)
+        result += "M-";
+    if ((modifiers & Modifiers::System) != Modifiers::None)
+        result += "D-";
+
+    return result;
+}
+
+template <>
+void cubos::core::data::serialize<Modifiers>(Serializer& ser, const Modifiers& obj, const char* name)
```
RiscadoA, line 28: Replace `cubos::core::data` with `data`

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+    std::string str;
+    des.read(str);
+    obj = io::stringToModifiers(str);
+}
+
+std::string io::keyToString(Key key)
+{
+    // TODO: Add more keys.
+    switch (key)
```
RiscadoA, line 60: Add missing keys

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+template <>
+void cubos::core::data::serialize<Key>(Serializer& ser, const Key& obj, const char* name)
+{
+    ser.write(io::keyToString(obj), name);
+}
+
+Key io::stringToKey(const std::string& str)
+{
+    // TODO: Add more keys.
+    if (str == "a")
```
RiscadoA, line 136: Add missing keys

RiscadoA: Also, maybe use a macro for this? Look at the GLFW window code for converting from/to GLFW key codes.

core/src/cubos/core/io/keyboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+        return Key::Invalid;
+}
+
+template <>
+void cubos::core::data::deserialize<Key>(Deserializer& des, Key& obj)
+{
+    std::string str;
+    des.read(str);
+    obj = io::stringToKey(str);
+}
```
RiscadoA, line 206: Missing newline at EOF

engine/include/cubos/engine/input/action.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+        ~InputAction() = default;
+
+        const std::vector<KeyWithModifiers>& keys() const;
+        std::vector<KeyWithModifiers>& keys();
+
+        bool pressed() const;
+        void pressed(bool pressed);
+
+        std::string toString() const;
```
RiscadoA, line 26: Unnecessary, it can be debug printed anyway

engine/include/cubos/engine/input/axis.hpp
```diff
@@ -0,0 +1,39 @@
...
+        const std::vector<KeyWithModifiers>& positive() const;
+        const std::vector<KeyWithModifiers>& negative() const;
+
+        std::vector<KeyWithModifiers>& positive();
+        std::vector<KeyWithModifiers>& negative();
+
+        float value() const;
+        void value(float value);
+
+        std::string toString() const;
```
RiscadoA, line 30: Unnecessary, it can be debug printed anyway

engine/include/cubos/engine/input/bindings.hpp
```diff
@@ -0,0 +1,32 @@
...
+        InputBindings() = default;
+        ~InputBindings() = default;
+
+        const std::unordered_map<std::string, InputAction>& actions() const;
+        const std::unordered_map<std::string, InputAxis>& axes() const;
+
+        std::unordered_map<std::string, InputAction>& actions();
+        std::unordered_map<std::string, InputAxis>& axes();
+
+        std::string toString() const;
```
RiscadoA, line 26: Unnecessary, it can be debug printed anyway

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,77 @@
...
+        /// Handle a modifiers event
+        /// @param event The modifiers event
+        void handle(const core::io::ModifiersEvent& event);
+
+        /// Discard remaining events.
+        /// @param event The event to discard
+        inline void handle(const core::io::WindowEvent& event)
+        {
+            (void)event;
+        }
```
RiscadoA, line 53: I would remove this and move the logic to the caller

engine/include/cubos/engine/input/plugin.hpp
```diff
@@ -0,0 +1,23 @@
...
+    ///
+    /// Resources:
+    /// - `Input`: stateful input manager, used to query the input state.
+    ///
+    /// Dependencies:
+    /// - `assetsPlugin`
+    ///
+    /// @param cubos CUBOS. main class.
+    void inputPlugin(Cubos& cubos);
+} // namespace cubos::engine
```
RiscadoA, line 23: Missing line at EOF

engine/include/cubos/engine/input/plugin.hpp
```diff
@@ -0,0 +1,23 @@
...
+    /// @details
+    ///
+    /// Events:
+    /// - `InputEvent`: emitted when an input event occurs.
+    ///
+    /// Resources:
+    /// - `Input`: stateful input manager, used to query the input state.
+    ///
+    /// Dependencies:
+    /// - `assetsPlugin`
```
RiscadoA, line 19: Add `windowPlugin`

engine/src/cubos/engine/input/action.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+bool InputAction::pressed() const
+{
+    return mPressed;
+}
+
+void InputAction::pressed(bool pressed)
+{
+    mPressed = pressed;
+}
```
RiscadoA, line 23: Missing line

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+    mBindings.clear();
+    CUBOS_DEBUG("Input bindings cleared");
+}
+
+void Input::removeBoundPlayer(std::vector<BindingIndex>& boundKeys, int player)
+{
+    boundKeys.erase(
+        std::remove_if(boundKeys.begin(), boundKeys.end(), [player](const auto& idx) { return idx.player == player; }),
+        boundKeys.end());
+}
```
RiscadoA, line 24: What in tarnation is this - at least separate it into two lines and add a comment explaining what this does

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    mBindings.erase(player);
+    CUBOS_DEBUG("Input bindings cleared for player {}", player);
+}
+
+void Input::bind(const InputBindings& bindings, int player)
+{
+    clear(player);
```
RiscadoA, line 55:
```suggestion
    this->clear(player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+        boundKeys.end());
+}
+
+void Input::clear(int player)
+{
+    for (const auto& action : mBindings[player].actions())
+    {
+        for (const auto& key : action.second.keys())
+        {
+            removeBoundPlayer(mBoundActions[key.key()], player);
```
RiscadoA, line 32:
```suggestion
            this->removeBoundPlayer(mBoundActions[key.key()], player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    for (const auto& axis : mBindings[player].axes())
+    {
+        for (const auto& pos : axis.second.positive())
+        {
+            removeBoundPlayer(mBoundAxes[pos.key()], player);
```
RiscadoA, line 40:
```suggestion
            this->removeBoundPlayer(mBoundAxes[pos.key()], player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        for (const auto& neg : axis.second.negative())
+        {
+            removeBoundPlayer(mBoundAxes[neg.key()], player);
```
RiscadoA, line 45:
```suggestion
            this->removeBoundPlayer(mBoundAxes[neg.key()], player);
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+}
+
+bool Input::anyPressed(std::vector<KeyWithModifiers>& keys) const
+{
+    for (auto& key : keys)
+    {
+        if (key.pressed())
+            return true;
```
RiscadoA, line 141:
```suggestion
        if (key.pressed())
        {
            return true;
        }
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+void Input::handle(const ModifiersEvent& event)
+{
+    mModifiers = event.modifiers;
+}
+
+const std::unordered_map<int, InputBindings>& Input::bindings() const
+{
+    return mBindings;
+}
```
RiscadoA, line 197: Missing newline

engine/src/cubos/engine/input/key_with_modifiers.cpp
```diff
@@ -0,0 +1,71 @@
...
+template <>
+void cubos::core::data::deserialize<KeyWithModifiers>(Deserializer& des, KeyWithModifiers& obj)
+{
+    std::string str;
+    des.readString(str);
+
+    std::size_t split = str.find_last_of('-');
+    if (split == std::string::npos)
+    {
+        obj = KeyWithModifiers(stringToKey(str), Modifiers::None);
```
RiscadoA, line 65: Prefer brace initialization over parentheses
```suggestion
        obj = KeyWithModifiers{stringToKey(str), Modifiers::None};
```

engine/src/cubos/engine/input/plugin.cpp
```diff
@@ -0,0 +1,40 @@
...
+static void bridge(Write<Assets> assets)
+{
+    assets->registerBridge(".bind", std::make_unique<JSONBridge<InputBindings>>());
+}
+
+static void update(Write<Input> input, EventReader<WindowEvent> events)
+{
+    for (auto event : events)
+    {
+        std::visit([&input](auto e) { input->handle(e); }, event);
```
RiscadoA, line 25: 'Catch' the specific event types with the visitor, and add an `auto` catch all for other event types which doesn't call the input object.

RiscadoA (changes requested):

engine/src/cubos/engine/input/plugin.cpp
luishfonseca: that requires an overload template, let's not please

RiscadoA: Sure...

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,80 @@
...
+        /// Sets the bindings for a specific player
+        /// @param bindings The bindings to set
+        /// @param player The player whose bindings will be set
+        void bind(const InputBindings& bindings, int player = 0);
+
+        /// Gets an action state for a specific player
+        /// @param actionName The name of the action
+        /// @param player The player whose action state will be retrieved
+        /// @return True if the action exists and is pressed, false otherwise.
+        bool pressed(const std::string& actionName, int player = 0) const;
```
RiscadoA, line 35: Prefer using `std::string_view` over `const std::string&`

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+        auto& action = mBindings[boundAction.player].actions()[boundAction.name];
+
+        handleKeys(action.keys(), event);
+
+        bool pressed = anyPressed(action.keys());
+        if (action.pressed() != pressed)
+        {
+            action.pressed(pressed);
+            CUBOS_INFO("Action {} was {}", boundAction.name, pressed ? "pressed" : "released");
```
RiscadoA, line 164: This is too spammy, maybe make it `_TRACE`

engine/src/cubos/engine/input/input.cpp
```diff
@@ -0,0 +1,206 @@
...
+        }
+        if (anyPressed(axis.negative()))
+        {
+            value -= 1.0f;
+        }
+
+        if (axis.value() != value)
+        {
+            axis.value(value);
+            CUBOS_INFO("Axis {} value is {}", boundAxis.name, value);
```
RiscadoA, line 193: Also too spammy, use `_TRACE`

engine/src/cubos/engine/input/key_with_modifiers.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return mPressed;
+}
+
+void KeyWithModifiers::pressed(bool pressed)
+{
+    if (mPressed == pressed)
+        return;
+
+    mPressed = pressed;
+    CUBOS_DEBUG("Key {} was {}", modifiersToString(mModifiers) + keyToString(mKey), pressed ? "pressed" : "released");
```
RiscadoA, line 48: I know this is useful to debug the Input plugin, but should we keep it from now on? If we do, it should be `_TRACE`

engine/include/cubos/engine/input/plugin.hpp
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+#include <cubos/engine/cubos.hpp>
+#include <cubos/engine/input/input.hpp>
+
+namespace cubos::engine
+{
+    /// Plugin which adds input handling to CUBOS.
+    ///
+    /// @details
```
RiscadoA, line 12: Check out the `Scene` plugin docs. This documentation should make it clear that a bridge is being added for loading JSON `InputBindings` with the extension `.bind`

RiscadoA (changes requested):

engine/include/cubos/engine/input/input.hpp
luishfonseca: had to use `const char*` because of the unordered_map find()

engine/include/cubos/engine/input/action.hpp
```diff
@@ -0,0 +1,43 @@
+/// @file
+/// @brief Contains the InputAction class.
+#pragma once
```
RiscadoA, line 3: Add newline between docs and code

engine/include/cubos/engine/input/plugin.hpp
```diff
@@ -0,0 +1,29 @@
...
+    /// Events:
+    /// - `InputEvent`: emitted when an input event occurs. (TODO)
+    ///
+    /// Resources:
+    /// - `Input`: stateful input manager, used to query the input state.
+    ///
+    /// Dependencies:
+    /// - `assetsPlugin`
+    /// - `windowPlugin`
+    ///
```
RiscadoA, line 26: Missing tag documentation. I suggest `cubos.input` since there aren't any other relevant tags anyway.

RiscadoA (changes requested):

RiscadoA (approved):

RiscadoA:
LGTM, just missing a better sample now

engine/samples/input/main.cpp
```diff
@@ -0,0 +1,198 @@
...
+using cubos::core::Settings;
+using cubos::core::data::Debug;
+using cubos::core::ecs::Read;
+using cubos::core::ecs::Write;
+using namespace cubos::engine;
+
+static const Asset<InputBindings> bindingsAsset = AnyAsset("bf49ba61-5103-41bc-92e0-8a442d7842c3");
+int showcase = 0;
+bool nextPressed = false;
+bool explained = false;
```
RiscadoA, line 17: I know this is just a sample but since it is a sample we should show good practices: this should be a resource.
Add something like
```suggestion
struct State
{
    int showcase = 0;
    bool nextPressed = false;
    bool explained = false;
};
```

engine/samples/input/main.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+    if (!explained)
+    {
+        CUBOS_WARN("This sample will present multiple showcases of the input system. Press Enter to advance to the "
+                   "next showcase.");
+        explained = true;
+    }
+}
+
+static void showcase_x_or_z(Read<Input> input)
```
RiscadoA, line 29: We use camelCase

RiscadoA (changes requested):

RiscadoA (approved):

RiscadoA (approved):

core/include/cubos/core/data/deserializer.hpp
```diff
@@ -234,4 +234,18 @@ namespace cubos::core::data
...
         des.endDictionary();
     }
+
+    /// Overload for serializing std::pair.
+    /// @tparam V1 The type of the first value.
+    /// @tparam V2 The type of the second value.
+    /// @param des The deserializer.
+    /// @param pair The pair to deserialize.
+    template <typename V1, typename V2>
+    inline void deserialize(Deserializer& des, std::pair<V1, V2>& pair)
```
RiscadoA, line 244:
```suggestion
    /// @tparam T The type of the first value.
    /// @tparam U The type of the second value.
    /// @param des The deserializer.
    /// @param pair The pair to deserialize.
    template <typename T, typename U>
    inline void deserialize(Deserializer& des, std::pair<T, U>& pair)
```

core/include/cubos/core/data/serializer.hpp
```diff
@@ -242,4 +242,19 @@ namespace cubos::core::data
...
+    /// @param pair The pair to serialize.
+    /// @param name The name of the pair (optional).
+    template <typename V1, typename V2>
+    inline void serialize(Serializer& ser, const std::pair<V1, V2>& pair, const char* name)
+    {
+        ser.beginObject(name);
+        ser.write(pair.first, "first");
+        ser.write(pair.second, "second");
+        ser.endObject();
+    }
```
RiscadoA, line 259:
```suggestion
    /// @tparam T The type of the first value.
    /// @tparam U The type of the second value.
    /// @param ser The serializer.
    /// @param pair The pair to serialize.
    /// @param name The name of the pair (optional).
    template <typename T, typename U>
    inline void serialize(Serializer& ser, const std::pair<T, U>& pair, const char* name)
    {
        ser.beginObject(name);
        ser.write(pair.first, "first");
        ser.write(pair.second, "second");
        ser.endObject();
    }
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -34,6 +37,12 @@ namespace cubos::engine
         /// @return True if the action exists and is pressed, false otherwise.
         bool pressed(const char* actionName, int player = 0) const;
 
+        /// Gets a key state
+        /// @param Key The key
+        /// @param modifiers The modifiers (optional)
```
RiscadoA, line 42:
```suggestion
        /// Gets a key state.
        /// @param Key The key.
        /// @param modifiers The modifiers (optional).
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -34,6 +37,12 @@ namespace cubos::engine
...
+        /// @return True if the key is pressed, false otherwise.
+        bool pressed(Key key, Modifiers modifiers = Modifiers::None) const;
+
         /// Gets an axis value for a specific player
         /// @param axisName The name of the axis
         /// @param player The player whose axis value will be retrieved
```
RiscadoA:
```suggestion
        /// Gets an axis value for a specific player.
        /// @param axisName The name of the axis.
        /// @param player The player whose axis value will be retrieved.
```

engine/src/cubos/engine/input/axis.cpp
```diff
@@ -1,25 +1,29 @@
+#include <cubos/core/io/keyboard.hpp>
 #include <cubos/core/log.hpp>
 
 #include <cubos/engine/input/axis.hpp>
 
+using cubos::core::io::Key;
+using cubos::core::io::Modifiers;
+
```
RiscadoA, line 8:
```suggestion
```

engine/src/cubos/engine/input/bindings.cpp
```diff
@@ -77,3 +83,30 @@ void cubos::core::data::deserialize<InputBindings>(Deserializer& des, InputBindi
...
+template <>
+void cubos::core::data::deserialize<Key, Modifiers>(Deserializer& des, std::pair<Key, Modifiers>& obj)
+{
+    std::string str;
+    des.readString(str);
+
+    std::size_t split = str.find_last_of('-');
+    if (split == std::string::npos)
+    {
+        obj = std::pair(stringToKey(str), Modifiers::None);
```
RiscadoA, line 106:
```suggestion
        obj = {stringToKey(str), Modifiers::None};
```

engine/src/cubos/engine/input/bindings.cpp
```diff
@@ -77,3 +83,30 @@ void cubos::core::data::deserialize<InputBindings>(Deserializer& des, InputBindi
...
+    }
+    else
+    {
+        obj = std::pair(stringToKey(str.substr(split + 1)), stringToModifiers(str.substr(0, split + 1)));
```
RiscadoA, line 110:
```suggestion
        obj = {stringToKey(str.substr(split + 1)), stringToModifiers(str.substr(0, split + 1))};
```

engine/src/cubos/engine/input/input.cpp
```diff
@@ -91,6 +91,16 @@ bool Input::pressed(const char* actionName, int player) const
     return aIt->second.pressed();
 }
 
+bool Input::pressed(Input::Key key, Modifiers modifiers) const
+{
+    auto it = mPressedKeys.find(key);
+    if (it != mPressedKeys.end())
```
RiscadoA, line 97:
```suggestion
    if (auto it = mPressedKeys.find(key); it != mPressedKeys.end())
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+    class Input final
+    {
+    public:
+        using Key = core::io::Key;
+        using Modifiers = core::io::Modifiers;
+
+        Input() = default;
+        ~Input() = default;
+
+        /// Clears all bindings
```
RiscadoA, line 22:
```suggestion
        /// Clears all bindings.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        void clear();
+
+        /// Clears all bindings for a specific player
+        /// @param player The player whose bindings will be cleared
```
RiscadoA, line 26:
```suggestion
        /// Clears all bindings for a specific player.
        /// @param player The player whose bindings will be cleared.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        void clear(int player);
+
+        /// Sets the bindings for a specific player
+        /// @param bindings The bindings to set
+        /// @param player The player whose bindings will be set
```
RiscadoA, line 31:
```suggestion
        /// Sets the bindings for a specific player.
        /// @param bindings The bindings to set.
        /// @param player The player whose bindings will be set.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        void bind(const InputBindings& bindings, int player = 0);
+
+        /// Gets an action state for a specific player
+        /// @param actionName The name of the action
+        /// @param player The player whose action state will be retrieved
```
RiscadoA, line 36:
```suggestion
        /// Gets an action state for a specific player.
        /// @param actionName The name of the action.
        /// @param player The player whose action state will be retrieved.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+        /// Gets a key state
+        /// @param Key The key
+        /// @param modifiers The modifiers (optional)
+        /// @return True if the key is pressed, false otherwise.
+        bool pressed(Key key, Modifiers modifiers = Modifiers::None) const;
+
+        /// Gets an axis value for a specific player
+        /// @param axisName The name of the axis
+        /// @param player The player whose axis value will be retrieved
```
RiscadoA, line 48:
```suggestion
        /// Gets an axis value for a specific player.
        /// @param axisName The name of the axis.
        /// @param player The player whose axis value will be retrieved.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        /// Handle a key event
+        /// @param event The key event
+        void handle(const core::io::KeyEvent& event);
+
+        /// Handle a modifiers event
+        /// @param event The modifiers event
+        void handle(const core::io::ModifiersEvent& event);
+
+        /// Discard remaining events.
+        /// @param event The event to discard
```
RiscadoA, line 61:
```suggestion
        /// Handle a key event.
        /// @param event The key event.
        void handle(const core::io::KeyEvent& event);

        /// Handle a modifiers event.
        /// @param event The modifiers event.
        void handle(const core::io::ModifiersEvent& event);

        /// Discard remaining events.
        /// @param event The event to discard.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        inline void handle(const core::io::WindowEvent& event)
+        {
+            (void)event;
+        }
+
+        /// @return The bindings
```
RiscadoA, line 67:
```suggestion
        /// @return The bindings.
```

RiscadoA (changes requested):

engine/include/cubos/engine/input/input.hpp
luishfonseca:
```suggestion
        /// Gets an axis value for a specific player.
        /// @param axisName The name of the axis.
        /// @param player The player whose axis value will be retrieved.
```

engine/include/cubos/engine/input/input.hpp
```diff
@@ -0,0 +1,88 @@
...
+        void handle(const core::io::ModifiersEvent& event);
+
+        /// Discard remaining events.
+        /// @param event The event to discard.
+        inline void handle(const core::io::WindowEvent& event)
+        {
+            (void)event;
+        }
+
+        /// @return The bindings.
```
RiscadoA, line 67:
```suggestion
        /// @return The bindings for each player.
```

RiscadoA (changes requested):

RiscadoA (approved):

luishfonseca:
91 no counter de conversation e comecei isto ontem, assustador

PR #388, opened 2023-05-27
Use CCache

luishfonseca:
Blocked by #589


CMakeLists.txt
```diff
@@ -26,6 +26,29 @@ if(ENABLE_COVERAGE)
...
 endif()
 
+# Enable CCache
+find_program(CCACHE_EXE NAMES "ccache")
+if(CCACHE_EXE)
+    option(USE_CCACHE "Enable CCache" ON)
+    if(USE_CCACHE)
+        if (CMAKE_HOST_WIN32)
+            file(COPY_FILE
+                "C:/ProgramData/chocolatey/lib/ccache/tools/ccache-4.8.3-windows-x86_64/ccache.exe" ${CMAKE_BINARY_DIR}/cl.exe
```
RiscadoA, line 36: :fearful: isn't pinning a version like this dangerous?

RiscadoA (approved):
LGTM, good work :relieved: I'm a bit concerned about that hardcoded path on `CMakeLists.txt`

CMakeLists.txt
luishfonseca: Yes this is disgusting, chocolatey (windows package manager) install ccache on that path and then creates a wrapper. CMake's find_program only finds the wrapper, and sadly it doesn't work.

I should make it slightly less awful by turning the version into a variable, I'll look into it

PR #409, opened 2023-05-30
Move Input plugin's pressed to Window plugin's keyPressed

luishfonseca:
No more need to keep state about the pressed keys in Input plugin

### DIFF
--- a/core/include/cubos/core/io/window.hpp
+++ b/core/include/cubos/core/io/window.hpp
@@ -167,6 +167,12 @@ namespace cubos::core::io
         /// @return Text from the clipboard.
         virtual const char* getClipboard() const = 0;
 
+        /// @return Whether the key is currently pressed.
+        virtual bool keyPressed(Key key) const = 0;
+
+        /// @return Whether the key and modifiers are currently pressed.
+        virtual bool keyPressed(Key key, Modifiers modifiers) const = 0;
+
     protected:
         /// Asks the implementation to fill the event queue with new events.
         virtual void pollEvents() = 0;
--- a/core/include/cubos/core/io/window.hpp
+++ b/core/include/cubos/core/io/window.hpp
@@ -170,7 +170,7 @@ namespace cubos::core::io
 
         /// @param key Key to check.
         /// @param modifiers Modifiers to check.
-        /// @return Whether the key and modifiers are currently pressed.
+        /// @return Whether the key and modifiers (or a superset of) are currently pressed.
         virtual bool pressed(Key key, Modifiers modifiers = Modifiers::None) const = 0;
 
     protected:
--- a/core/include/cubos/core/io/window.hpp
+++ b/core/include/cubos/core/io/window.hpp
@@ -126,27 +126,27 @@ namespace cubos::core::io
         /// Swaps the window buffers.
         virtual void swapBuffers() = 0;
 
-        /// Returns the window render device.
-        virtual gl::RenderDevice& getRenderDevice() const = 0;
+        /// @return The render device associated with this window.
+        virtual gl::RenderDevice& renderDevice() const = 0;
 
-        /// Returns the window size, in screen coordinates.
-        /// May differ from the framebuffer size in some displays.
-        virtual glm::ivec2 getSize() const = 0;
+        /// @return The window size, in screen coordinates. May differ from the framebuffer size.
+        virtual glm::ivec2 size() const = 0;
 
-        /// Returns the window framebuffer size in pixels.
-        virtual glm::ivec2 getFramebufferSize() const = 0;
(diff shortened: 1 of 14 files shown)


core/include/cubos/core/io/window.hpp
```diff
@@ -167,6 +167,12 @@ namespace cubos::core::io
         /// @return Text from the clipboard.
         virtual const char* getClipboard() const = 0;
 
+        /// @return Whether the key is currently pressed.
+        virtual bool keyPressed(Key key) const = 0;
+
+        /// @return Whether the key and modifiers are currently pressed.
+        virtual bool keyPressed(Key key, Modifiers modifiers) const = 0;
```
RiscadoA, line 174: Why not just one with `modifiers` set to None by default?

RiscadoA: Also missing param docs

core/src/cubos/core/io/glfw_window.hpp
```diff
@@ -37,9 +37,15 @@ namespace cubos::core::io
         void setCursor(std::shared_ptr<Cursor> cursor) override;
         void setClipboard(const std::string& text) override;
         const char* getClipboard() const override;
+        bool keyPressed(Key key) const override;
+        bool keyPressed(Key key, Modifiers modifiers) const override;
+
+        Modifiers modifiers() const;
```
RiscadoA, line 43: Why not make this virtual and exposed on `BaseWindow`?

core/src/cubos/core/io/glfw_window.hpp
```diff
@@ -37,9 +37,15 @@ namespace cubos::core::io
...
+        void modifiers(Modifiers modifiers);
```
RiscadoA, line 44: Why would this be used externally?

RiscadoA (changes requested):

core/src/cubos/core/io/glfw_window.hpp
luishfonseca: Querying modifiers is misleading, eg. once you press "Control" the modifiers mask will be still be 0, only when you press the next key it will be Modifiers::Control.

I think it's better not to expose it. GLFW also doesn't expose it, probably for the same reasons.

core/src/cubos/core/io/glfw_window.hpp
luishfonseca: It isn't, but I need it in static functions that receive the window as an argument

core/src/cubos/core/io/glfw_window.cpp
```diff
@@ -274,6 +275,34 @@ const char* GLFWWindow::getClipboard() const
...
+    return glfwGetKey(mHandle, cubosToGlfwKey(key)) == GLFW_PRESS;
+#else
+    UNSUPPORTED();
+#endif
+}
+
+bool GLFWWindow::keyPressed(Key key, Modifiers modifiers) const
+{
+#ifdef WITH_GLFW
+    return this->keyPressed(key) && this->modifiers() == modifiers;
```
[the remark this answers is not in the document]
luishfonseca: This is a good reason as why to use `{ "ctrl-a": [ "C-a" ] }` instead of direct window access.
Input->pressed("ctrl-a") will be true even if you then press shift afterwards, only returning to false once you release A.
Window->pressed(Key::A) will go back as soon as you press other modifiers.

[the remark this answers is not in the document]
RiscadoA: I think the solution is actually adding a proper `modifiers` getter - we just have to update it correctly.
I thought about it again and I think we should go with the bit masking - if the user could query the current modifiers they would be able to handle that part of the logic if they wish to.
It should all be clear in the docs too.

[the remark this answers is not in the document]
RiscadoA: If you want to release `C-a` when `Control` is released, then does it make sense to think in modifiers anymore? I don't think so. You are basically just considering that an action is only triggered when a group of keys is pressed.
You could make it so that instead of modifiers, you just store lists of keys to pressed: `Control+A` `A+B+C` - it could be useful to not be limited to modifiers.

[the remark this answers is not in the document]
RiscadoA: I mean, you could even implement copy paste behaviour by providing yet another input primitive - sequences of keys. E.g., Ctrl must be pressed and only then should A be pressed. I think this is out of scope.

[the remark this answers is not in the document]
luishfonseca: Opened an issue, let's merge this as is

RiscadoA:
You can merge it when CI finishes @luishfonseca.

PR #426, opened 2023-06-17
Add split screen to the renderer

RiscadoA:


engine/include/cubos/engine/renderer/plugin.hpp
```diff
@@ -25,11 +25,12 @@ namespace cubos::engine
...
-    /// Resource which identifies the entity which is the current active camera.
-    /// The entity referenced by this resource must have a `Camera` component.
-    struct ActiveCamera
+    /// @brief Resource which identifies the camera entities to be used by the renderer.
+    struct ActiveCameras
     {
-        core::ecs::Entity entity; ///< The entity which is the active camera.
+        /// @brief Entities which represent the current active cameras. If more than one is set,
+        /// the renderer automatically splits the screen.
+        core::ecs::Entity entities[2];
```
luishfonseca, line 33: We just need two cameras for MOJO, but this should be eventually expanded to account for greater amounts or be well documented that only two are supported.
If that work will be done after this being merged, add a comment in the code and open an issue.

engine/include/cubos/engine/renderer/plugin.hpp
```diff
@@ -46,11 +47,12 @@ namespace cubos::engine
     /// Resources:
     /// - `Renderer`: handle to the renderer.
     /// - `RendererFrame`: holds the current frame information.
-    /// - `ActiveCamera`: the entity which represents the active camera.
+    /// - `RendererEnvironment`: holds the environment information (ambient light, sky gradient).
+    /// - `ActiveCameras`: the entity which represents the active camera.
```
luishfonseca, line 51: This comment should mention multiple cameras

engine/src/cubos/engine/renderer/deferred_renderer.cpp
```diff
@@ -747,6 +776,14 @@ void DeferredRenderer::onRender(const Camera& camera, const RendererFrame& frame
...
         mSsaoTexBp->bind(mSampler);
     }
+    mUVScaleBp->setConstant(
+        glm::vec2((float)camera.viewportSize.x / (float)mSize.x, (float)camera.viewportSize.y / (float)mSize.y));
+    mUVOffsetBp->setConstant(glm::vec2((float)camera.viewportPosition.x / (float)mSize.x,
+                                       (float)camera.viewportPosition.y / (float)mSize.y));
+    mSkyGradientBottomBp->setConstant(frame.skyGradient(0));
+    mSkyGradientTopBp->setConstant(frame.skyGradient(1));
+    mInvVBp->setConstant(glm::inverse(mvp.v));
+    mInvPBp->setConstant(glm::inverse(mvp.p));
```
luishfonseca, line 786: This function explains each step with a comment, here should not be an exception

engine/src/cubos/engine/renderer/plugin.cpp
```diff
@@ -99,21 +100,59 @@ static void framePointLights(Write<RendererFrame> frame, Query<Read<PointLight>,
...
+        cameras[0].viewportPosition = {0, 0};
+        cameras[0].viewportSize = size;
+    }
+    else if (cameraCount == 2)
+    {
+        cameras[0].viewportPosition = {0, 0};
+        cameras[0].viewportSize = {size.x / 2, size.y};
+        cameras[1].viewportPosition = {size.x / 2, 0};
+        cameras[1].viewportSize = {size.x / 2, size.y};
+    }
```
luishfonseca, line 149: Not a fan of this, but we can make it procedural once we have more than two cameras

luishfonseca (changes requested):

engine/src/cubos/engine/renderer/deferred_renderer.cpp
RiscadoA: It's part of the same step as the calls above - setting the pipeline state.

luishfonseca (dismissed):

luishfonseca (approved):

PR #423, opened 2023-06-16
Design and implement reflection

RiscadoA:
- [X] Add reflection base type `Type`.
- [X] Add type kinds `PrimitiveType`, `ObjectType`, `VariantType`, `ArrayType` and `DictionaryType`.
- [X] Add utility macros for implement reflection easily.
- [X] Document code througly and add a reflection sample.
- [x] Achieve 100% coverage or near it for the reflection module.
- [x] Automatically set default and move constructors for compiled types.

Implementing new serialization system will be left to the next PR.

PR #438, opened 2023-06-26
Implement new serialization based on new reflection

RiscadoA:
Depends on #437 and #423.


RiscadoA:
Superseded by new serialization system altogether :)

PR #440, opened 2023-08-23
Switch to m.css for the doxygen documentation

RiscadoA:


docs/pages/1_structure/namespaces.dox
luishfonseca, line 1: Why do we need to declare the namespaces here? shouldn't doxygen read the code to find it? This allows the documentation to desynchronize with the code

RiscadoA: Where should we document them? There isn't a 'root' file for the namespace where it makes sense to put the docs in. One option would be create to a header which only documents the namespace but thats kinda weird.

RiscadoA: I don't think it will be that big of a problem - our namespaces probably won't change that much. On the engine side where things change more quickly, we won't get any more namespaces probably.

luishfonseca: Since I don't see a better solution, lgtm

luishfonseca (approved):

PR #503, opened 2023-09-06
Cleanup and document `engine-sample.cars`

fallenatlas:
Blocked by #508.

- [x] Self-review changes.


engine/samples/cars/components.hpp
```diff
@@ -2,8 +2,10 @@
 
 #include <glm/glm.hpp>
 
+/// [Car component]
 struct [[cubos::component("car", VecStorage)]] Car
```
RiscadoA, line 6:
```suggestion
struct [[cubos::component("car")]] Car
```

The storage isn't needed anymore, defaults to `VecStorage` :relieved:

engine/samples/cars/main.cpp
```diff
@@ -17,30 +17,47 @@ using cubos::core::gl::Grid;
...
 };
+/// [Spawner resource]
 
-static void settings(Write<Settings> settings)
+/// [Set settings]
+static void settingsSystem(Write<Settings> settings)
 {
     settings->setString("assets.io.path", SAMPLE_ASSETS_FOLDER);
 }
+/// [Set settings]
```
RiscadoA, line 39: I think I would omit this from the docs - its not related to the actual sample, and its something we must do on all samples using assets, which shouldn't be needed in a 'real world' scenario.

engine/samples/cars/main.cpp
```diff
@@ -57,34 +74,56 @@ static void setup(Commands cmds, Write<Assets> assets, Write<Renderer> renderer,
...
+    cmds.create()
+        .add(DirectionalLight{glm::vec3(1.0F), 1.0F})
+        .add(LocalToWorld{})
+        .add(Rotation{glm::quat(glm::vec3(glm::radians(45.0F), glm::radians(45.0F), 0))});
 }
+/// [Spawn light]
 
-static void spawn(Commands cmds, Write<Spawner> spawner, Read<DeltaTime> deltaTime, Read<Assets> assets)
+/// [Spawn system]
+static void spawnSystem(Commands cmds, Write<Spawner> spawner, Read<DeltaTime> deltaTime, Read<Assets> assets)
```
RiscadoA, line 110:
```suggestion
/// [Spawn cars system]
static void spawnCarsSystem(Commands cmds, Write<Spawner> spawner, Read<DeltaTime> deltaTime, Read<Assets> assets)
```

engine/samples/cars/main.cpp
```diff
@@ -105,20 +145,34 @@ static void move(Query<Write<Car>, Write<Position>, Write<Rotation>> query, Read
...
     }
 }
+/// [Move system]
 
 int main(int argc, char** argv)
 {
     Cubos cubos{argc, argv};
+    /// [Adding the env settings plugin]
     cubos.addPlugin(envSettingsPlugin);
+    /// [Adding the env settings plugin]
```
RiscadoA, line 155: No need to document this, its not directly related to the functionality being showcased.

engine/samples/cars/main.cpp
```diff
@@ -105,20 +145,34 @@ static void move(Query<Write<Car>, Write<Position>, Write<Rotation>> query, Read
...
     cubos.addResource<Spawner>();
-
-    cubos.startupSystem(settings).tagged("cubos.settings");
-    cubos.startupSystem(setup).tagged("cubos.assets").after("cubos.renderer.init");
-    cubos.system(spawn);
-    cubos.system(move);
+    /// [Adding component and resource]
+
+    /// [Adding systems]
+    cubos.startupSystem(settingsSystem).tagged("cubos.settings");
```
RiscadoA, line 168: Irrelevant system, dont document it

engine/samples/cars/page.md
```diff
@@ -0,0 +1,58 @@
+# Cars {#examples-engine-cars}
+
+@brief Sample that spawns multiple assets and updates them every frame.
```
RiscadoA, line 3:
```suggestion
@brief Loading voxel grids and rendering them.
```

Perhaps? Spawn may suggest erroneously that we're using blueprints in some way or another.

engine/samples/cars/page.md
```diff
@@ -0,0 +1,58 @@
...
+
```
RiscadoA, line 4: Would be nice to include a screenshot of the sample here, so that the reader knows what they are supposed to see.

engine/samples/cars/page.md
```diff
@@ -0,0 +1,58 @@
...
+This example shows how the @ref renderer-plugin plugin, @ref voxels-plugin plugin and @ref env-settings-plugin, can be used in a scene to spawn and move assets.
```
RiscadoA, line 5:
```suggestion
This example shows how the @ref renderer-plugin plugin and @ref voxels-plugin plugin, can be used to load and render voxel objects.
```

engine/samples/cars/page.md
```diff
@@ -0,0 +1,58 @@
...
+
+The first thing we're going to worry about is setting up the path for the assets we're going to use in our settings resource. In this example, the path is set in a macro defined in a CMakeLists.txt file but you can give it the path directly as a string.
```
RiscadoA, line 7: This would be best as a disclaimer in the code I think, but honestly I'm still not sure how to set the path properly

engine/samples/cars/page.md
```diff
@@ -0,0 +1,58 @@
...
+
+@snippet cars/main.cpp Set settings
+
+Alternatively, the path can be set from the command line arguments given when starting cubos. For this we need the @ref env-settings-plugin, included from the @ref engine/env_settings/plugin.hpp header.
+
+@snippet cars/main.cpp Adding the env settings plugin
+
```
RiscadoA, line 14: Ditto

engine/samples/cars/main.cpp
```diff
@@ -17,30 +17,47 @@ using cubos::core::gl::Grid;
...
     // Load the palette asset and add two colors to it.
     auto palette = assets->write(PaletteAsset);
-    auto black = palette->add({{0.1F, 0.1F, 0.1F, 1.0F}});
-    auto white = palette->add({{0.9F, 0.9F, 0.9F, 1.0F}});
+    palette->add({{0.1F, 0.1F, 0.1F, 1.0F}});
+    palette->add({{0.9F, 0.9F, 0.9F, 1.0F}});
 
     // Set the renderer's palette to the one we just modified.
     (*renderer)->setPalette(*palette);
+}
```
RiscadoA, line 51: Perhaps on this case its best to keep this system merged with `spawnFloorSystem`. You end up having less repetition and the code is less confusing, since we don't need to search for the materials again.

RiscadoA (changes requested):
Thanks for working on this @fallenatlas! Added some suggestions

RiscadoA:
Also, I just realized that perhaps we could call this sample `engine-sample.voxels`. It basically demonstrates how the `voxelsPlugin` can be used, right?

fallenatlas:
We could do that but maybe the movement part wouldn't really be needed if we were to rename it, because it doesn't really have to do with voxels or rendering, I think.

RiscadoA:
Yeah you're right. Unsure if we should simplify this or keep it.
One point in favor of removing the movement would be to have a 'go to' sample for loading voxel models.

Still, I guess it can be nice to have a sample which does more than one thing.
Eventually we'll definitely want to have more complete samples which are playable, a bit like our demo.

What do you think? @fallenatlas @luishfonseca @DiogoMendonc-a

fallenatlas:
I think it makes sense to have samples that don't focus only on one thing, but yeah, I don't know if at this stage it's better to just simplify or if something more "normal" like this is actually useful.

RiscadoA:
Well lets simplify it then. Its easier to maintain, and probably more helpful this way - assuming its goal is teaching how to use grid and palette assets.

Maybe the sample should be just a car, without any floor, floating.

RiscadoA:
We discussed this in the face to face meeting and decided that the engine samples should all be 'plugin specific', thus we should remove the movement part and leave only the loading a grid and palette part.

engine/samples/voxels/page.md
```diff
@@ -0,0 +1,30 @@
...
+This example shows how the @ref voxels-plugin plugin, can be used to load voxel grids.
+
+![](voxels/output.png)
+
+The plugin function is included from the @ref engine/voxels/plugin.hpp header.
+To see the asset, you'll also need to perform a basic configuration of the @ref renderer-plugin plugin like shown in the @ref examples-renderer-sample sample.
+
+@snippet voxels/main.cpp Adding the plugin
+
+Lets start by getting the assets we want to use. Each asset has an id through which we can get a handle to it. This id is specified in the .meta file associated to the asset.
```
RiscadoA, line 14:
```suggestion
Lets start by defining the handles of the assets we want to use, as done in the @ref examples-engine-assets samples.
```

engine/samples/voxels/page.md
```diff
@@ -0,0 +1,30 @@
...
+
+@snippet voxels/main.cpp Get handles to assets
+
+Next, let's assign the palette, that we have in our assets folder, to be used by the renderer.
```
RiscadoA, line 18:
```suggestion
In this sample, instead of creating a new palette, we just read the data from the asset identified from the `PaletteAsset` handle we defined previously. Internally, the @ref cubos::engine::Assets "assets manager" will automatically load the palette asset if it hasn't been loaded before.
```

RiscadoA (changes requested):

RiscadoA (dismissed):

RiscadoA (approved):

PR #504, opened 2023-09-07
Cleanup and document `engine-sample.events`

fallenatlas:
- [x] Self-review changes.

### DIFF
--- a/engine/samples/events/main.cpp
+++ b/engine/samples/events/main.cpp
@@ -5,71 +5,87 @@
 
 using namespace cubos::core;
 
+/// [Event struct]
 struct MyEvent
 {
     int value;
 };
+/// [Event struct]
 
 struct State
 {
     int step;
 };
 
+/// [Event reader system]
+static void firstSystem(ecs::EventReader<MyEvent> reader) 
+{
+    for (const auto& event : reader)
+    {
+        CUBOS_INFO("A read {}", event.value);
+    }
+}
+/// [Event reader system]
+
+/// [Event writer system]
+static void secondSystem(ecs::EventWriter<MyEvent> writer, ecs::Write<State> state, ecs::Write<cubos::engine::ShouldQuit> quit)
+{
+    state->step += 1;
+    if (state->step == 1) // Write 1 2 3 on first run.
+    {
+        writer.push({1});
+        writer.push({2});
+        writer.push({3});
+        CUBOS_INFO("B wrote 1 2 3");
+    }
+    else if (state->step == 2)
+    {
+        quit->value = true; // Stop the loop.
+        writer.push({4});
+        writer.push({5});
+        writer.push({6});
+        CUBOS_INFO("B wrote 4 5 6");
+    }
+}
+/// [Event writer system]
+
+static void thirdSystem(ecs::EventReader<MyEvent> reader)
+{
+    for (const auto& event : reader)
+    {
+        CUBOS_INFO("C read {}", event.value);
+    }
+}
+
+static void fourthSystem(ecs::EventReader<MyEvent> reader)
+{
+    for (const auto& event : reader)
+    {
+        CUBOS_INFO("D read {}", event.value);
+    }
+}
+
 int main()
 {
     cubos::engine::Cubos cubos;
     cubos.addResource<State>(State{.step = 0});
-    cubos.addEvent<MyEvent>();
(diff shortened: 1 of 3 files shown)


engine/samples/events/page.md
```diff
@@ -0,0 +1,28 @@
+# Events {#examples-engine-events}
+
+@brief Using the @ref cubos::core::ecs::EventReader and @ref cubos::core::ecs::EventWriter
```
RiscadoA, line 3:
```suggestion
@brief Using the @ref cubos::core::ecs::EventReader "EventReader" and @ref cubos::core::ecs::EventWriter "EventWriter".
```

This way the namespace is hidden while still keeping the reference, becomes easier to read. :relieved:

engine/samples/events/page.md
```diff
@@ -0,0 +1,28 @@
...
+
+This example shows how the @ref cubos::core::ecs::EventReader and the @ref cubos::core::ecs::EventWriter can be used and configured, in a simple scene, to emit and read events. These are in the core, meaning, you don't need to add them to cubos.
```
RiscadoA, line 5:
```suggestion
This example shows how the @ref cubos::core::ecs::EventReader "EventReader" and the @ref cubos::core::ecs::EventWriter "EventWriter" system arguments can be used to communicate from one system to another.
```

engine/samples/events/page.md
```diff
@@ -0,0 +1,28 @@
...
+
+Firstly, we need to create and register the event we want to emit. Here, our event has one variable, however, you can give it any number of variables of any type you want.
```
RiscadoA, line 7:
```suggestion
Firstly, we need to create and register the event we want to emit. Here, our event is a simple struct with a single field, however, you can use any type you want.
```

engine/samples/events/page.md
```diff
@@ -0,0 +1,28 @@
...
+
+@snippet events/main.cpp Event struct
+@snippet events/main.cpp Adding event
+
+To receive these events, we can make a simple system, using the @ref cubos::core::ecs::EventReader and iterating through all the events it has. This will be the layout of all our reader systems (A, C, D).
+
+@snippet events/main.cpp Event reader system
+
+Now, to emit these events, we will use the @ref cubos::core::ecs::EventWriter. This system will emit 3 events on the first frame and another 3 on the second frame. By setting the value of the ShouldQuit resource to true on the second frame, the engine stops before reaching the third frame.
```
RiscadoA, line 16:
```suggestion
Now, to emit these events, we will use the @ref cubos::core::ecs::EventWriter "EventWriter" system argument. This system will emit 3 events on the first frame and another 3 on the second frame. By setting the value of the @ref cubos::engine::ShouldQuit "ShouldQuit" resource to true on the second frame, the engine stops before reaching the third frame. 
```

engine/samples/events/page.md
```diff
@@ -0,0 +1,28 @@
...
+@brief Using the @ref cubos::core::ecs::EventReader and @ref cubos::core::ecs::EventWriter
+
+This example shows how the @ref cubos::core::ecs::EventReader and the @ref cubos::core::ecs::EventWriter can be used and configured, in a simple scene, to emit and read events. These are in the core, meaning, you don't need to add them to cubos.
+
+Firstly, we need to create and register the event we want to emit. Here, our event has one variable, however, you can give it any number of variables of any type you want.
+
+@snippet events/main.cpp Event struct
+@snippet events/main.cpp Adding event
+
+To receive these events, we can make a simple system, using the @ref cubos::core::ecs::EventReader and iterating through all the events it has. This will be the layout of all our reader systems (A, C, D).
```
RiscadoA, line 12:
```suggestion
To receive these events, we can make a simple system which takes the @ref cubos::core::ecs::EventReader "EventReader" system argument and iterates through all the events it has. This will be the layout of all our reader systems (A, C, D).
```

engine/samples/events/main.cpp
```diff
@@ -5,71 +5,85 @@
 
 using namespace cubos::core;
```
RiscadoA, line 6: I think replacing this with `using cubos::core::ecs::Blabla` for each used type would be more useful.
Then you could also show in the sample docs that `using cubos::core::ecs::EventReader/EventWriter` is being used for convenience.

RiscadoA (changes requested):
Thanks for working on this @fallenatlas!
Overall LGTM, just added some minor suggestions.

RiscadoA (approved):

PR #508, opened 2023-09-10
Cleanup and document `engine-sample.assets`

RiscadoA:
# Description

- Removed the asset creation/saving code from the `assets.json` sample, moved it to a new `assets.saving` sample.
- Wrote documentation for all three asset samples.

## Checklist

- [x] Self-review changes.


engine/samples/assets/json/page.md
```diff
@@ -0,0 +1,26 @@
+# Loading Serializable Assets {#examples-engine-assets-json}
+
+@brief Loading serializable assets from JSON.
+
+We'll use the following type as an example:
+
+@snippet assets/json/main.cpp Asset type
+
+We can make it serializable by implementing the following specializations:
```
luishfonseca, line 9: Explain what the `const char* name` argument is and how it should be used

luishfonseca (changes requested):

engine/samples/assets/json/page.md
RiscadoA: No use documenting/explaining serialization since we should get rid of it quickly.

luishfonseca (approved):

PR #555, opened 2023-09-17
Move `cubos::core::Settings` to engine

RiscadoA:
# Description

Merges `envSettingsPlugin` and `fileSettingsPlugin` into a single `settingsPlugin`.
Moves `Settings` from core to the `settingsPlugin`.
Changes `Cubos` to always add the resource `Arguments`, even if `argc, argv` isn't passed to its constructor.

## Checklist

- [x] Self-review changes.


engine/src/cubos/engine/settings/plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    return settings;
+}
+
+static Settings loadFromFile(std::string_view path)
+{
+    // Mount the real settings file as `/settings.json`.
+    FileSystem::mount("/settings.json",
+                      std::make_unique<StandardArchive>(path, false /*isDirectory*/, false /*readOnly*/));
```
luishfonseca, line 47: does this behave well if there is no settings file at path?

luishfonseca (changes requested):
As we are making this a plugin it should have a sample and docs page

engine/src/cubos/engine/settings/plugin.cpp
RiscadoA: Yeah, `StandardArchive` docs states that it creates an empty file if there isn't one.

luishfonseca (approved):

PR #562, opened 2023-09-21
Cleanup and document `engine-sample.scene`

DiogoMendonc-a:


engine/samples/scene/page.md
```diff
@@ -0,0 +1,64 @@
...
+
+Now that we have our scene file, let's get CUBOS. to load it.
+The first thing we're going to need is a reference to the scene asset. 
+For the purposes of this sample we can simply use an hardcoded reference to the asset.
+
+@snippet scene/main.cpp Setting the asset
+
+Then we'll need a system that spawns that scene.
+To do this we simply get the Scene object from the asset, and then spawn its entities.
+`commands.spawn` will create in the world a copy of every entitiy defined in the scene.
```
RiscadoA, line 46:
```suggestion
@ref cubos::core::ecs::Commands::spawn "Commands::spawn" will create in the world a copy of every entity defined in the scene's blueprint.
```

engine/samples/scene/page.md
```diff
@@ -0,0 +1,64 @@
...
+It won't remove the entities already there, so if you want to close a scene, you'll have to do it yourself.
+
+@snippet scene/main.cpp Spawning the scene
+
+And now we'll have this system run at startup.
+It is not required, you can spawn scenes at any point in the execution of the program,
+but for the purpose of demonstrating the functionality, spawning once at startup is enough.
```
RiscadoA, line 53:
```suggestion
In this case, we'll run this system at startup, since we want to spawn it a single time.
Since it's a startup system, we'll have to tag it with `cubos.assets` to make sure it runs only after the scene bridge has been registered.
On a real game, you could have, for example, a scene for an enemy which you spawn multiple times, instead of just once at startup.
```

engine/samples/scene/page.md
```diff
@@ -0,0 +1,64 @@
...
+
+@snippet scene/main.cpp Adding the system
+
+Finally, let's print the loaded scene, to check that everything is working correctly.
+
+@snippet scene/main.cpp Displaying the scene
```
RiscadoA, line 59: I'm not sure if we should show this here...
It's kinda scary and not related to the scene at all.
Maybe we should just leave it in the sample and omit it in the docs, like just saying that the samples prints the result but not show any code related to it.
What do you think?

engine/samples/scene/page.md
```diff
@@ -0,0 +1,64 @@
...
+
+If you run the sample, it should give you a list that has:
+- an entity with `num` set to 0, with no parent. This is the `main` entity.
+- two entities with `num` set to 1, with same parent, who has `num` set to 0. These are the `root` entities of each instance of the subscene.
+- two entities with `num` set to 2, with different parents, but both of them having `num` set to 1.  These are the `child` entities of each instance of the subscene.
```
RiscadoA, line 64: Maybe its best to not go into so much detail, its easier for it to become outdated.
Should be enough to mention that if you run the sample it will print the spawned entities.

RiscadoA (changes requested):
Thank you for working on this @DiogoMendonc-a!
It would also be nice to reference this example in the scene plugin documentation, as we did with the renderer plugin.

engine/samples/scene/page.md
DiogoMendonc-a: I guess we could, I just find it w bit weird ending a tutorial page on an instruction, instead of on something the user can actually see to check if everything is working right

engine/samples/scene/page.md
DiogoMendonc-a: For this to become outdated the sample itself would have to be altered. I put so much detail because the output is not the easiest to read, you have to go from index to nums.

engine/samples/scene/page.md
RiscadoA: Why not then keep the output description and hide the print implementation?

DiogoMendonc-a: Sure

RiscadoA (approved):
Ship it 🚢

PR #592, opened 2023-09-26
Move voxel stuff to engine

RiscadoA:
# Description

Move `core::gl::Vertex` and `core::gl::triangulate` to the `rendererPlugin`.
Move `parseQB` to `quadrados convert`.
Move `core::gl::Grid`, `core::gl::Palette` and `core::gl::Material` to the `voxelsPlugin`.

## Checklist

- [x] Self-review changes.


RiscadoA:
Since these types are now in the `engine` namespace instead of `core::gl`, maybe we should change their names to something more descriptive such as `VoxelGrid`, `VoxelMaterial`, `VoxelPalette`, `VoxelVertex`. What do you think?

luishfonseca:
> Since these types are now in the `engine` namespace instead of `core::gl`, maybe we should change their names to something more descriptive such as `VoxelGrid`, `VoxelMaterial`, `VoxelPalette`, `VoxelVertex`. What do you think?

That would be better, yes

luishfonseca (changes requested):
`Grid`, `Material` and `Palette` should be prefixed with `Voxel`. For `Vertex`, I think it's ok to leave it as is.

RiscadoA:
> `Grid`, `Material` and `Palette` should be prefixed with `Voxel`. For `Vertex`, I think it's ok to leave it as is.

The vertex type is specific for voxels - it stores a palette material index.
Prolly should also prefix it.

luishfonseca (approved):

PR #594, opened 2023-09-26
Move UI Code to the Engine

luishfonseca:


RiscadoA:
I think the imgui dependency should be moved entirely to the engine side, under the imgui plugin. Whats your opinion on that? Its not as if we're wrapping it in the core.

luishfonseca:
That will be my next commit

engine/include/cubos/engine/tools/utils/serialization.hpp
```diff
@@ -1,12 +1,17 @@
+/// @dir
```
RiscadoA, line 1: I would move this to `engine/imgui/serialization.hpp`

RiscadoA (changes requested):

engine/include/cubos/engine/imgui/imgui.hpp
```diff
@@ -1,19 +1,23 @@
 /// @file
```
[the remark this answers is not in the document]
luishfonseca: I'll move the header file to engine/src

RiscadoA:
I assume this is no longer draft?

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -16,8 +18,8 @@ namespace cubos::core::ui
     ///
     /// @param pkg Packaged object to show.
     /// @param name Name of the object.
-    /// @ingroup core-ui
-    void showPackage(const data::Package& pkg, const std::string& name);
+    /// @ingroup utils-tool-plugin
```
RiscadoA, line 21:
```suggestion
    /// @ingroup imgui-plugin
```

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -1,12 +1,14 @@
...
+/// @ingroup imgui-plugin
 
 #pragma once
 
 #include <cubos/core/data/package.hpp>
 
-namespace cubos::core::ui
+using Package = cubos::core::data::Package;
+
+namespace cubos::engine::tools
```
RiscadoA, line 11:
```suggestion
namespace cubos::engine
```
imguiPlugin isn't part of the tools namespace.

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -1,12 +1,14 @@
...
 /// @brief Functions for showing and editing serializable objects in the UI.
-/// @ingroup core-ui
+/// @ingroup imgui-plugin
 
 #pragma once
 
 #include <cubos/core/data/package.hpp>
 
-namespace cubos::core::ui
+using Package = cubos::core::data::Package;
```
RiscadoA, line 9: Don't use `using` in headers.

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -29,8 +31,8 @@ namespace cubos::core::ui
     /// @param pkg Packaged object to edit.
     /// @param name Name of the object.
     /// @return True if the object was modified, false otherwise.
-    /// @ingroup core-ui
-    bool editPackage(data::Package& pkg, const std::string& name);
+    /// @ingroup utils-tool-plugin
```
RiscadoA, line 34:
```suggestion
    /// @ingroup imgui-plugin
```

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -40,11 +42,11 @@ namespace cubos::core::ui
     /// @tparam T Type of the serializable object.
     /// @param object Object to show.
     /// @param name Name of the object.
-    /// @ingroup core-ui
+    /// @ingroup utils-tool-plugin
```
RiscadoA, line 45:
```suggestion
    /// @ingroup imgui-plugin
```

engine/include/cubos/engine/imgui/serialization.hpp
```diff
@@ -58,11 +60,11 @@ namespace cubos::core::ui
     /// @param object Object to edit.
     /// @param name Name of the object.
     /// @return True if the object was edited, false otherwise.
-    /// @ingroup core-ui
+    /// @ingroup utils-tool-plugin
```
RiscadoA, line 63:
```suggestion
    /// @ingroup imgui-plugin
```

engine/src/cubos/engine/imgui/imgui.hpp
RiscadoA, line 1: I would prefix these function names with `imgui` or something, otherwise its very easy to get linking clashes with other functions.

engine/src/cubos/engine/imgui/imgui.hpp
```diff
@@ -1,19 +1,23 @@
...
 
 #pragma once
 
 #include <cubos/core/gl/render_device.hpp>
 #include <cubos/core/io/window.hpp>
 
-namespace cubos::core::ui
+using Framebuffer = cubos::core::gl::Framebuffer;
+using Window = cubos::core::io::Window;
+using WindowEvent = cubos::core::io::WindowEvent;
```
RiscadoA, line 12: Although this is a 'private' header, I think we should follow the same standards as other headers and thus not use `using` anywhere here.

RiscadoA (changes requested):
Good work! Just needs some naming/styling changes.

engine/src/cubos/engine/imgui/plugin.cpp
```diff
@@ -1,17 +1,18 @@
-#include <cubos/core/ui/imgui.hpp>
-
 #include <cubos/engine/imgui/plugin.hpp>
 #include <cubos/engine/window/plugin.hpp>
 
+#include "imgui.hpp"
```
RiscadoA, line 4: Use full include path for consistency.

RiscadoA (changes requested):

engine/src/cubos/engine/imgui/plugin.cpp
luishfonseca: not quite, broad_phase.cpp also includes without full path. As these are next to each other I would prefer relative paths

RiscadoA (approved):

Issue #602, opened 2023-09-27
Add formatting pre-commit hook 


RiscadoA:
# Problem
Currently our fornatting workflow creates styling commits.
This clutters up the git history.
The perfect solution would be each commit producing formatted code, but not everyone has their formatter setup to format on save.

# Solution
We could add a pre-commit hook which checks if the staged code is formatted, and if not, asks the dev if they want:
- to cancel the commit;
- to format the code automatically;
- to commit anyway.

Either:
- we couple this with #601, such that unformatted code is just formatted automatically on main.
- we add an action which rejects unformatted code.

PR #605, opened 2023-09-27
Add formatting pre-commit hook

luishfonseca:
# Description

Adds auto format pre commit hook and fails unformatted PRs

## Checklist

- [x] Self-review changes.


docs/pages/1_getting_started/main.md
```diff
@@ -39,6 +39,15 @@ from the root directory of the repository:
 git submodule update --init --recursive
 ```
 
+### Formatting
+
+To contribute you must ensure your code is correctly formatted. Install
+`clang-format` and run the following command to enable the pre-commit hook:
```
RiscadoA, line 45:
```suggestion
`clang-format` and run the following command to enable a pre-commit hook
which formats your changes automatically:
```

.githooks/pre-commit
RiscadoA, line 1: I would prefer if it warned the user first that their changes are unformatted, and ask for permission to format them, as described in #602. What do you think?

RiscadoA (changes requested):

.githooks/pre-commit
luishfonseca: That's more infra intensive and breaks gui git clients like vscode.
If the user refuses to format it'll get rejected by the cli anyway so they might as well not commit it.

RiscadoA (approved):

PR #617, opened 2023-09-28
Bump lib dependencies

luishfonseca:
# Description

Bumped a bunch of dependencies

## Checklist

- [x] Self-review changes.


RiscadoA:
Don't .forget to update the [getting started page](https://gamedevtecnico.github.io/cubos/getting-started.html) dependency table.

RiscadoA:
What's the reasoning behind removing the documentation deps?

core/CMakeLists.txt
```diff
@@ -93,8 +93,10 @@ cubos_common_target_options(cubos-core)
 # Link dependencies
 
 if (WITH_OPENGL)
-    add_subdirectory(lib/glad)
-    target_link_libraries(cubos-core PRIVATE glad)
+    set(GLAD_SOUURCES_DIR "lib/glad")
+    add_subdirectory("${GLAD_SOUURCES_DIR}/cmake" glad_cmake SYSTEM)
+    glad_add_library(glad REPRODUCIBLE API gl:core=3.3)
+    target_link_libraries(cubos-core PUBLIC glad)
```
RiscadoA, line 99: Why PUBLIC instead of PRIVATE?

RiscadoA (changes requested):

core/CMakeLists.txt
luishfonseca: Just copied it from the glad documentation, I don't actually know how cmake works

RiscadoA: By making it PUBLIC you're making dependencies of CUBOS. have it on their include dir too.
Previously it was PRIVATE, try setting it back and see if anything breaks.

RiscadoA (approved):

PR #654, opened 2023-10-02
Do not read from write only light buffer mapping

### LINKED ISSUE #653: Voxel objects rendered red in samples in some GPUs
# Description

In some machines all voxel objects are being rendered as if only the red channel was available.
Surprisingly, it doesn't apply to the background of the renderer sample.

# Reproduce

Confirmed to happen at least in some of models of Intel UHD graphics.

# Expected behaviour
Should show all colors.

# Screenshots
image

# Additional context
After investigating a capture made using RenderDoc I found that the light buffer had zero lights, except that the ambient light was set to `(1, 0, 0, 0)`. This means that the red color we're seeing is just the object under a red light.

RiscadoA:
# Description

Attempts to fix #653.
Turns out we were reading from a supposedly WRITE ONLY buffer. I suspect this might have caused the memory to be corrupted, which led to the ambient lighting being set to (1, 0, 0, 0), and all other values to 0.

Edit: turns out it didn't fix it :(

## Checklist

- [x] Self-review changes.
- [x] ~~Test that it actually fixes the bug.~~

### DIFF
--- a/engine/src/cubos/engine/renderer/deferred_renderer.cpp
+++ b/engine/src/cubos/engine/renderer/deferred_renderer.cpp
@@ -642,9 +642,10 @@ void DeferredRenderer::onRender(const glm::mat4& view, const Viewport& viewport,
     lightData.ambientLight = glm::vec4(frame.ambient(), 1.0F);
 
     // Spotlights.
+    uint32_t lightI = 0;
     for (const auto& [transform, light] : frame.spotLights())
     {
-        if (lightData.numSpotLights >= CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT)
+        if (lightI >= CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT)
         {
             CUBOS_WARN("Number of spot lights to be drawn "
                        "this frame exceeds the maximum allowed ({}).",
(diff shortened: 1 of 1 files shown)


engine/src/cubos/engine/renderer/deferred_renderer.cpp
```diff
@@ -670,13 +673,15 @@ void DeferredRenderer::onRender(const glm::mat4& view, const Viewport& viewport,
...
+        lightData.directionalLights[lightI].intensity = light.intensity;
+        lightI += 1;
     }
+    lightData.numDirectionalLights = lightI;
 
     // Point lights.
+    lightI = 0;
     for (const auto& [transform, light] : frame.pointLights())
     {
         if (lightData.numPointLights >= CUBOS_DEFERRED_RENDERER_MAX_POINT_LIGHT_COUNT)
```
RiscadoA, line 687:
```suggestion
        if (lightI >= CUBOS_DEFERRED_RENDERER_MAX_POINT_LIGHT_COUNT)
```

engine/src/cubos/engine/renderer/deferred_renderer.cpp
```diff
@@ -650,16 +651,18 @@ void DeferredRenderer::onRender(const glm::mat4& view, const Viewport& viewport,
...
+        lightData.spotLights[lightI].spotCutoff = glm::cos(light.spotAngle);
+        lightI += 1;
     }
+    lightData.numSpotLights = lightI;
 
     // Directional lights.
+    lightI = 0;
     for (const auto& [transform, light] : frame.directionalLights())
     {
         if (lightData.numDirectionalLights >= CUBOS_DEFERRED_RENDERER_MAX_DIRECTIONAL_LIGHT_COUNT)
```
RiscadoA, line 668:
```suggestion
        if (lightI >= CUBOS_DEFERRED_RENDERER_MAX_DIRECTIONAL_LIGHT_COUNT)
```

engine/src/cubos/engine/renderer/deferred_renderer.cpp
```diff
@@ -640,6 +640,7 @@ void DeferredRenderer::onRender(const glm::mat4& view, const Viewport& viewport,
     lightData.ambientLight = glm::vec4(frame.ambient(), 1.0F);
 
     // Spotlights.
+    uint32_t lightI = 0;
     for (const auto& [transform, light] : frame.spotLights())
     {
         if (lightData.numSpotLights >= CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT)
```
RiscadoA, line 646:
```suggestion
        if (lightI >= CUBOS_DEFERRED_RENDERER_MAX_SPOT_LIGHT_COUNT)
```

PR #664, opened 2023-10-04
fix(meta): fixed `.clang-format`

roby2014:
# Description

Removed duplicate key `SortIncludes` mapping.

## Checklist

- [X] Self-review changes.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,6 @@
 ---
 BasedOnStyle: Microsoft
 NamespaceIndentation: All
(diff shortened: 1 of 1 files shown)


RiscadoA:
Usually for commits copes we use the B-label names (e.g., build, rendering, window, physics), or if those do not apply, the A-label names (e.g. meta, engine, core). In this case, instead of `misc`, you should have used `meta`. Dw about this time tho, just something to take note of :relieved:

PR #666, opened 2023-10-04
Split Collider and ColliderShape

luishfonseca:
# Description

There's shared behavior between colliders, let's make use of it.

## Checklist

- [x] Self-review changes.


engine/include/cubos/engine/collisions/aabb.hpp
RiscadoA, line 1: You forgot to change the docs at the top of the file.

RiscadoA: Also, why don't we have AABBs on core/geom?

engine/include/cubos/engine/collisions/shapes/capsule.hpp
```diff
@@ -0,0 +1,17 @@
...
+/// @brief Component @ref cubos::engine::CapsuleCollisionShape.
+/// @ingroup collisions-plugin
+
+#pragma once
+
+#include <cubos/core/geom/capsule.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component which adds a capsule collision shape to an entity, used with a collider component.
```
RiscadoA, line 11:
```suggestion
    /// @brief Component which adds a capsule collision shape to an entity, used with a @ref Collider component.
```

engine/include/cubos/engine/collisions/shapes/box.hpp
```diff
@@ -0,0 +1,17 @@
...
+/// @brief Component @ref cubos::engine::BoxCollisionShape.
+/// @ingroup collisions-plugin
+
+#pragma once
+
+#include <cubos/core/geom/box.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component which adds a box collision shape to an entity, used with a collider component.
```
RiscadoA, line 11:
```suggestion
    /// @brief Component which adds a box collision shape to an entity, used with a @ref Collider component.
```

engine/include/cubos/engine/collisions/collider.hpp
```diff
@@ -0,0 +1,29 @@
...
+    {
+        glm::mat4 transform{1.0F}; ///< Transform of the collider.
+
+        ColliderAABB localAABB; ///< Local space AABB of the collider.
+        ColliderAABB worldAABB; ///< World space AABB of the collider.
+
+        /// @brief Margin of the collider.
+        ///
+        /// When the collider shape has sharp edges, a margin is needed.
+        float margin;
```
RiscadoA, line 25: Shouldn't we provide a sane default? Otherwise the user will have to specify it or just leave it to the will of the UB god.

RiscadoA (changes requested):

engine/include/cubos/engine/collisions/collider.hpp
luishfonseca: I'm setting it in the plugin, based on the associated shape. It isn't meant to be modified by users.

RiscadoA: Then it should be documented as such.

engine/include/cubos/engine/collisions/aabb.hpp
luishfonseca: moved it

engine/include/cubos/engine/collisions/collider.hpp
luishfonseca: done

core/include/cubos/core/geom/aabb.hpp
```diff
@@ -0,0 +1,114 @@
+/// @file
+/// @brief Component @ref cubos::core::geom::AABB.
+/// @ingroup core-geom
+
+#pragma once
+
+#include <glm/glm.hpp>
```
RiscadoA, line 7: Use vec3.hpp

core/include/cubos/core/geom/aabb.hpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// @brief Sets the maximum point of the AABB.
+        /// @param max Maximum point of the AABB.
+        void max(const glm::vec3& max)
+        {
+            diag[1] = max;
+        }
+
+        /// @brief Gets a @ref Box representation of the AABB.
```
RiscadoA, line 48: Why do we have both aabbs and boxes? Im confused, how are they different

luishfonseca: The AABB is an axis aligned box, this allow only two points to be stored and for quick intersection tests. Sometimes we need to create general box shape from an AABB and that's what this function is for.

luishfonseca: But they have actually the same internal representation, I could just merge them

luishfonseca: Actually nvm, the Box will be always centered on the collider transform. The AABB can be lopsided.

RiscadoA (approved):

PR #694, opened 2023-10-08
Implement reflection for all components

RiscadoA:
# Description

Implements reflection for all components and for their fields.
Adds some utilities to make this easier.

Read this commit by commit, I split it nicely :relieved: 

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] ~~Write new samples.~~


core/include/cubos/core/ecs/component/reflection.hpp
```diff
@@ -0,0 +1,60 @@
...
+    public:
+        /// @brief Constructs.
+        /// @param name Component type name, including namespace.
+        ComponentTypeBuilder(std::string name)
+            : mType(reflection::Type::create(std::move(name)))
+        {
+            mType.with(reflection::ConstructibleTrait::typed<T>()
+                           .withDefaultConstructor()
+                           .withCopyConstructor()
+                           .withMoveConstructor()
```
luishfonseca, line 32: Adding these 3 constructors is so frequent we should have a short hand for them, maybe `.withBasicConstructors()`?

core/include/cubos/core/ecs/component/reflection.hpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+            mType.with(std::move(mFields));
+            return mType;
+        }
+
+    private:
+        reflection::Type& mType;
+        reflection::FieldsTrait mFields;
+    };
+} // namespace cubos::core::ecs
```
luishfonseca, line 60: Missing trailing new line

core/include/cubos/core/reflection/traits/fields.hpp
```diff
@@ -54,19 +54,30 @@ namespace cubos::core::reflection
         /// @param type Field type.
         /// @param name Field name.
         /// @param addressOf Getter for the address of the field on a given instance.
+        void addField(const Type& type, std::string name, AddressOf* addressOf);
+
+        /// @copydoc addField(const Type&, std::string, AddressOf*)
         /// @return Trait.
         FieldsTrait&& withField(const Type& type, std::string name, AddressOf* addressOf) &&;
```
luishfonseca, line 61: Why is this needed? The templated `withField` calls `addField`

luishfonseca (changes requested):

core/include/cubos/core/reflection/traits/fields.hpp
RiscadoA: We use in the `ComponentTypeBuilder` to avoid having to do this:
```cpp
mFieldsTrait = std::move(mFieldsTrait).withField(...);
```
This would mean having to implement the assignment operator for `FieldsTrait`. This way we can just do:
```cpp
mFieldsTrait.addField(...);
```

luishfonseca: The review was about the 3 argument `withField`, as I tought it was dead code as both templated `addField` and `withField` used the 3 argument `addField`.
This isn't the case, glm reflection uses 3 argument `withField` so it makes sense to keep it.

luishfonseca (approved):

PR #696, opened 2023-10-08
Add `UnorderedBimap`

RiscadoA:
# Description

Adds `UnorderedBimap`.

## Checklist

- [x] Self-review changes.
- [X] Ensure test coverage.
- [X] ~~Write new samples.~~

### DIFF
--- a/core/include/cubos/core/memory/unordered_bimap.hpp
+++ b/core/include/cubos/core/memory/unordered_bimap.hpp
@@ -24,7 +24,7 @@ namespace cubos::core::memory
         /// @note If any of the values already exists, the old entries with them are removed.
         /// @param left Left value.
         /// @param right Right value.
-        void add(L left, R right)
+        void insert(L left, R right)
         {
             auto leftIt = mLeftToRight.find(left);
             if (leftIt != mLeftToRight.end())
@@ -45,7 +45,7 @@ namespace cubos::core::memory
         /// @brief Removes the entry associated to the given left value.
         /// @param left Left value.
         /// @return Whether the entry was removed.
-        bool removeLeft(const L& left)
+        bool eraseLeft(const L& left)
         {
             auto it = mLeftToRight.find(left);
             if (it == mLeftToRight.end())
@@ -61,7 +61,7 @@ namespace cubos::core::memory
         /// @brief Removes the entry associated to the given right value.
         /// @param right Right value.
         /// @return Whether the entry was removed.
-        bool removeRight(const R& right)
+        bool eraseRight(const R& right)
         {
             auto it = mRightToLeft.find(right);
             if (it == mRightToLeft.end())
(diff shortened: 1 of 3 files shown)


core/include/cubos/core/memory/unordered_bimap.hpp
```diff
@@ -0,0 +1,164 @@
...
+            if (leftIt != mLeftToRight.end())
+            {
+                mRightToLeft.erase(leftIt->second);
+            }
+
+            auto rightIt = mRightToLeft.find(right);
+            if (rightIt != mRightToLeft.end())
+            {
+                mLeftToRight.erase(rightIt->second);
+            }
```
luishfonseca, line 39: doesn't insert or assign already cover this?

core/include/cubos/core/memory/unordered_bimap.hpp
```diff
@@ -0,0 +1,164 @@
...
+            return mLeftToRight.find(left) != mLeftToRight.end();
+        }
+
+        /// @brief Checks if the map contains the given right value.
+        /// @param right Right value.
+        /// @return Whether the map contains the value.
+        bool hasRight(const R& right) const
+        {
+            return mRightToLeft.find(right) != mRightToLeft.end();
+        }
```
luishfonseca, line 105: Why use `.find` and not `.contains`

core/include/cubos/core/memory/unordered_bimap.hpp
luishfonseca, line 1: When implementing these structures its better to be consistent with STL on naming, to avoid guessing. So:
`add` -> `insert_or_assign`
`remove(Left|Right)` -> `erase(Left|Right)`
`has[Left|Right]` -> `contains[Left|Right]`
`get(Left|Right)` -> `at(Left|Right)`

Maybe also wrap `find`?

luishfonseca (approved):

luishfonseca (changes requested):
this is what i meant to do the first time

core/include/cubos/core/memory/unordered_bimap.hpp
RiscadoA: Nope, this handles situations like having (1, a), (2, b) and then inserting (1, b).

 If you had only insert_and_emplace, the result would be:
 
## Initial state

### L to R

(1, a)
(2, b)

### R to L

(a, 1)
(b, 2)

## Final state

### L to R

(1, b)
(2, b) <- should have been removed

### R to L

(a, 1) <- should have been removed
(b, 1)

core/include/cubos/core/memory/unordered_bimap.hpp
RiscadoA: Assumed not all compilers would support it for some reason. will switch

luishfonseca (approved):

PR #701, opened 2023-10-08
Merging transform components into a single one

luishfonseca:
# Description

Unifies Position, Scale, Rotation and LocalToWorld components in a single Transform component

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


RiscadoA:
I think we can drop this. Doesn't seem to make sense to go in this direction with the new query stuff we got + observers.

PR #703, opened 2023-10-08
feat(memory): use reflection on TypeMap

RiscadoA:
# Description

Changes `TypeMap` to use reflection instead of `std::type_index`.
Had to change existing ECS code which depended on this use the reflection type instead of `std::type_index`.
Most of that code will be removed or refactored on a later PR.

## Checklist

- [x] Self-review changes.
- [X] Ensure test coverage.
- [X] ~~Write new samples.~~

### DIFF
--- a/core/include/cubos/core/ecs/component/manager.hpp
+++ b/core/include/cubos/core/ecs/component/manager.hpp
@@ -13,6 +13,7 @@
 #include <typeindex>
 
 #include <cubos/core/ecs/component/storage.hpp>
+#include <cubos/core/memory/type_map.hpp>
 
 namespace cubos::core::ecs
 {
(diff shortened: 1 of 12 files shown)


core/include/cubos/core/ecs/component/manager.hpp
```diff
@@ -187,18 +188,16 @@ namespace cubos::core::ecs
...
-        std::unordered_map<std::type_index, std::size_t> mTypeToIds;
-
-        std::vector<Entry> mEntries; ///< Registered component storages.
+        memory::TypeMap<std::size_t> mTypeToIds; ///< Maps component types to component IDs.
+        std::vector<Entry> mEntries;             ///< Registered component storages.
     };
 
     // Implementation.
 
     template <typename T>
```
luishfonseca, line 197: Why don't we enforce the concept in templates like these?

RiscadoA: We should, but these things will all be refactored in a later PR as to not clutter this one.
The ECS still uses `std::type_index` in some places, here I just did the bare minimum changes to make it compile.

luishfonseca (approved):

Issue #270, opened 2022-12-14
Add palette editor


RiscadoA:
Add a plugin which allows the user to open inspect/edit a palette asset.
Palettes should also be automatically reloaded by the renderer when they change (maybe this is already done?)

PR #711, opened 2023-10-11
feat(engine): add push method to VoxelPalette

roby2014:
# Description

## Checklist

- [X] Self-review changes.
- [X] Ensure test coverage.
- [ ] Write new samples. (#270 will do it)

### DIFF
--- a/engine/include/cubos/engine/voxels/palette.hpp
+++ b/engine/include/cubos/engine/voxels/palette.hpp
@@ -80,6 +80,12 @@ namespace cubos::engine
         /// @return Index of the material in the palette (1-based, 0 is empty).
         uint16_t add(const VoxelMaterial& material, float similarity = 1.0F);
 
+        /// @brief Pushes a material to the palette without checking for uniqueness.
+        /// @note If the palette is already at its maximum capacity, this function will abort.
+        /// @param material Material to push.
+        /// @return Size of the palette.
+        uint16_t push(const VoxelMaterial& material);
+
         /// @brief Merges another palette into this one.
         /// @note All materials equal to @ref VoxelMaterial::Empty will be considered empty and may be
         /// overwritten.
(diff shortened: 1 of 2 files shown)


engine/include/cubos/engine/voxels/palette.hpp
```diff
@@ -80,6 +80,11 @@ namespace cubos::engine
         /// @return Index of the material in the palette (1-based, 0 is empty).
         uint16_t add(const VoxelMaterial& material, float similarity = 1.0F);
 
+        /// @brief Pushes a material to the palette, without checking for uniqueness..
```
RiscadoA, line 83:
```suggestion
        /// @brief Pushes a material to the palette without checking for uniqueness.
```

engine/src/cubos/engine/voxels/palette.cpp
```diff
@@ -100,6 +100,18 @@ uint16_t VoxelPalette::add(const VoxelMaterial& material, float similarity)
     return this->size();
 }
 
+uint16_t VoxelPalette::push(const VoxelMaterial& material)
+{
+    if (this->size() == UINT16_MAX)
+    {
+        CUBOS_ERROR("Cannot add new material: palette is full");
+        return -1;
```
RiscadoA, line 108:
```suggestion
        CUBOS_FAIL("Cannot add new material: palette is full");
```

Aborting here is okay, but we should add a `@note` to the docs warning of it.

RiscadoA (changes requested):
Thanks for working on this! Replaced clang-tidy's review by my own since the suggestion will fix the warning anyway

RiscadoA:
LGTM :) please squash the commits before mergung: you can do so by using git rebase -i, switch pick to fixup on the last commit and the running git push --force-with-lease

RiscadoA (approved):

PR #713, opened 2023-10-11
Implement Initial Particle Physics

fallenatlas:
# Description

Physics for bodies that have no rotation and are just a point in space.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


engine/src/cubos/engine/physics/plugin.cpp
```diff
@@ -0,0 +1,63 @@
...
+    if (physicsMass.inverseMass <= 0.0f) return;
+
+    // Work out the acceleration from the force.
+    glm::vec3 resultingAcceleration = physicsVelocity.acceleration;
+    resultingAcceleration += physicsVelocity.totalForce * physicsMass.inverseMass;
+
+    // Update linear velocity from the acceleration.
+    physicsVelocity.velocity += resultingAcceleration * duration;
+
+    // Impose drag.
```
luishfonseca, line 28: I'm not a fan of calling this drag, makes damping look optional.
Damping is a necessity since semi implicit euler will increase to total energy in the system overtime, violating conservation of energy. Normally it is a parameter of the simulation that is adjusted globally

engine/include/cubos/engine/physics/physics_velocity.hpp
```diff
@@ -0,0 +1,24 @@
...
+    /// @brief Component which defines parameters of a particle.
+    /// @note Should be used with @ref PhysicsMass and @ref Position.
+    /// @ingroup physics-plugin
+    struct [[cubos::component("cubos/physics_velocity", VecStorage)]] PhysicsVelocity
+    {
+        glm::vec3 velocity;
+        glm::vec3 acceleration;
+        glm::vec3 totalForce;
+        glm::vec3 gravity;
+        float damping;
```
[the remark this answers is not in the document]
fallenatlas: I agree we don't need acceleration, at the moment I'm using that as gravity which is wrong. The gravity here is supposed to be an acceleration, and I was thinking of it as being specific for this object, like you could set a different gravity for each object. We could have maybe a global value if this isn't set but that was the idea. So if we go with this, how would we have a constant specific to an entity without it being on one of it's components? Or should we just have a global gravity force?

engine/src/cubos/engine/physics/plugin.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <glm/glm.hpp>
+
+using cubos::core::ecs::Query;
+using cubos::core::ecs::Read;
+using cubos::core::ecs::Write;
+
+using namespace cubos::engine;
+
+// Semi-implicit euler integration
```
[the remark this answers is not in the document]
fallenatlas: The plan is to do that, yeah. I implemented this just as a base before attempting that one, bc I haven't looked much into the solver part yet.

engine/include/cubos/engine/physics/physics_velocity.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+        bool added;
```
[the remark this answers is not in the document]
fallenatlas: This is meant to be temporary. This is to make the systems that apply forces only apply that force once per physics update. Only works if the frame rate is higher than the physics update, which it currently is. I don't know, which way we should go on that yet. Do you have any ideas on it?

engine/src/cubos/engine/physics/plugin.cpp
fallenatlas: I agree, I'll make the damping global. We should keep drag for the component right? I think for most situations that covers what people will need for drag on an object.

luishfonseca: I don't think drag should be built into the core of our physics engine, users may set a global drag "feel" by adjusting the damping. If they then require different drags (e.g. water) a system to apply that force can be trivially written by them.

engine/include/cubos/engine/physics/physics_velocity.hpp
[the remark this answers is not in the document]
RiscadoA: If we wanted to have each entity with a different gravity value, I would prefer it to be on a different component. That way we could even have, for example, a global gravity resource `Gravity`, which you could override by adding it to a specific entity and setting it to another value. What do you think?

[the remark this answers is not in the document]
fallenatlas: I like that, it's better than what I originally thought. In the meantime, I also remembered we could just have a component which holds a modifier that we always use when applying the gravity. It could also have a bool to deactivate gravity, even though this can be done by setting the value to 0, but it would be more of a user friendly thing, for the editor too. What do you think?

engine/samples/physics/main.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+static void pushCarSystem(Query<Write<PhysicsVelocity>> query, Write<MaxTime> time, Read<DeltaTime> deltaTime) {
+    for (auto [entity, velocity] : query) {
+        if (time->current < time->max) {
+            if (time->current == 0.0F)
+            {
+                velocity->impulse += glm::vec3(0.0F, 5000.0F, 0.0F);
+            }
+            velocity->force = velocity->force + glm::vec3(0.0F, 0.0F, -5000.0F);
+            time->current = time->current + deltaTime->value;
```
RiscadoA, line 83:
```suggestion
            velocity->force += glm::vec3(0.0F, 0.0F, -5000.0F);
            time->current += deltaTime->value;
```

engine/include/cubos/engine/physics/resources/accumulator.hpp
```diff
@@ -0,0 +1,15 @@
...
+#pragma once
+
+namespace cubos::engine
+{
+    /// @brief Resource which holds the progress to the next integration.
+    /// @ingroup physics-plugin
+    struct Accumulator
+    {
+        float value = 0.0F;
+    };
```
RiscadoA, line 14: This should be internal, and probably renamed to `PhysicsAccumulator`.

engine/include/cubos/engine/physics/plugins/gravity.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <cubos/engine/physics/components/physics_velocity.hpp>
+#include <cubos/engine/physics/components/physics_mass.hpp>
+
+namespace cubos::engine
+{
+    /// @defgroup physics-plugin Physics
+    /// @ingroup engine
+    /// @brief Adds gravity to particles.
+    /// ## Resources
+    /// - @ref Gravity - holds the global value of gravity.
```
RiscadoA, line 20:
```suggestion
    /// @defgroup physics-gravity-plugin Physics
    /// @ingroup physics-plugin
    /// @brief Adds gravity to particles.
    ///
    /// ## Resources
    /// - @ref Gravity - holds the global value of gravity.
```

engine/include/cubos/engine/physics/plugins/gravity.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+    /// @brief Plugin entry function.
+    /// @param cubos @b CUBOS. main class
+    /// @ingroup physics-plugin
```
RiscadoA, line 24:
```suggestion
    /// @ingroup physics-gravity-plugin
```

engine/include/cubos/engine/physics/plugins/gravity.hpp
```diff
@@ -0,0 +1,26 @@
+/// @file
+/// @brief Gravity plugin, resource @ref cubos::engine::Gravity.
+/// @ingroup physics-plugin
```
RiscadoA, line 3:
```suggestion
/// @ingroup physics-gravity-plugin
```

engine/include/cubos/engine/physics/plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+    ///
+    /// ## Internal Simulation Tags
+    /// - `cubos.physics.simulation.prepare`
+    /// - `cubos.physics.simulation.apply_impulses`
+    /// - `cubos.physics.simulation.substeps.integrate`
+    /// - `cubos.physics.simulation.substeps.correct_position`
+    /// - `cubos.physics.simulation.substeps.update_velocity`
+    /// - `cubos.physics.simulation.clear_forces`
+    /// - `cubos.physics.simulation.decrease_accumulator`
+    ///
```
RiscadoA, line 56:
```suggestion
```

engine/include/cubos/engine/physics/plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+    /// ## User Tags
```
RiscadoA, line 57:
```suggestion
    /// ## Tags
```

engine/include/cubos/engine/physics/plugin.hpp
```diff
@@ -0,0 +1,64 @@
+/// @dir
+/// @brief @ref physics-plugin plugin directory.
+
+/// @file
+/// @brief Plugin entry point, resources @ref cubos::engine::FixedDeltaTime and @ref cubos::engine::Accumulator and components @ref
+/// cubos::engine::PhysicsVelocity and @ref cubos::engine::PhysicsMass.
```
RiscadoA, line 6:
```suggestion
/// @brief Plugin entry point.
```

RiscadoA (changes requested):

engine/src/cubos/engine/physics/gravity.cpp
```diff
@@ -0,0 +1,31 @@
...
+static void applyGravitySystem(Query<Write<PhysicsVelocity>, Read<PhysicsMass>> query, Read<Gravity> gravity)
+{
+    for (auto [entity, velocity, mass] : query)
+    {
+        if (mass->inverseMass <= 0.0f) return;
+
+        // Apply gravity force
+        glm::vec3 gravitationForce = mass->mass * gravity->value;
+
+        velocity->force = velocity->force + gravitationForce;
```
fallenatlas, line 22:
```suggestion
        velocity->force += gravitationForce;
```

RiscadoA (approved):

PR #716, opened 2023-10-13
 fix(engine): add assignment of 'reflectedId' in AnyAsset::operator Asset<T>()

roby2014:
Fixed crashing when loading assets.


RiscadoA (approved):
Ty for the quick fix! Ship it🚢

PR #718, opened 2023-10-14
Toggleable debug camera

roby2014:
# Description

Added debug camera plugin

## Checklist

- [X] Self-review changes.
- [X] Ensure test coverage.
- [X] Write new samples.


DiogoMendonc-a (approved):

DiogoMendonc-a:
If we're not adding it in this pr, we should create an issue for adding controls to this camera

RiscadoA (commented):
Shouldn't this plugin provide free camera movement controls too? Im not sure on how we would do that though. Should it be done directly here, or should we add a 'FreeCameraController' component in another plugin, which this one would depend on? Im leaning more towards a separate plugin, might be useful outside this plugin

RiscadoA:
Okay we both wrote the same thing at the same time @DiogoMendonc-a xD

DiogoMendonc-a:
Great mind think alike 😌

RiscadoA (approved):

PR #727, opened 2023-10-17
Add flag to control asset versioning when getting write access

roby2014:
# Description

Added a flag to  `Assets::write` in order to control whether the version of assets should be incremented when getting write access.

## Checklist

- [X] Self-review changes.
- [ ] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/assets/assets.hpp
+++ b/engine/include/cubos/engine/assets/assets.hpp
@@ -152,14 +152,15 @@ namespace cubos::engine
         ///
         /// @tparam T Type of the asset data.
         /// @param handle Handle to get the asset data for.
+        /// @param incVersion Whether to increase the asset's version.
         /// @return Reference to the asset data.
         template <typename T>
-        inline AssetWrite<T> write(Asset<T> handle)
+        inline AssetWrite<T> write(Asset<T> handle, bool incVersion = true)
         {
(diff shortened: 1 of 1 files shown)


RiscadoA:
This PR should also add a method used to bump the version manually, but I'm not sure how we would do this. Maybe we should add a third locking function, instead of adding an argument to the existing one.

E.g.:
```cpp
auto assetWrite = assets->potentialWrite(asset);
/* ... potentially do changes, just like before */
if (modified)
{
    // consumes the assetWrite lock, unlocking the asset, but before that, increments the asset's version.
    assets->confirmWrite(std::move(assetWrite));
}
```

The thing is, currently `AssetWrite` is just an alias for `core::memory::WriteGuard<T, std::unique_lock<std::shared_mutex>>`, and for this to work, it would also need to store the UUID of the asset being accessed. Otherwise, we would need to add an asset handle argument to the `confirmWrite` method, which is duplicating information and thus error prone.

Maybe instead of making it an alias, we could make it a child class of the `WriteGuard` and then just store the extra UUID field, which could then be used in the `confirmWrite` implementation.

RiscadoA (changes requested):
Forgot to request changes!

RiscadoA:
Closing due to inactivity

PR #728, opened 2023-10-17
Add `imguiPlugin` sample 

roby2014:
# Description

Added `imguiPlugin` sample

## Checklist

- [X] Self-review changes.


RiscadoA (changes requested):
Thanks for working on this! Looks good to me :) just some nitpicks + noticed you forgot to add this sample to the list on `docs/pages/3_examples/2_engine/main.md`. Its necessary, as without it the page won't be indexed properly.

image

engine/samples/imgui/main.cpp
```diff
@@ -0,0 +1,35 @@
...
+using namespace cubos::engine;
+
+/// [ImGui window example]
+static void exampleWindow()
+{
+    ImGui::Begin("Dear ImGui + CUBOS.");
+
+    ImGui::ShowDemoWindow();
+
+    ImGui::End();
```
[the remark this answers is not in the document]
roby2014: It's not a problem I think but I can put it outside yeah

RiscadoA (changes requested):
We currently do not include the whole sample code at the end of the examples. Maybe we should, but I think that that should be consistent among all examples.

Some are too long though, so idk how to feel about it. Whats your opinion?

One more thing, if you want the entire file, you do not need the snippet, you can just `@include` it.

roby2014:
> We currently do not include the whole sample code at the end of the examples. Maybe we should, but I think that that should be consistent among all examples.
> 
> Some are too long though, so idk how to feel about it. Whats your opinion?
> 
> One more thing, if you want the entire file, you do not need the snippet, you can just `@include` it.

I always like when docs provide the full example. Also maybe some people just wanna c+p it away. So idk, maybe we can decide this on the next meeting?

engine/samples/imgui/page.md
```diff
@@ -0,0 +1,27 @@
+# ImGui {#examples-engine-imgui}
+
+@brief Using the @ref imgui-plugin plugin.
+
+The ImGui plugin allows you to integrate the [Dear ImGui](https://github.com/ocornut/imgui) library with CUBOS. engine, making it quick, straightforward, and effortless to create user interfaces and interactive debugging tools within your applications.
```
DiogoMendonc-a, line 5: very minor nitpick, but I think it sounds better if we write it either `with Cubos., making it ...` or `with the CUBOS. engine, making it...`

engine/samples/imgui/page.md
```diff
@@ -0,0 +1,27 @@
...
+
+Once the ImGui plugin is added, you can create systems to display ImGui windows and widgets. Here's an example of how to create an ImGui window.
+
+@snippet imgui/main.cpp ImGui window example
+
+Ensure that you add your system with the `cubos.imgui` tag; otherwise, the ImGui elements from that system won't be be visible.
+
+@snippet imgui/main.cpp Adding the system
+
+Pretty simple right? You're now equipped to craft and utilize ImGui's functions to design your cool user interface.
```
DiogoMendonc-a, line 25: It'd be nice if we had a link here to where people can check how to use imgui stuff. Someone coming to this sample is probably looking for information that is more in depth than just the process of spawning a window.

DiogoMendonc-a (changes requested):

engine/samples/imgui/page.md
roby2014: Well for that we could just add ImGui's documentation link or something.

DiogoMendonc-a: Yeah, that was what I meant, sorry if I wasn't clear, it wasn't telling you to document all of imgui yourself

DiogoMendonc-a (approved):

RiscadoA (approved):

PR #756, opened 2023-10-30
feat(ecs): add negative tags

RiscadoA:
# Description

Adds the `.noTag` and `.noStartupTag` methods to `Cubos`, analogs to `.tag` and `.startupTag` but which can be used to apply configuration to all systems except the ones with the given tag. This was implemented by simply storing the negative tag settings on the same tag name prefixed with '!'. Then, when we apply the tag settings to the systems, we first add the tags `!x` to the system if they're not tagged with `x`.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] ~~Write new samples.~~ (I don't think this deserves a sample of its own)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Log the system order on the dispatcher (#414, **@RiscadoA**).
+- Negative tags to the dispatcher (#390, **@RiscadoA**).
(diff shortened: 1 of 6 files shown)


fallenatlas (approved):

RiscadoA:
Thanks for reviewing this @fallenatlas! Honestly I'm not sure if we should merge this as it is right now.
Since @joaomanita is working on the dispatcher right now and we will have to make big changes to its structure anyway, it probably makes more sense to rethink this later. Marking this as stale and leaving it here, as it might still be useful.

PR #762, opened 2023-10-31
Add arrow gizmo

DiogoMendonc-a:


RiscadoA:
Btw you've been using `engine` as a commit scope, but you can be more specific. In this case, `gizmo` would be more appropriate.

engine/src/cubos/engine/gizmos/gizmos.cpp
```diff
@@ -94,3 +94,11 @@ void Gizmos::drawCutCone(const std::string& id, glm::vec3 firstBaseCenter, float
...
          space);
 }
+
+void Gizmos::drawArrow(const std::string& id, glm::vec3 origin, glm::vec3 direction, float girth, float width,
+                       float ratio, float lifespan, Space space)
+{
+    auto p = direction * ratio;
+    drawCutCone(id, origin, girth, origin + p, girth, lifespan, space);
+    drawCutCone(id, origin + p, width, origin + direction, 0.0F, lifespan, space);
+}
```
RiscadoA, line 104: Add missing newline at EOF

.github/CODEOWNERS
```diff
@@ -9,4 +9,6 @@
 **/engine/input/ @luishfonseca
 **/engine/physics/ @luishfonseca @fallenatlas
```
RiscadoA, line 11:
```suggestion
```

PR #775, opened 2023-11-04
Remove old serialization code from `SceneBridge`

RiscadoA:
# Description

Makes the ECS use the actual type name to identify components instead of the attribute name - still haven't removed it from quadrados, but the whole attribute will soon go away anyway. 
Changes the `SceneBridge` to use the new `JSONDeserializer` instead of the old one.

## Checklist

- [x] Self-review changes.
- [x] Update samples.

### DIFF
--- a/engine/include/cubos/engine/scene/bridge.hpp
+++ b/engine/include/cubos/engine/scene/bridge.hpp
@@ -6,7 +6,7 @@
 
 #include <cubos/core/reflection/reflect.hpp>
 
-#include <cubos/engine/assets/bridge.hpp>
+#include <cubos/engine/assets/bridges/file.hpp>
 #include <cubos/engine/scene/scene.hpp>
 
 namespace cubos::engine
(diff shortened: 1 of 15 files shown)


engine/include/cubos/engine/scene/bridge.hpp
```diff
@@ -25,35 +25,34 @@ namespace cubos::engine
...
     /// @ingroup scene-plugin
-    class SceneBridge : public AssetBridge
+    class SceneBridge : public FileBridge
     {
     public:
         /// @brief Constructs a bridge.
         ///
         SceneBridge()
-            : AssetBridge(typeid(Scene))
+            : FileBridge(typeid(Scene))
```
roby2014, line 50: `FileBridge` from #795 kinda changed how things work here, maybe to avoid conflicts, rebase and use the new version?

roby2014 (changes requested):

engine/include/cubos/engine/scene/bridge.hpp
roby2014: Wait and #795 depends on this PR :V

RiscadoA: Feel free to merge the other one, it uses just a single commits from this PR, its okay :relieved: I'll rebase it on main later

roby2014 (approved):

PR #777, opened 2023-11-05
Implement gizmo selection

DiogoMendonc-a:


engine/src/cubos/engine/gizmos/renderer.cpp
```diff
@@ -1,12 +1,35 @@
...
+#include <string>
 
 using cubos::engine::GizmosRenderer;
 
 using namespace cubos::core::gl;
 
-void GizmosRenderer::init(RenderDevice* currentRenderDevice)
+void GizmosRenderer::initIdTexture(glm::ivec2 size)
+{
+    cubos::core::gl::Texture2DDesc texDesc;
```
RiscadoA, line 10: Unnecessary namespace, and same below. You already have 'using namespace cubos::core::gl' at the top of the file.

engine/include/cubos/engine/gizmos/gizmos.hpp
```diff
@@ -90,27 +91,47 @@ namespace cubos::engine
...
 
+        /// @brief Gets whether the left mouse button was pressed over a gizmo.
+        /// @param id Identifier of the gizmo.
+        /// @return Whether the gizmo was pressed.
+        bool pressed(const std::string& id) const;
+
+        /// @brief Gets whether the mouse button is over a gizmo.
+        /// @param id Identifier of the gizmo.
+        /// @return Whether the mouse button is over a gizmo.
+        bool hover(const std::string& id) const;
```
RiscadoA, line 102:
```suggestion
        bool hovered(const std::string& id) const;
```

engine/include/cubos/engine/gizmos/gizmos.hpp
```diff
@@ -90,27 +91,47 @@ namespace cubos::engine
         void drawWireBox(const std::string& id, glm::vec3 corner, glm::vec3 oppositeCorner, float lifespan = 0.0F,
                          Space space = Space::World);
 
+        /// @brief Gets whether the left mouse button was pressed over a gizmo.
+        /// @param id Identifier of the gizmo.
+        /// @return Whether the gizmo was pressed.
+        bool pressed(const std::string& id) const;
+
+        /// @brief Gets whether the mouse button is over a gizmo.
```
RiscadoA, line 99:
```suggestion
        /// @brief Checks whether the mouse button is over a gizmo. 
```

engine/include/cubos/engine/gizmos/gizmos.hpp
```diff
@@ -90,27 +91,47 @@ namespace cubos::engine
         void drawWireBox(const std::string& id, glm::vec3 corner, glm::vec3 oppositeCorner, float lifespan = 0.0F,
                          Space space = Space::World);
 
+        /// @brief Gets whether the left mouse button was pressed over a gizmo.
```
RiscadoA, line 94:
```suggestion
        /// @brief Checks whether the left mouse button was pressed over a gizmo.
```

engine/include/cubos/engine/gizmos/gizmos.hpp
```diff
@@ -90,27 +91,47 @@ namespace cubos::engine
...
+        /// @param id Identifier of the gizmo.
+        /// @return Whether the gizmo was pressed.
+        bool pressed(const std::string& id) const;
+
+        /// @brief Gets whether the mouse button is over a gizmo.
+        /// @param id Identifier of the gizmo.
+        /// @return Whether the mouse button is over a gizmo.
+        bool hover(const std::string& id) const;
+
+        void registerInputGizmo(uint32_t id, bool pressed);
```
RiscadoA, line 104:
```suggestion
        void handleInput(uint32_t hovered, bool pressed);
```

Also add missing docs, don't forget to specify that this is called automatically by a system registered by the plugin.

engine/include/cubos/engine/gizmos/gizmos.hpp
```diff
@@ -90,27 +91,47 @@ namespace cubos::engine
...
             /// @param mvp The view projection matrix to use for drawing the gizmo.
-            virtual void draw(GizmosRenderer& renderer, const glm::mat<4, 4, float, glm::packed_highp>& mvp) = 0;
+            virtual void draw(GizmosRenderer& renderer, DrawPhase phase,
+                              const glm::mat<4, 4, float, glm::packed_highp>& mvp) = 0;
 
             /// @brief Decreases the time the gizmo has left before it is destroyed.
             /// @param delta Seconds since the last frame.
             bool decreaseLifespan(float delta);
 
+            uint32_t id; ///< Gizmo identifier.
```
RiscadoA, line 130:
```suggestion
            const uint32_t id; ///< Gizmo identifier.
```

engine/src/cubos/engine/gizmos/gizmos.cpp
```diff
@@ -16,13 +16,29 @@ using cubos::engine::GizmosRenderer;
...
+}
+
+bool Gizmos::pressed(const std::string& id) const
+{
+    return mPressed && mHasher(id) == mIdInteraction && mIdInteraction != 0;
+}
+
+bool Gizmos::hover(const std::string& id) const
+{
+    return !mPressed && mHasher(id) == mIdInteraction && mIdInteraction != 0;
```
RiscadoA, line 39: I don't really like the combination of using a hash as ID and also having magic values such as 0. But its okay to leave this as is. When we generalize screen picking to all kinds of entities, we'll have to change the logic on how we identify stuff anyway.

engine/src/cubos/engine/gizmos/plugin.cpp
```diff
@@ -183,6 +203,9 @@ static void drawGizmosSystem(Write<Gizmos> gizmos, Write<GizmosRenderer> gizmosR
 {
     auto screenSize = (*window)->framebufferSize();
 
+    gizmosRenderer->renderDevice->setFramebuffer(gizmosRenderer->framebuffer);
```
RiscadoA, line 206: I would change the name of the `framebuffer` field to `idFramebuffer` or something like that. This way its a bit confusing

engine/src/cubos/engine/gizmos/plugin.cpp
```diff
@@ -193,9 +216,49 @@ static void drawGizmosSystem(Write<Gizmos> gizmos, Write<GizmosRenderer> gizmosR
...
+            }
+        }
+        else if (std::holds_alternative<ResizeEvent>(event))
+        {
+            gizmosRenderer->resizeTexture(std::get<ResizeEvent>(event).size);
+        }
+    }
+
+    auto* texBuffer = new uint16_t[(std::size_t)gizmosRenderer->textureSize.x *
+                                   (std::size_t)gizmosRenderer->textureSize.y * (std::size_t)2U];
```
RiscadoA, line 242:
```suggestion
                                   (std::size_t)gizmosRenderer->textureSize.y * 2U];
```

The cast shouldn't be necessary.

engine/src/cubos/engine/gizmos/plugin.cpp
```diff
@@ -193,9 +216,49 @@ static void drawGizmosSystem(Write<Gizmos> gizmos, Write<GizmosRenderer> gizmosR
...
+
+    gizmosRenderer->idTexture->data(texBuffer);
```
RiscadoA, line 244: OpenGL provides [glReadPixels](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glReadPixels.xhtml) which seems a much better fit to use here. I think that with that you wouldn't even need the integer masking down there, as you could directly extract the data in the format you desire. I think using it instead is a good idea - copying the whole screen texture every frame just to check a single pixel is a big cost to pay - and the engine is already slow as it is :cry:

engine/src/cubos/engine/gizmos/renderer.cpp
```diff
@@ -101,20 +176,20 @@ void GizmosRenderer::initCutConePrimitive()
...
     float verts[CutConeVertsPerBase * 6];
 
-    renderDevice->setShaderPipeline(pipeline);
+    renderDevice->setShaderPipeline(drawPipeline);
     cutConePrimitive.vb = renderDevice->createVertexBuffer(sizeof(verts), verts, Usage::Dynamic);
     cutConePrimitive.vaDesc.buffers[0] = cutConePrimitive.vb;
     cutConePrimitive.va = renderDevice->createVertexArray(cutConePrimitive.vaDesc);
 
-    unsigned int indices[12 * CutConeVertsPerBase - 12];
+    std::size_t indices[12 * CutConeVertsPerBase - 12];
```
RiscadoA, line 188: This is wrong, the buffer takes an `UInt`, not a `std::size_t`. This only works on your machine because you're compiling this for x86, which means they both have the same size.

RiscadoA (changes requested):
Thanks for working on this @DiogoMendonc-a!

engine/src/cubos/engine/gizmos/plugin.cpp
```diff
@@ -193,9 +216,48 @@ static void drawGizmosSystem(Write<Gizmos> gizmos, Write<GizmosRenderer> gizmosR
...
+                gizmosRenderer->mousePressed = std::get<MouseButtonEvent>(event).pressed;
+            }
+        }
+        else if (std::holds_alternative<ResizeEvent>(event))
+        {
+            gizmosRenderer->resizeTexture(std::get<ResizeEvent>(event).size);
+        }
+    }
+
+    auto* texBuffer = new unsigned short[gizmosRenderer->textureSize.x * gizmosRenderer->textureSize.y * 2];
```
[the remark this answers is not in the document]
DiogoMendonc-a: Cause i'm using 2 channels, red and green

engine/src/cubos/engine/gizmos/gizmos.cpp
DiogoMendonc-a: I don't like it either, but there's no reasonable way to store strings on a texture, and using integer identifiers could easily lead to even more collisions

RiscadoA: We can easily use integer identifies by storing a map of strings to integers on the CPU side, but yeah, that isn't necessary rn.

RiscadoA (approved):

RiscadoA:
Btw you wrote 'gizmos' but the right scope is 'gizmo', it should match the label

DiogoMendonc-a:
> Btw you wrote 'gizmos' but the right scope is 'gizmo', it should match the label

In that case, I think we should change the label. It's the gizmos plugin, the gizmos resource, it would make sense for it to be the gizmos label

PR #779, opened 2023-11-05
Implement `DataInspector`

roby2014:
# Description

- Implemented `DataInspector` resource, using the new reflection system, to replace the Package stuff
- Added demonstration to imgui sample
- Added support for `AnyValue` default value
- implement dictTrait iterator assign operator

## Checklist

- [X] Self-review changes.
- [X] Ensure test coverage.
- [X] Write new samples.


engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace cubos::engine
+{
+    /// @brief Resource which allows user to inspect or modify object's properties in the UI.
+    ///
+    /// @ingroup imgui-plugin
+    class DataInspector final
+    {
+        /// @brief Alias for @ref core::reflection::Type.
+        using Type = core::reflection::Type;
```
RiscadoA, line 17: I think its best to omit the alias here, its just some extra characters in the methods and it won't ever be used through the class elsewhere.

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+        /**
+         * @brief Display a reflective object, with "read/write" permissions.
+         *
+         * @note This should always be called inside a `ImGui::BeginTable(..., 2)` and `ImGui::EndTable()`.
+         * @param name The name of the object.
+         * @param type The reflection type of the object.
+         * @param value A pointer to the object's value.
+         * @return True if object was modified.
+         */
```
RiscadoA, line 38: Use triple slash docs style (same in other funcs)

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,42 @@
...
+        bool edit(const std::string& name, const Type& type, void* value);
+    };
+
+} // namespace cubos::engine
```
RiscadoA, line 42: Add missing new line at EOF

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,42 @@
...
+/// @brief Resource @ref cubos::engine::DataInspector.
+/// @ingroup imgui-plugin
+
+#pragma once
+
+#include <cubos/core/reflection/type.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Resource which allows user to inspect or modify object's properties in the UI.
```
RiscadoA, line 11:
```suggestion
    /// @brief Resource which allows the user to inspect or modify any reflectable value in the UI.
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,42 @@
...
+    ///
```
RiscadoA, line 12:
```suggestion
    /// @see Take a look at the @ref examples-engine-imgui example for a demonstration of this resource.
```

engine/samples/data_inspector/main.cpp
RiscadoA, line 1: I would make this part of the ImGui sample

engine/src/cubos/engine/imgui/data_inspector.cpp
RiscadoA, line 1: I think you should, as we discussed previously, replace these static show/edit functions for static inspect functions, which take a non-const pointer to the value and a bool indicating whether the operation is readOnly.

RiscadoA (changes requested):
Thanks for working on this @roby2014! One step closer to getting rid of the old namespace :eyes:

roby2014:
Looks good for a review. Next step would be to remove the old serialization method and use the new resource on the Entity inspector plugin & etc... Can I do that in this PR as well?

RiscadoA:
Ty! Will review friday. I think its best to leave it to another PR - you can rebase it on this one

core/include/cubos/core/memory/any_value.hpp
```diff
@@ -18,12 +18,20 @@ namespace cubos::core::memory
...
 
+        /// @brief Move assignment.
+        /// @param other Value.
+        AnyValue& operator=(AnyValue&& other) noexcept;
+
         /// @brief Default constructs a value with the given type.
-        /// @note @p type must have @ref reflection::ConstructibleTrait and be default
+        /// @note @p type must have @ref reflection::ConstructibleTrait and be default.
+        /// @note mType member must not be null.
         /// constructible.
```
RiscadoA, line 35:
```suggestion
        /// @note @p type must have @ref reflection::ConstructibleTrait and be default
        /// constructible.
```

core/include/cubos/core/memory/any_value.hpp
```diff
@@ -44,6 +52,7 @@ namespace cubos::core::memory
         static AnyValue moveConstruct(const reflection::Type& type, void* value) noexcept;
 
         /// @brief Get the type of the elements stored in the vector.
+        /// @note mType member must not be null.
```
RiscadoA, line 55:
```suggestion
        /// @note @ref valid() must return true. 
```

And add a method `bool valid() const` which returns whether a value is held (`mType != nullptr`).

core/src/cubos/core/memory/any_value.cpp
```diff
@@ -63,7 +80,8 @@ AnyValue AnyValue::moveConstruct(const Type& type, void* value) noexcept
 
 const Type& AnyValue::type() const
 {
-    return mType;
+    CUBOS_ASSERT(mType != nullptr, "Type must not be null");
```
RiscadoA, line 83:
```suggestion
    CUBOS_ASSERT(this->valid(), "This must hold a value");
```

core/src/cubos/core/memory/any_value.cpp
```diff
@@ -63,7 +80,8 @@ AnyValue AnyValue::moveConstruct(const Type& type, void* value) noexcept
...
+    return *mType;
 }
 
 void* AnyValue::get()
```
RiscadoA, line 87: Change the get methods to also assert `this->valid`

core/tests/memory/any_value.cpp
```diff
@@ -35,6 +35,14 @@ TEST_CASE("memory::AnyValue")
             CHECK(any.type().is<int>());
             CHECK(*static_cast<int*>(any.get()) == value);
         }
+
+        SUBCASE("move assignment")
+        {
+            int value = 1337;
+            AnyValue any = AnyValue::moveConstruct(reflect<int>(), &value);
+            CHECK(any.type().is<int>());
+            CHECK(*static_cast<int*>(any.get()) == value);
```
RiscadoA, line 44: I would change this to
```suggestion
            int value = 1337;
            AnyValue any{};
            REQUIRE_FALSE(any.valid());
            any = AnyValue::moveConstruct(reflect<int>(), &value);
            REQUIRE(any.valid());
            CHECK(any.type().is<int>());
            CHECK(*static_cast<int*>(any.get()) == value);
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,66 @@
...
+    public:
+        /// @brief Inspect a reflective object.
+        ///
+        /// @note This should always be called inside a `ImGui::BeginTable(..., 2)` and `ImGui::EndTable()`.
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param value A pointer to the object's value.
+        /// @param readOnly A flag that indicates wether this field can be modified.
+        /// @return True if object was modified.
+        bool inspect(const std::string& name, const core::reflection::Type& type, void* value, bool readOnly = true);
```
RiscadoA, line 26: I would keep inspect something internal and here have only `show` and `edit`, show receiving `const void*`. This way the user doesn't have to use const_cast.

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    private:
+        std::string mTextBuffer;
+        bool mKeyExists;
+        cubos::core::memory::AnyValue mKey;
+        cubos::core::memory::AnyValue mValue;
```
RiscadoA, line 32: Usually we put the private fields below the private methods

engine/samples/imgui/img2.png
RiscadoA, line 1: Isn't the map missing 'remove' buttons? Also, are we keeping labels on both columns then?

engine/samples/imgui/main.cpp
```diff
@@ -1,12 +1,92 @@
...
+
+#include <cubos/engine/imgui/data_inspector.hpp>
 #include <cubos/engine/imgui/plugin.hpp>
+
 /// [Including plugin headers]
 
+using cubos::core::ecs::Read;
+using cubos::core::ecs::Write;
+using cubos::core::reflection::reflect;
+using cubos::engine::DataInspector;
```
RiscadoA, line 23:
```suggestion
using cubos::core::reflection::reflect;

using cubos::engine::DataInspector;
```

engine/samples/imgui/main.cpp
```diff
@@ -1,12 +1,92 @@
...
+    CUBOS_REFLECT;
+    std::string name;
+    int32_t age;
+    float weight;
+    bool dead;
+};
+
+CUBOS_REFLECT_IMPL(Person)
+{
+    return cubos::core::reflection::Type::create("Person").with(cubos::core::reflection::FieldsTrait()
```
RiscadoA, line 38:
```suggestion
    return Type::create("Person").with(FieldsTrait()
```

Add using statements for this

engine/samples/imgui/main.cpp
```diff
@@ -1,12 +1,92 @@
...
+{
+    int integer;
+    Person person;
+    std::vector<int32_t> vec;
+    std::map<int32_t, int32_t> map;
+};
+/// [Creating a dummy resource]
+
+/// [Filling the dummy resource 1]
+static void fillDumyResourceSystem(Write<DummyResource> data)
```
RiscadoA, line 55:
```suggestion
static void fillDummyResourceSystem(Write<DummyResource> data)
```

engine/include/cubos/engine/imgui/data_inspector.hpp
RiscadoA: Would also be cool to add template methods with signatures `show(const std::string& name, const T& value);` which call reflect for us.

engine/samples/imgui/main.cpp
```diff
@@ -1,12 +1,92 @@
...
+    static const auto& vecType = reflect<std::vector<int32_t>>();
+    static const auto& mapType = reflect<std::map<int32_t, int32_t>>();
+
+    ImGui::Begin("Data Inspector");
+    ImGui::BeginTable("id1", 2);
+
+    inspector->inspect("data->integer", integerType, (void*)&data->integer);
+    inspector->inspect("data->person", personType, (void*)&data->person, false);
+    inspector->inspect("data->vec", vecType, (void*)&data->vec, false);
+    inspector->inspect("data->map", mapType, (void*)&data->map, false);
```
RiscadoA, line 83: If you add the templated show/edit methods, you no longer need those static vars above

engine/samples/imgui/main.cpp
```diff
@@ -30,6 +110,14 @@ int main()
     cubos.system(exampleWindowSystem).tagged("cubos.imgui");
     /// [Adding the system]
 
+    /// [Filling the dummy resource 2]
+    cubos.addResource<DummyResource>();
```
RiscadoA, line 114: You could initialize the resource here btw, and thus remove the need for a startup system:
```suggestion
    cubos.addResource<DummyResource>(DummyResource{
        .integer = 1337,
        /*...*/
    });
```

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+        {
+            showStructured(name, type, [&]() {
+                for (auto [field, fieldVal] : view)
+                {
+                    changed |= inspect(field->name(), field->type(), fieldVal, readOnly);
+                }
+            });
+        }
+    }
+    else
```
RiscadoA, line 176: I think you need to handle `StringConversionTrait` here: for example, `RenderableGrid` components contain a `Asset<Grid>` field which only provides that trait.

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+        {
+            auto* p = static_cast<std::string*>(value);
+            *p = mTextBuffer;
+            modified = true;
+        }
+    }
+    ImGui::PopID();
+
+    return modified;
+}
```
RiscadoA, line 309: Missing newline at EOF

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ImGui::TableSetColumnIndex(0);
+    ImGui::AlignTextToFramePadding();
+    ImGui::TreeNodeEx(name.c_str(),
+                      ImGuiTreeNodeFlags_Leaf | ImGuiTreeNodeFlags_NoTreePushOnOpen | ImGuiTreeNodeFlags_Bullet);
+    if (ImGui::IsItemHovered(ImGuiHoveredFlags_AllowWhenDisabled))
+    {
+        ImGui::SetTooltip("%s", type.name().c_str());
+    }
+
+    auto str = getPrimitiveValue(value, type);
```
RiscadoA, line 276: Wait what is this being used for?

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+        ImGui::SetTooltip("%s", type.name().c_str());
+    }
+
+    std::string str = getPrimitiveValue(value, type);
+    ImGui::TableSetColumnIndex(1);
+    ImGui::TextUnformatted(str.c_str());
+}
+
+void DataInspector::showStructured(const std::string& name, const Type& type, auto fn)
```
RiscadoA, line 239: Maybe rename this method to `withStructure` or something since it is both used 'edit' and 'show' methods.

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+    RETURN_FORMAT_STRING_IF_TYPE(int32_t)
+    RETURN_FORMAT_STRING_IF_TYPE(uint8_t)
+    RETURN_FORMAT_STRING_IF_TYPE(uint16_t)
+    RETURN_FORMAT_STRING_IF_TYPE(uint32_t)
+    RETURN_FORMAT_STRING_IF_TYPE(uint64_t)
+    RETURN_FORMAT_STRING_IF_TYPE(float)
+    RETURN_FORMAT_STRING_IF_TYPE(double)
+    if (type.is<bool>())
+    {
+        return *(bool*)(value) ? "true" : "false";
```
RiscadoA, line 211:
```suggestion
        return *static_cast<bool*>(value) ? "true" : "false";
```

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    return changed;
+}
+
+/// @brief Returns a string represeting primitive value, depending on it's [TYPE].
+#define RETURN_FORMAT_STRING_IF_TYPE(TYPE)                                                                             \
+    if (type.is<TYPE>())                                                                                               \
+    {                                                                                                                  \
+        return std::to_string(*(TYPE*)(value));                                                                        \
```
RiscadoA, line 195:
```suggestion
        return std::to_string(*static_cast<TYPE*>(value));                                                                        \
```

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+        return *(bool*)(value) ? "true" : "false";
+    }
+    if (type.is<std::string>())
+    {
+        mTextBuffer = *static_cast<const std::string*>(value);
+        return mTextBuffer;
+    }
+
+    CUBOS_ERROR("Could not get primitive object as formatted string (unknown primitive).");
```
RiscadoA, line 219:
```suggestion
    CUBOS_ERROR("Expected value with primitive type, found unsupported type '{}'", type.name());
```

engine/src/cubos/engine/imgui/data_inspector.cpp
```diff
@@ -0,0 +1,309 @@
...
+        if (ImGui::InputText(name.c_str(), &mTextBuffer, ImGuiInputTextFlags_EnterReturnsTrue))
+        {
+            auto* p = static_cast<std::string*>(value);
+            *p = mTextBuffer;
+            modified = true;
+        }
+    }
+    ImGui::PopID();
+
+    return modified;
```
RiscadoA, line 308: You should give an error and show ??? or something if the type is unsupported here, right now you just return without doing anything

engine/src/cubos/engine/imgui/plugin.cpp
```diff
@@ -41,6 +42,8 @@ void cubos::engine::imguiPlugin(Cubos& cubos)
     cubos.addPlugin(windowPlugin);
     cubos.addPlugin(rendererPlugin);
 
+    cubos.addResource<DataInspector>();
```
RiscadoA, line 45: You should also add this resource to the imgui plugin docs on the `## Resources` section (see other plugins)

RiscadoA (changes requested):
Thanks for working on this @roby2014! Good work :) One of the next steps we should worry about concerning this section of the code is allowing users of data inspector to register custom behavior for their types. That might require some changes later on, but it will be mostly restructuring methods and most of the code you wrote will remain the same.

engine/samples/imgui/img2.png
roby2014: Yeah my bad. 
Labels cant be removed, since they work like an "id" for the input scalars etc.. (we can removeit but then we would have to work with pushid pop id every input scalar, what do you prefer?

RiscadoA: Hmm. I prefer `PushId` `Pop`, although it is a nuisance code wise, having the labels repeated everywhere is triggering my OCD

roby2014:
> Thanks for working on this @roby2014! Good work :) One of the next steps we should worry about concerning this section of the code is allowing users of data inspector to register custom behavior for their types. That might require some changes later on, but it will be mostly restructuring methods and most of the code you wrote will remain the same.

Do you want to handle that custom type behaviour also in this PR?

RiscadoA:
> > Thanks for working on this @roby2014! Good work :) One of the next steps we should worry about concerning this section of the code is allowing users of data inspector to register custom behavior for their types. That might require some changes later on, but it will be mostly restructuring methods and most of the code you wrote will remain the same.
> 
> Do you want to handle that custom type behaviour also in this PR?

Nope, lets leave that for another PR. This one is already really large

roby2014:
> > > Thanks for working on this @roby2014! Good work :) One of the next steps we should worry about concerning this section of the code is allowing users of data inspector to register custom behavior for their types. That might require some changes later on, but it will be mostly restructuring methods and most of the code you wrote will remain the same.
> > 
> > 
> > Do you want to handle that custom type behaviour also in this PR?
> 
> Nope, lets leave that for another PR. This one is already really large

Alright, opened a tracking issue: #810

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    private:
+        /// @brief Inspect a reflective object.
+        ///
+        /// @note This should always be called inside a `ImGui::BeginTable(..., 2)` and `ImGui::EndTable()`.
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param value A pointer to the object's value.
+        /// @param readOnly A flag that indicates wether this field can be modified.
+        /// @return True if object was modified.
```
RiscadoA, line 68:
```suggestion
        /// @brief Displays a reflectable value on the UI, optionally allowing modification.
        /// @param name Value name.
        /// @param type Value type.
        /// @param value Pointer to value.
        /// @param readOnly Whether the value can be modified.
        /// @return Whether the value was modified.
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,108 @@
...
+        /// @param type The reflection type of the object.
+        /// @param value A const pointer to the object's value.
+        void showPrimitiveObject(const std::string& name, const core::reflection::Type& type, const void* value);
+
+        /// @brief Display a structured object, inside a `ImGui::TreeNode`.
+        ///
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param fn Lambda that should display object's structure and other ImGui percs by itself when the node is
+        /// open.
```
RiscadoA, line 99:
```suggestion
        /// @brief Displays a structured value inside a `ImGui::TreeNode`.
        /// @param name Value name.
        /// @param type Value type.
        /// @param fn Functor called when the node is open.
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,108 @@
...
+        /// @param value A const pointer to the object's value.
+        /// @return String that represents primitive value.
+        std::string getPrimitiveValue(const void* value, const core::reflection::Type& type);
+
+        /// @brief Display a primitive, or "???" if unknown type.
+        ///
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param value A const pointer to the object's value.
+        void showPrimitiveObject(const std::string& name, const core::reflection::Type& type, const void* value);
```
RiscadoA, line 92:
```suggestion
        /// @brief Displays a value of a known type, or "???" if its unknown.
        /// @param name Value name.
        /// @param type Value type.
        /// @param value Pointer to value.
        void showKnown(const std::string& name, const core::reflection::Type& type, const void* value);
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,108 @@
...
+        /// @return True if value was succesfully modified.
+        bool editPrimitive(const std::string& name, const core::reflection::Type& type, void* value);
+
+        /// @brief Tries to get a representative string of primitive value.
+        ///
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param value A const pointer to the object's value.
+        /// @return String that represents primitive value.
+        std::string getPrimitiveValue(const void* value, const core::reflection::Type& type);
```
RiscadoA, line 85:
```suggestion
        /// @brief Gets a representative string for a value of a known type, or "???" if its unknown.
        ///
        /// @param name The name of the object.
        /// @param type The reflection type of the object.
        /// @param value A const pointer to the object's value.
        /// @return String that represents primitive value.
        std::string stringKnown(const void* value, const core::reflection::Type& type);
```

engine/include/cubos/engine/imgui/data_inspector.hpp
```diff
@@ -0,0 +1,108 @@
...
+        /// @return True if object was modified.
+        bool inspect(const std::string& name, const core::reflection::Type& type, void* value, bool readOnly = true);
+
+        /// @brief Displays input scalars, gets user input and modifies value.
+        ///
+        /// @param name The name of the object.
+        /// @param type The reflection type of the object.
+        /// @param value A pointer to the object's value.
+        /// @return True if value was succesfully modified.
+        bool editPrimitive(const std::string& name, const core::reflection::Type& type, void* value);
```
RiscadoA, line 77:
```suggestion
        /// @brief Displays a value of a known type, or "???" if its unknown. Allows modification.
        /// @param name Value name.
        /// @param type Value type.
        /// @param value Pointer to value.
        /// @return Whether the value was modified.
        bool editKnown(const std::string& name, const core::reflection::Type& type, void* value);
```

roby2014:
todo: dictionary iterators tests

PR #783, opened 2023-11-07
feat(engine): add toolbox plugin

teres4:
# Description

Add toolbox plugin/resource, modified existing tools to now use said plugin

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// ## Resources
+    /// - @ref Toolbox 
+
+    /// @brief Resource which manages other tools windows.
+    class Toolbox final
+    {
+        public:
+
+        std::map<std::string, bool> toolsMap;
```
roby2014, line 33: I think we don't care about order, and `std::unordered_map` is faster(?) (or maybe does not matter at all)

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
```diff
@@ -0,0 +1,47 @@
...
+using cubos::engine::Cubos;
+using cubos::core::ecs::Write;
+
+using namespace tesseratos;
+
+static void showToolbox(Write<Toolbox> toolbox)
+{
+    ImGui::Begin("Toolbox");
+
+    for(auto a : toolbox->toolsMap)
```
roby2014, line 22: Kinda confused here. You can make it more readable by using modern C++ stuff: `for (auto& [key, val] : map)`

roby2014 (changes requested):

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,55 @@
+
+
```
roby2014, line 2: 2 empty lines

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,55 @@
...
+        bool isOpen(std::string toolname)
+        {
+            if (toolsMap.contains(toolname))
+            {
+                return toolsMap[toolname];
+            }
+            toolsMap[toolname] = false;
+            return false;
+        }
```
roby2014, line 38: I think you can just do `return toolsMap[toolname];` since `operator[]` will create a record if it does not exist (https://en.cppreference.com/w/cpp/container/unordered_map/operator_at).

roby2014 (changes requested):

roby2014:
After commiting, make sure to rebase history and make it all into 1 commit. If you need help with that ask on the discord thread!

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,55 @@
...
+        {
+            toolsMap[toolname] = false;
+        }
+    };
+
+    /// @brief Plugin entry function.
+    /// @param cubos @b CUBOS. main class
+    /// @ingroup toolbox-plugin
+    void toolboxPlugin(cubos::engine::Cubos& cubos);
+} // namespace tesseratos
```
roby2014, line 55: Missing EOF empty line

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    ImGui::End();
+}
+
+void tesseratos::toolboxPlugin(Cubos& cubos)
+{
+    cubos.addResource<Toolbox>();
+    cubos.addPlugin(cubos::engine::imguiPlugin);
+    cubos.system(showToolbox).tagged("cubos.imgui");
+}
```
roby2014, line 43: Missing EOF end of line

roby2014 (dismissed):

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
RiscadoA, line 38: We usually move these function definitions to the `.cpp` file. Here you would just keep the declaration
```cpp
bool isOpen(std::string toolName);
```

In the source file, you would define it:
```cpp
bool Toolbox::isOpen(std::string toolName)
{
  ...
}
```
(Same for the other methods)

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    ///
+    /// ## Resources
+    /// - @ref Toolbox
+
+    /// @brief Resource which manages other tools windows.
+    class Toolbox final
+    {
+    public:
+        std::unordered_map<std::string, bool> toolsMap;
```
RiscadoA, line 28: This field should be private, you can move it to the bottom of the class definition, and prefix it with `m`. E.g:
```cpp
private:
     std::unordered_map<std::string, bool> mToolsMap;
};
```

RiscadoA (commented):
Thanks for working on this @teres4! Over all LGTM, just some formatting/styling issues to fix. Also, it seems our static analysis tool which checks for those type of errors is failing :thinking: not just on your PR, I will have to check that

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
teres4: but don't we want to assign a boolean first? false by default right?

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
teres4: okay, im allowed to make other method(s) to be able iterate thru the map right? just making sure

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void Toolbox::close(std::string toolName)
+{
+    mToolsMap[toolName] = false;
+}
+
+void Toolbox::doClick(std::string toolName)
+{
+    mToolsMap[toolName] = !mToolsMap[toolName];
```
DiogoMendonc-a, line 38: Should we have a some kind of check here?
For tools that keep an internal state it doesn't make sense to me to be able to just close them while that state is altered.
For example, the palette editor. If there's an unsaved edit, does it make sense to have those unsaved changes hidden in a closed tool?

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    ///
+    /// ## Resources
+    /// - @ref Toolbox
+
+    /// @brief Resource which manages other tools windows.
+    class Toolbox final
+    {
+    public:
+        bool isOpen(std::string toolName);
```
DiogoMendonc-a, line 26: Don't forget to document these functions

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
```diff
@@ -0,0 +1,71 @@
...
+    return mToolsMap.end();
+}
+
+static void showToolbox(Write<Toolbox> toolbox)
+{
+    ImGui::Begin("Toolbox");
+
+    for (auto it = toolbox->begin(); it != toolbox->end(); ++it)
+    {
+        if (ImGui::Button(data(it->first)))
```
DiogoMendonc-a, line 57: I think you should use something here that shows whether the tool is open or not, like a checkbox

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
```diff
@@ -0,0 +1,71 @@
...
+auto Toolbox::end()
+{
+    return mToolsMap.end();
+}
+
+static void showToolbox(Write<Toolbox> toolbox)
+{
+    ImGui::Begin("Toolbox");
+
+    for (auto it = toolbox->begin(); it != toolbox->end(); ++it)
```
DiogoMendonc-a, line 55: using the modern for loop would make this more readable. something like `auto& [tool, open] : toolbox`

DiogoMendonc-a (dismissed):

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <cubos/engine/cubos.hpp>
+
+namespace tesseratos
+{
+    /// @defgroup tesseratos-toolbox-plugin Toolbox
+    /// @ingroup tesseratos
+    /// @brief Adds a resource used to keep track of whether each tool is open or not.
+    ///
```
RiscadoA, line 17: Unnecessary empty doc line here
```suggestion
```

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
+/// @dir
+/// @brief @ref toolbox-plugin plugin directory.
```
RiscadoA, line 2:
```suggestion
/// @brief @ref tesseratos-toolbox-plugin plugin directory.
```

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+/// @file
+/// @brief Plugin entry point.
+/// @ingroup toolbox-plugin
```
RiscadoA, line 6:
```suggestion
/// @ingroup tesseratos-toolbox-plugin
```

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+        auto end();
+
+    private:
+        std::unordered_map<std::string, bool> mToolsMap;
+    };
+
+    /// @brief Plugin entry function.
+    /// @param cubos @b CUBOS. main class
+    /// @ingroup toolbox-plugin
```
RiscadoA, line 44:
```suggestion
    /// @ingroup tesseratos-toolbox-plugin
```

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// @defgroup tesseratos-toolbox-plugin Toolbox
+    /// @ingroup tesseratos
+    /// @brief Adds a resource used to keep track of whether each tool is open or not.
+    ///
+    ///
+    /// ## Resources
+    /// - @ref Toolbox
+
+    /// @brief Resource which manages other tools windows.
```
RiscadoA, line 22:
```suggestion
    /// @brief Resource which manages other tools windows.
    /// @ingroup tesseratos-toolbox-plugin
```

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace tesseratos
+{
+    /// @defgroup tesseratos-toolbox-plugin Toolbox
+    /// @ingroup tesseratos
+    /// @brief Adds a resource used to keep track of whether each tool is open or not.
+    ///
+    ///
+    /// ## Resources
+    /// - @ref Toolbox
```
RiscadoA, line 20:
```suggestion
    /// - @ref Toolbox
    ///
    /// ## Dependencies
    /// - @ref imguiPlugin
```

RiscadoA: You should also add `- @ref toolboxPlugin` to the dependency docs of the other plugins you changed

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+        void open(std::string toolName);
+
+        void close(std::string toolName);
+
+        void doClick(std::string toolName);
+
+        auto begin();
+
+        auto end();
```
RiscadoA, line 36: These methods are actually templated methods (auto is just a shorter way to write a template), and thus for these types specifically you have to define them here. Otherwise, if someone outside toolbox/plugin.cpp tries to use this method, there will be an error, as the return type is unknown and the function isn't defined.

```suggestion
        auto begin()
        {
            return mToolsMap.begin();
        }

        auto end()
        {
            return mToolsMap.end();
        }
```

RiscadoA (changes requested):

tools/tesseratos/src/tesseratos/toolbox/plugin.cpp
RiscadoA: This is indeed a problem... we should discuss it in the meeting. One solution is to just treat this as 'hiding' the tool and not closing anything.

DiogoMendonc-a: That is very error prone imo. You close the tool with unsaved changes, spend two hours working on other stuff, based on the unsaved changes you have, and then close the program without ever saving because you forget there was dirty stuff in the background

tools/tesseratos/include/tesseratos/toolbox/plugin.hpp
```diff
@@ -0,0 +1,67 @@
...
+        {
+            return mToolsMap.begin();
+        }
+
+        /// @brief Returns the end of the ToolsMap
+        /// @return end of the map
+        auto end() const
+        {
+            return mToolsMap.end();
+        }
```
RiscadoA, line 57:
```suggestion
        /// @brief Opens a tool.
        /// @param toolName Tool name.
        void open(const std::string& toolName);

        /// @brief Closes a tool.
        /// @param toolName Tool name.
        void close(const std::string& toolName);

        /// @brief If the given tool is open, closes it. Otherwise, opens it.
        /// @param toolName Tool name
        void toggle(const std::string& toolName);

        /// @brief Returns the beginning of the map with known tools.
        /// @return Beginning of the map.
        auto begin() const
        {
            return mToolsMap.begin();
        }

        /// @brief Returns the end of the map with known tools.
        /// @return End of the map.
        auto end() const
        {
            return mToolsMap.end();
        }
```

RiscadoA (approved):
LGTM, you can merge whenever :relieved:

PR #791, opened 2023-11-09
Add `JSONSerializer`

Dacops:
# Description

JSONSerializer implementation based on the Serializer class

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] Write new samples.


core/samples/data/ser/json/main.cpp
```diff
@@ -0,0 +1,159 @@
...
+        return json.dump();
+    }
+
+protected:
+    bool decompose(const Type& type, const void* value) override;
+
+private:
+    nlohmann::json json;
+};
+/// [JSON serializer]
```
RiscadoA, line 39: Instead of declaring this here, this should be in a new file `include/cubos/core/data/ser/json.hpp`. You can copy it from DebugSerializer's `debug.hpp`.

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,125 @@
...
+        return json.dump();
+    }
+
+protected:
+    bool decompose(const Type& type, const void* value) override;
+
+private:
+    nlohmann::json json;
+};
+/// [JSON serializer]
```
RiscadoA, line 39: Same as above

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,31 @@
+/// @file
+/// @brief Class @ref cubos::core::data::JSONSerializer.
+/// @ingroup core-data-ser
+
+#pragma once
+
+/// [JSON package]
```
RiscadoA, line 7: This comment shouldn't be here

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <nlohmann/json.hpp>
+
+#include <cubos/core/data/ser/serializer.hpp>
+#include <cubos/core/memory/stream.hpp>
```
RiscadoA, line 11: You don't use streams, you can remove this include

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace cubos::core::data
+{
+    /// @brief Implementation of the abstract Serializer class for serializing to JSON.
+    class JSONSerializer : public Serializer
+    {
+    public:
+        /// @brief Constructs.
```
RiscadoA, line 19: You can remove this constructor, in this case the default us enough.

RiscadoA (changes requested):
Don't forget to rebase on main!

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,31 @@
...
+        /// @brief Outputs the serialized JSON.
+        std::string output() const;
+
+    protected:
+        bool decompose(const reflection::Type& type, const void* value) override;
+
+    private:
+        nlohmann::json mJSON = nlohmann::json::object();    ///< The JSON object that holds the serialized data.
+    };
+} // namespace cubos::core::datax
```
RiscadoA, line 31: Extra x at the end, and missing newline at EOF

core/samples/data/ser/custom/main.cpp
```diff
@@ -104,4 +104,4 @@ int main()
 
 /// [Output]
 // [{x: 1, y: 2, z: 3, }, {x: 4, y: 5, z: 6, }, ]
-/// [Output]
+/// [Output]
```
RiscadoA, line 107: Missing newline at EOF

core/include/cubos/core/data/ser/json.hpp
Dacops: but there's no default constructor in Serializer?

RiscadoA: You don't have to define it, the compiler does it implicitly.

Dacops: ok, so to where do I move the AUTO_HOOK calls?

RiscadoA: I'm retarded, disregard kek

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// @brief Implementation of the abstract Serializer class for serializing to JSON.
+    class JSONSerializer : public Serializer
+    {
+    public:
+        /// @brief Constructs.
+        JSONSerializer();
+
+        /// @brief Outputs the serialized JSON.
+        nlohmann::json output();
```
RiscadoA, line 21:
```suggestion
        /// @brief Outputs the serialized JSON. @ref write must be called before this.
        /// @return JSON.
        nlohmann::json output();
```

core/include/cubos/core/data/ser/json.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+    protected:
+        bool decompose(const reflection::Type& type, const void* value) override;
+
+    private:
+        nlohmann::json mJSON = nlohmann::json::object(); ///< The JSON object that holds the serialized data.
```
RiscadoA, line 27:
```suggestion
        nlohmann::json mJSON = nlohmann::json::object(); ///< Resulting JSON object.
```

core/samples/data/ser/json/main.cpp
RiscadoA, line 1: You should also create a page for this sample, check out `page.md` files in other samples for an example.

core/samples/data/ser/json/main.cpp
```diff
@@ -0,0 +1,40 @@
+#include <cubos/core/data/ser/json.hpp>
+#include <cubos/core/log.hpp>
+#include <cubos/core/memory/stream.hpp>
+
+using cubos::core::data::JSONSerializer;
+using cubos::core::memory::Stream;
```
RiscadoA, line 6:
```suggestion
```

core/samples/data/ser/json/main.cpp
```diff
@@ -0,0 +1,40 @@
+#include <cubos/core/data/ser/json.hpp>
+#include <cubos/core/log.hpp>
+#include <cubos/core/memory/stream.hpp>
```
RiscadoA, line 3:
```suggestion
```

core/samples/data/ser/json/main.cpp
```diff
@@ -0,0 +1,40 @@
+#include <cubos/core/data/ser/json.hpp>
+#include <cubos/core/log.hpp>
```
RiscadoA, line 2:
```suggestion
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
+#include <cubos/core/log.hpp>
+
+/// [JSON serializer]
+#include <cubos/core/data/ser/json.hpp>
+
+using cubos::core::data::JSONSerializer;
+using cubos::core::reflection::Type;
+
+auto jsonVal = nlohmann::json::object(); ///< The current JSON value that is being serialized.
```
RiscadoA, line 9: Do not use a global variable. Instead, write directly to `this->mJSON` on the hooks.

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+    AUTO_HOOK(int64_t);
+    AUTO_HOOK(uint8_t);
+    AUTO_HOOK(uint16_t);
+    AUTO_HOOK(uint32_t);
+    AUTO_HOOK(uint64_t);
+    AUTO_HOOK(float);
+    AUTO_HOOK(double);
+    AUTO_HOOK(std::string);
+}
+/// [Setting up hooks]
```
RiscadoA, line 40:
```suggestion
```
These comments are only used in samples to create snippets which can be referenced from the markdown pages.

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <cubos/core/reflection/type.hpp>
+
+using cubos::core::reflection::ArrayTrait;
+using cubos::core::reflection::FieldsTrait;
+
+bool JSONSerializer::decompose(const Type& type, const void* value)
+{
+    if (type.has<ArrayTrait>())
+    {
+        nlohmann::json jsonArr; ///< Hold current JSON array
```
RiscadoA, line 59:
```suggestion
        auto jsonArr = nlohmann::json::array();
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+        mJSON = jsonArr;
+        return true;
+    }
+    /// [Decomposing types]
+
+    /// [Decomposing types with fields]
+    if (type.has<FieldsTrait>())
+    {
+        nlohmann::json jsonObj; ///< Hold current JSON object
```
RiscadoA, line 80:
```suggestion
        auto jsonObj = nlohmann::json::object();
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+        nlohmann::json jsonArr; ///< Hold current JSON array
+
+        const auto& arrayTrait = type.get<ArrayTrait>();
+
+        for (const auto* element : arrayTrait.view(value))
+        {
+            if (!this->write(arrayTrait.elementType(), element))
+            {
+                return false;
```
RiscadoA, line 67: When returning false, we should probably print a warning so that the dev knows what failed during serialization. Here you could do something like `CUBOS_WARN("Could not deserialize array element");`

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <cubos/core/reflection/traits/array.hpp>
+#include <cubos/core/reflection/traits/fields.hpp>
+#include <cubos/core/reflection/type.hpp>
+
+using cubos::core::reflection::ArrayTrait;
+using cubos::core::reflection::FieldsTrait;
+
+bool JSONSerializer::decompose(const Type& type, const void* value)
+{
+    if (type.has<ArrayTrait>())
```
RiscadoA, line 57: You should also handle de `DictionaryTrait` and the `StringConversionTrait`. Check out the `DebugSerializer` for examples.

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+        nlohmann::json jsonObj; ///< Hold current JSON object
+
+        for (const auto& [field, fieldValue] : type.get<FieldsTrait>().view(value))
+        {
+            if (!this->write(field->type(), fieldValue))
+            {
+                return false;
+            }
+            jsonObj[field->name()] = jsonVal;
```
RiscadoA, line 88:
```suggestion
            jsonObj[field->name()] = mJSON;
```

RiscadoA (changes requested):
Thanks for working on this! This is an important part of the code which we don't want to randomly explode on devs, so we should cover it on the tests. Check out the [JSON deserializer test](https://github.com/GameDevTecnico/cubos/blob/main/core/tests/data/des/json.cpp) to see something similar

core/samples/data/ser/json/main.cpp
Dacops: Stream is used to show the output: line 34: "Stream::stdOut.print(json.dump());"

core/samples/data/ser/json/main.cpp
Dacops: Stream is used to show the output: line 34: "Stream::stdOut.print(json.dump());"

core/src/cubos/core/data/ser/json.cpp
Dacops: jsonVal needs to be local to be used in the macro. mJSON is a private member of JSONSerializer

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+using cubos::core::data::JSONSerializer;
+using cubos::core::reflection::Type;
+
+auto jsonVal = nlohmann::json::object(); ///< The current JSON value that is being serialized.
+
+/// [MACRO to add hooks for primitive types]
+#define AUTO_HOOK(type, ...)                                                                                           \
+    this->hook<type>([](const type& value) {                                                                           \
+        jsonVal = value;                                                                                               \
```
RiscadoA, line 14:
```suggestion
    this->hook<type>([this](const type& value) {                                                                           \
        mJSON = value;                                                                                               \
```

core/src/cubos/core/data/ser/json.cpp
RiscadoA: I donºt understand wdym, the hook is internal to the serializer and thus you can access `mJSON` from there. Suggested a change on the hook which allows you to remove `jsonVal`

core/samples/data/ser/json/main.cpp
```diff
@@ -0,0 +1,46 @@
...
+    outer1.push_back(inner1);
+
+    std::vector<std::vector<glm::ivec3>> outer2;
+    outer2.push_back(inner2);
+
+    vec.push_back(outer1);
+    vec.push_back(outer2);
+
+    JSONSerializer ser{};
+    ser.write(vec);
```
RiscadoA, line 37:
```suggestion
    std::vector<std::vector<std::vector<glm::ivec3>>> vec = {
      {{{1, 2, 3}, {4, 5, 6}}},
      {{{7, 8, 9}, {10, 11, 12}}}
    };
    
    JSONSerializer ser{};
    ser.write(vec);
```

core/samples/data/ser/json/page.md
```diff
@@ -0,0 +1,28 @@
...
+
+@see Full source code [here](https://github.com/GameDevTecnico/cubos/tree/main/core/samples/data/ser/json).
+
+You'll need to include 
+@ref core/data/ser/json.hpp. For simplicity, in this sample we'll use
+the following aliases:
+
+@snippet data/ser/custom/main.cpp Include
+
+Using the json serializer is as simple as calling @ref
```
RiscadoA, line 13:
```suggestion
Using the JSON serializer is as simple as calling @ref
```

core/samples/data/ser/json/page.md
```diff
@@ -0,0 +1,28 @@
...
+cubos::core::data::JSONSerializer::write "JSONSerializer::write" on the data we want to
+serialize.
+
+# Update this section with final sample version
+
+In this case, we'll be serializing a `std::vector<glm::ivec3>`, which is
+an array of objects with three `int32_t` fields.
+
+# Update this section with final sample version
```
RiscadoA, line 22: ?

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,128 @@
+#include <cubos/core/log.hpp>
+
+/// [JSON serializer]
```
RiscadoA, line 3:
```suggestion
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <cubos/core/data/ser/json.hpp>
+
+using cubos::core::data::JSONSerializer;
+using cubos::core::reflection::Type;
+
+/// [MACRO to add hooks for primitive types]
```
RiscadoA, line 9:
```suggestion
// Macro used to reduce code duplication for primitive type hooks.
```
I think you misunderstood those /// [] comments: they're meant to be used only in samples, so that we can reference them in the page itself. In ordinary source files, you can just use normal comments formatting.

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        auto jsonArr = nlohmann::json::array();
+
+        const auto& arrayTrait = type.get<ArrayTrait>();
+
+        for (const auto* element : arrayTrait.view(value))
+        {
+            if (!this->write(arrayTrait.elementType(), element))
+            {
+                CUBOS_WARN("Could not deserialize array element");
```
RiscadoA, line 69:
```suggestion
                CUBOS_WARN("Could not serialize array element");
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+        mJSON = jsonArr;
+        return true;
+    }
+
+    if (type.has<StringConversionTrait>())
+    {
+        const auto& trait = type.get<StringConversionTrait>();
+        mJSON = trait.into(value);
+    }
```
RiscadoA, line 83:
```suggestion
        return true;
    }
```

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    if (type.has<FieldsTrait>())
+    {
+        auto jsonObj = nlohmann::json::object();
+
+        for (const auto& [field, fieldValue] : type.get<FieldsTrait>().view(value))
+        {
+            if (!this->write(field->type(), fieldValue))
+            {
+                return false;
```
RiscadoA, line 93:
```suggestion
                CUBOS_WARN("Could not serialize field '{}'", field->name());
                return false;
```

core/tests/data/ser/json.cpp
```diff
@@ -0,0 +1,93 @@
+#include <doctest/doctest.h>
+#include <glm/vec3.hpp>
+
+#include <cubos/core/data/ser/debug.hpp>
```
RiscadoA, line 4:
```suggestion
```

core/tests/data/ser/json.cpp
```diff
@@ -0,0 +1,93 @@
...
+    ser.write<type>(value);                                                                                            \
+    json = ser.output();                                                                                               \
+    CHECK((json.dump()) == (expected));
+
+TEST_CASE("data::JSONSerializer")
+{
+    JSONSerializer ser{};
+    auto json = ser.output();
+
+    /// [Primitive types]
```
RiscadoA, line 29:
```suggestion
    // Primitive types
```
Do the same for the comments below

core/tests/data/ser/json.cpp
```diff
@@ -0,0 +1,93 @@
...
+            return Type::create("ExampleStruct")
+                .with(FieldsTrait{}
+                          .withField("a", &ExampleStruct::a)
+                          .withField("b", &ExampleStruct::b)
+                          .withField("c", &ExampleStruct::c));
+        }
+        int a;
+        float b;
+        std::string c;
+    };
```
RiscadoA, line 90: Move the struct outside the function, and wrap it with `namespace {}`, as we do in the JSON deserializer test.

core/tests/data/ser/json.cpp
```diff
@@ -0,0 +1,93 @@
...
+    /// [String]
+    AUTO_TEST(std::string, "Hello, world!", "\"Hello, world!\"");
+
+    /// [UUID]
+    AUTO_TEST(uuids::uuid, *uuids::uuid::from_string("f7063cd4-de44-47b5-b0a9-f0e7a558c9e5"),
+              "\"f7063cd4-de44-47b5-b0a9-f0e7a558c9e5\"");
+
+    /// [Map]
+    std::map<std::string, bool> map{{"false", false}, {"true", true}};
+    AUTO_TEST(decltype(map), map, "{\"false\":false,\"true\":true}");
```
RiscadoA, line 52: You should also cover a map which doesn't have strings as keys, I suspect they will fail. E.g., `std::map<int, std::string>`. In JSON, you can't have a key of a structured type, unlike values. All keys must be strings. Thus, you will have to handle it specially. For example, in the JSON deserializer, I stored an extra field in the deserializer class, `mIsKey`. On hooks, if the `mIsKey` flag is set to true, instead of storing into `mJSON`, you would store into a `mKey` string variable. Then, on the dictionary trait, before calling write for the key, you would set `mIsKey` to true, and afterwards extract the result from the `mKey` variable, isntead of `mJSON`.

core/tests/data/ser/json.cpp
```diff
@@ -0,0 +1,93 @@
...
+                          .withField("b", &ExampleStruct::b)
+                          .withField("c", &ExampleStruct::c));
+        }
+        int a;
+        float b;
+        std::string c;
+    };
+    ExampleStruct myStruct{1, 2.5F, "Hello, world!"};
+    AUTO_TEST(decltype(myStruct), myStruct, "{\"a\":1,\"b\":2.5,\"c\":\"Hello, world!\"}");
+}
```
RiscadoA, line 93: Missing newline at EOF

RiscadoA (changes requested):
Thank you for working on this, sorry for request so many changes :sob: this will be critical infrastructure for things to come so we have to make sure it works properly

core/samples/data/ser/json/page.md
Dacops: forgot to update to last sample version lul

core/src/cubos/core/data/ser/json.cpp
Dacops: I don't get this change. It also fails the uuid test:
`CHECK( (json.dump()) == ("\"f7063cd4-de44-47b5-b0a9-f0e7a558c9e5\"") ) is NOT correct!
  values: CHECK( "Hello, world!" == "f7063cd4-de44-47b5-b0a9-f0e7a558c9e5" )`

core/tests/data/ser/json.cpp
Dacops: what does wrapping it with namespace{} does? I've just moved the struct to outside the functions and it still compile / passes the test.

core/src/cubos/core/data/ser/json.cpp
Dacops: nvm got it

core/samples/data/ser/json/page.md
RiscadoA, line 1: You need to add a reference to this page in `core/samples/data/ser/page.md`

RiscadoA (approved):
Looks good to me :) just missing one thing: you must add a reference to the sample in the serializer samples page for it to appear on the index.

RiscadoA (changes requested):
Ouh, just noticed, dont forget to fix your commit msgs before merging

core/src/cubos/core/data/ser/json.cpp
```diff
@@ -0,0 +1,131 @@
...
+    AUTO_HOOK(bool);
+    AUTO_HOOK(char);
+    AUTO_HOOK(int8_t);
+    AUTO_HOOK(int16_t);
+    AUTO_HOOK(int32_t);
+    AUTO_HOOK(int64_t);
+    AUTO_HOOK(uint8_t);
+    AUTO_HOOK(uint16_t);
+    AUTO_HOOK(uint32_t);
+    AUTO_HOOK(uint64_t);
```
RiscadoA, line 30: After rebasing on `origin/main` you'll have to change these hooks to the new primitive integer types.

RiscadoA (changes requested):

RiscadoA (approved):
LGTM! :tada:

PR #795, opened 2023-11-11
Replace `typeid` by reflection on asset bridges

roby2014:
~~Blocked by #799~~ and uses commits from #775

# Description

- Replace `typeid` by reflection on asset bridges
- Fixed tesseratos plugins where asset types are compared
- Fixed assets samples to use this new system

## Checklist

- [X] Compiled and executed all assets samples

### DIFF
--- a/engine/include/cubos/engine/assets/assets.hpp
+++ b/engine/include/cubos/engine/assets/assets.hpp
@@ -13,6 +13,7 @@
 
 #include <cubos/core/memory/guards.hpp>
 #include <cubos/core/memory/type_map.hpp>
+#include <cubos/core/reflection/reflect.hpp>
 
 #include <cubos/engine/assets/bridge.hpp>
 #include <cubos/engine/assets/meta.hpp>
@@ -141,7 +142,7 @@ namespace cubos::engine
             // Create a strong handle to the asset, so that the asset starts loading if it isn't already.
             auto strong = this->load(handle);
             auto lock = this->lockRead(strong);
-            auto data = static_cast<const T*>(this->access(strong, typeid(T), lock, false));
+            auto data = static_cast<const T*>(this->access(strong, core::reflection::reflect<T>(), lock, false));
             return AssetRead<T>(*data, std::move(lock));
         }
 
@@ -199,7 +200,8 @@ namespace cubos::engine
         template <typename T>
         inline Asset<T> create(T data)
         {
-            return this->create(typeid(T), new T(std::move(data)), [](void* data) { delete static_cast<T*>(data); });
+            return this->create(core::reflection::reflect<T>(), new T(std::move(data)),
+                                [](void* data) { delete static_cast<T*>(data); });
         }
 
         /// @brief Stores the given asset data in memory, associated with the given handle.
@@ -215,7 +217,7 @@ namespace cubos::engine
         template <typename T>
         inline AnyAsset store(AnyAsset handle, T data)
         {
-            return this->store(handle, typeid(T), new T(std::move(data)),
+            return this->store(handle, core::reflection::reflect<T>(), new T(std::move(data)),
                                [](void* data) { delete static_cast<T*>(data); });
         }
 
@@ -231,7 +233,7 @@ namespace cubos::engine
         ///
         /// @param handle Handle to check the type for.
         /// @return Asset type.
-        std::type_index type(const AnyAsset& handle) const;
+        const core::reflection::Type& type(const AnyAsset& handle) const;
 
     private:
         /// @brief Represents a known asset - may or may not be loaded.
@@ -247,9 +249,9 @@ namespace cubos::engine
             std::shared_mutex mutex;          ///< Mutex for the asset data.
             std::condition_variable_any cond; ///< Triggered when the asset is loaded.
 
-            void* data{nullptr};       ///< Pointer to the asset data, if loaded. Otherwise, nullptr.
-            std::type_index type;      ///< The type of the asset data - initially typeid(void).
-            void (*destructor)(void*); ///< The destructor for the asset data - initially nullptr.
+            void* data{nullptr};                         ///< Pointer to the asset data, if loaded. Otherwise, nullptr.
+            const core::reflection::Type* type{nullptr}; ///< Type of the asset data.
+            void (*destructor)(void*);                   ///< Destructor for the asset data - initially nullptr.
         };
 
         /// @brief Stores all data necessary to load an asset.
@@ -264,7 +266,7 @@ namespace cubos::engine
         /// @param data Asset data to store.
         /// @param destructor Destructor for the asset data.
         /// @return Strong handle to the asset.
-        AnyAsset create(std::type_index type, void* data, void (*destructor)(void*));
+        AnyAsset create(const core::reflection::Type& type, void* data, void (*destructor)(void*));
 
         /// @brief Untyped version of @ref store().
         /// @param handle Handle to associate the asset with.
@@ -272,7 +274,7 @@ namespace cubos::engine
         /// @param data Asset data to store.
         /// @param destructor Destructor for the asset data.
         /// @return Strong handle to the asset.
-        AnyAsset store(AnyAsset handle, std::type_index type, void* data, void (*destructor)(void*));
+        AnyAsset store(AnyAsset handle, const core::reflection::Type& type, void* data, void (*destructor)(void*));
 
         /// @brief Gets a pointer to the asset data associated with the given handle.
         ///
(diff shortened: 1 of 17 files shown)


engine/include/cubos/engine/assets/assets.hpp
```diff
@@ -39,6 +41,10 @@ namespace cubos::engine
     template <typename T>
     using AssetWrite = core::memory::WriteGuard<T, std::unique_lock<std::shared_mutex>>;
 
+    using Type = cubos::core::reflection::Type;
+
+    using cubos::core::reflection::reflect;
```
RiscadoA, line 46: Don't do this in headers - the using statements above are an exception because they're part of the interface of the Assets plugin and they greatly simplify the user's code. These aliases you defined are only useful in the implementation itself.

engine/include/cubos/engine/assets/assets.hpp
```diff
@@ -13,6 +13,8 @@
 
 #include <cubos/core/memory/guards.hpp>
 #include <cubos/core/memory/type_map.hpp>
+#include <cubos/core/reflection/type.hpp>
```
RiscadoA, line 16: Not sure but I think you only need the header below (`reflect`).

engine/include/cubos/engine/assets/bridge.hpp
```diff
@@ -28,11 +29,13 @@ namespace cubos::engine
     class AssetBridge
     {
     public:
+        using Type = cubos::core::reflection::Type;
```
RiscadoA, line 32: I would not pollute the namespace of the bridge too.

engine/include/cubos/engine/assets/bridges/binary.hpp
```diff
@@ -27,7 +28,7 @@ namespace cubos::engine
         /// @brief Constructs a bridge.
         /// @param littleEndian Whether to use little endian byte order.
         BinaryBridge(bool littleEndian = true)
-            : FileBridge(typeid(T))
+            : FileBridge(cubos::core::reflection::reflect<T>())
```
RiscadoA, line 31:
```suggestion
            : FileBridge(core::reflection::reflect<T>())
```

engine/include/cubos/engine/assets/bridge.hpp
```diff
@@ -10,6 +10,7 @@
 #include <typeindex>
 
 #include <cubos/engine/assets/asset.hpp>
+#include <cubos/core/reflection/type.hpp>
```
RiscadoA, line 13:
```suggestion
#include <cubos/core/reflection/reflect.hpp>
```

engine/include/cubos/engine/assets/bridges/binary.hpp
```diff
@@ -7,6 +7,7 @@
 #include <cubos/core/data/old/binary_deserializer.hpp>
 #include <cubos/core/data/old/binary_serializer.hpp>
 #include <cubos/core/log.hpp>
+#include <cubos/core/reflection/reflect.hpp>
```
RiscadoA, line 10:
```suggestion
```
Shouldn't be needed as base bridge includes it

engine/include/cubos/engine/assets/bridges/file.hpp
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cubos/core/memory/stream.hpp>
+#include <cubos/core/reflection/type.hpp>
```
RiscadoA, line 8:
```suggestion
```
Shouldn't be needed as base bridge includes it

engine/include/cubos/engine/assets/bridges/file.hpp
```diff
@@ -23,9 +24,9 @@ namespace cubos::engine
     public:
         /// @brief Constructs a bridge.
         ///
-        /// @param index Type of assets loaded by the bridge.
-        explicit FileBridge(std::type_index index)
-            : AssetBridge(index)
+        /// @param type Type of assets loaded by the bridge.
+        explicit FileBridge(const cubos::core::reflection::Type& type)
```
RiscadoA, line 28:
```suggestion
        explicit FileBridge(const core::reflection::Type& type)
```

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -30,7 +31,7 @@ namespace cubos::engine
         ///
         /// @param indentation Indentation level to use when saving the JSON file.
         JSONBridge(int indentation = 4)
-            : FileBridge(typeid(T))
+            : FileBridge(cubos::core::reflection::reflect<T>())
```
RiscadoA, line 34:
```suggestion
            : FileBridge(core::reflection::reflect<T>())
```

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -7,6 +7,7 @@
 #include <cubos/core/data/old/json_deserializer.hpp>
 #include <cubos/core/data/old/json_serializer.hpp>
 #include <cubos/core/log.hpp>
+#include <cubos/core/reflection/reflect.hpp>
```
RiscadoA, line 10:
```suggestion
```
Shouldn't be needed as base bridge includes it

engine/include/cubos/engine/scene/bridge.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include <cubos/engine/assets/bridge.hpp>
 #include <cubos/engine/scene/scene.hpp>
+#include <cubos/core/reflection/reflect.hpp>
```
RiscadoA, line 9:
```suggestion
```
Shouldn't be needed as base bridge includes it

engine/include/cubos/engine/scene/bridge.hpp
```diff
@@ -49,7 +50,7 @@ namespace cubos::engine
         /// @brief Constructs a bridge.
         ///
         SceneBridge()
-            : AssetBridge(typeid(Scene))
+            : AssetBridge(cubos::core::reflection::reflect<Scene>())
```
RiscadoA, line 53:
```suggestion
            : AssetBridge(core::reflection::reflect<Scene>())
```

RiscadoA (changes requested):
Thanks for working on this @roby2014! Didn't realize it also meant changing the internals of the `Assets` resource :pensive:

engine/include/cubos/engine/assets/bridges/json.hpp
roby2014: Shouldn't every file include what it needs?

RiscadoA: You're right, whoops

engine/include/cubos/engine/assets/assets.hpp
roby2014: Well it uses the `Type` class that comes from `type.hpp`

RiscadoA: But does it use any of its methods? `reflect.hpp` forward declares it, if you're just storing references to it its enough to include the other header.

engine/include/cubos/engine/assets/bridge.hpp
roby2014: Again, we use `Type` in this file which comes from type.hpp` (?)

RiscadoA (approved):

engine/include/cubos/engine/assets/assets.hpp
```diff
@@ -247,9 +249,9 @@ namespace cubos::engine
             std::shared_mutex mutex;          ///< Mutex for the asset data.
             std::condition_variable_any cond; ///< Triggered when the asset is loaded.
 
-            void* data{nullptr};       ///< Pointer to the asset data, if loaded. Otherwise, nullptr.
-            std::type_index type;      ///< The type of the asset data - initially typeid(void).
-            void (*destructor)(void*); ///< The destructor for the asset data - initially nullptr.
+            void* data{nullptr};                   ///< Pointer to the asset data, if loaded. Otherwise, nullptr.
+            core::reflection::Type* type{nullptr}; ///< The type of the asset data.
```
RiscadoA, line 253:
```suggestion
            const core::reflection::Type* type{nullptr}; ///< Type of the asset data.
```

engine/include/cubos/engine/assets/assets.hpp
```diff
@@ -247,9 +249,9 @@ namespace cubos::engine
...
+            void (*destructor)(void*);             ///< The destructor for the asset data - initially nullptr.
```
RiscadoA, line 254:
```suggestion
            void (*destructor)(void*);             ///< Destructor for the asset data - initially nullptr.
```

engine/src/cubos/engine/assets/assets.cpp
```diff
@@ -364,15 +366,15 @@ AnyAsset Assets::store(AnyAsset handle, std::type_index type, void* data, void (
...
+                     assetEntry->type->name());
         assetEntry->destructor(assetEntry->data);
     }
 
     // Mark it as loaded and set its data.
     assetEntry->status = Status::Loaded;
     assetEntry->version++;
     assetEntry->data = data;
-    assetEntry->type = type;
+    assetEntry->type = const_cast<Type*>(&type);
```
RiscadoA, line 377:
```suggestion
    assetEntry->type = &type;
```

RiscadoA (approved):
Thanks for working on this! Overall LGTM, but please remove that `const_cast` - types should never be referenced in a mutable way, only during their creation do we provide mutable access.

PR #804, opened 2023-11-12
Emit Collision Events

luishfonseca:
# Description

Emits AABB collision events, an hack just for DevGamm. This should eventually be redesigned once narrow phase is implemented.


engine/include/cubos/engine/collisions/collision_event.hpp
```diff
@@ -0,0 +1,23 @@
+/// @file
+/// @brief Component @ref cubos::engine::CollisionEvent.
```
RiscadoA, line 2: Not a component

RiscadoA (approved):
Thanks for working on this! Will be really cool if we can integrate this with physics for the demo @fallenatlas 👀

RiscadoA (approved):

PR #813, opened 2023-11-13
Remove bool input label from `DataInspector`

roby2014:
Oops


RiscadoA (approved):
Nice catch :relieved:

PR #824, opened 2023-11-18
Implement equality operator for `reflection::Type`

roby2014:
# Description

- add equality operator for `reflection::Type` that compares by address
- add simple tests for it

## Checklist

- [X] Self-review changes.
- [X] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/core/include/cubos/core/reflection/type.hpp
+++ b/core/include/cubos/core/reflection/type.hpp
@@ -82,6 +82,11 @@ namespace cubos::core::reflection
             return *static_cast<const T*>(this->get(Type::id<T>()));
         }
 
+        /// @brief Equality operator.
+        /// @param other Type object to compare.
+        /// @return Whether the objects have the same address, which indicates equality.
+        bool operator==(const Type& other) const;
+
     private:
         ~Type();
 
--- a/core/src/cubos/core/reflection/type.cpp
+++ b/core/src/cubos/core/reflection/type.cpp
@@ -57,6 +57,17 @@ const void* Type::get(uintptr_t id) const
     CUBOS_FAIL("No such trait in type \"{}\"", this->name());
 }
 
+bool Type::operator==(const Type& other) const
+{
+    if (this == &other)
+    {
+        return true;
+    }
+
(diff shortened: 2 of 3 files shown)


core/include/cubos/core/reflection/type.hpp
```diff
@@ -82,6 +82,11 @@ namespace cubos::core::reflection
             return *static_cast<const T*>(this->get(Type::id<T>()));
         }
 
+        /// @brief Equality operator.
+        /// @param other Type object to compare.
+        /// @return True if the objects have the same address, indicating equality.
```
RiscadoA, line 87:
```suggestion
        /// @return Whether the objects have the same address, which indicates equality.
```

core/src/cubos/core/reflection/type.cpp
```diff
@@ -57,6 +57,16 @@ const void* Type::get(uintptr_t id) const
...
 }
 
+bool Type::operator==(const Type& other) const
+{
+    bool equal = this == &other;
+    if (!equal)
+    {
+        CUBOS_ASSERT(this->mName != other.mName, "Detected type naming collision detected!");
+    }
+    return equal;
```
RiscadoA, line 67:
```suggestion
    if (this == &other)
    {
        return true;
    }
    
    CUBOS_ASSERT(this->mName != other.mName, "Two types cannot be created with the same");
    return false;
```

core/tests/reflection/type.cpp
```diff
@@ -34,5 +34,12 @@ TEST_CASE("reflection::Type")
         CHECK(type.get<int>() == 42);
     }
 
+    SUBCASE("with two traits")
+    {
+        auto& anotherType = type;
+        REQUIRE(anotherType == type);
+        Type::destroy(anotherType);
```
RiscadoA, line 41:
```suggestion
        auto& anotherType = type;
        REQUIRE(anotherType == type);
```
Destroy is already being called on line 44. This is a double delete error.
Also, would be cool to also test inequality, e.g., create another type with a different name and assert that == returns false.

RiscadoA (approved):
Thanks for working on this!

Issue #274, opened 2022-12-14
Add logs/console UI


RiscadoA:
Add a plugin which shows an `ImGui` window with the log of the engine. It should be possible to filter by log level.

This will require extending the core `log.hpp` in some way so that the logging is outputted to the window.

PR #830, opened 2023-11-19
Implement `Logger` and remove `spdlog`/`fmt`

RiscadoA:
# Description

Implements our own `Logger`, which no longer uses `spdlog` (functionally identical tho).
Also comes with a function which can be used to read the logs. Will be useful for implementing #274.
This refactor enables us to drop the old `Debug{}` shenanigan used to print non-primitive types, and to remove both the `fmt` and `spdlog` dependencies. One of disadvantages is having to include the external reflection header when trying to print external types, but I think its the best approach out of the two options.
This should also allow us to remove the old `DebugSerializer` entirely, but that will be left to the PR which nukes `data::old`.

Check out the commit messages, they are helpful to understand what I needed to change.

## Checklist

- [x] Self-review changes.
- [x] ~~Ensure test coverage.~~ (is it worth it to test the logger? :thinking:)
- [x] Update samples.

### DIFF
--- a/core/src/cubos/core/log.cpp
+++ b/core/src/cubos/core/log.cpp
@@ -1,26 +1,204 @@
-#include <spdlog/sinks/basic_file_sink.h>
-#include <spdlog/sinks/stdout_color_sinks.h>
+#include <chrono>
+#include <ctime>
+#include <mutex>
+#include <vector>
 
+#include <cubos/core/data/ser/debug.hpp>
 #include <cubos/core/log.hpp>
 
-void cubos::core::initializeLogger()
+using cubos::core::Logger;
+
+namespace
 {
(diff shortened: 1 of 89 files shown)


core/src/cubos/core/log.cpp
```diff
@@ -1,26 +1,204 @@
...
+static State& state()
+{
+    static State state{};
+    return state;
+}
+
+/// @brief Returns a string representing the given log level.
+/// @param level Log level.
+/// @return String representation.
+static const char* levelString(Logger::Level level)
```
roby2014, line 33: Maybe `EnumTrait::toString` can be used here

roby2014 (changes requested):

roby2014 (approved):
👍🏻 👍🏻

PR #833, opened 2023-11-19
Remove `[un]packs` from the ECS

RiscadoA:
# Description

Removes all `data::old::Package` related code from the ECS, which is not needed anymore after adding reflection and allowing access to the components through `World`.
Fixed a small unrelated bug introduced in #830, where I forgot to update the format string of the OpenGL message callback.

## Checklist

- [x] Self-review changes.


roby2014 (approved):
Finally !!! 🎉

PR #862, opened 2023-12-02
Add a new jsonbridge which uses the new serializers

Dacops:


engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -26,32 +26,26 @@ namespace cubos::engine
...
             // Dump the file stream into a string and initialize a JSON deserializer with it.
-            std::string json{};
-            stream.readUntil(json, nullptr);
-            core::data::old::JSONDeserializer deserializer{json};
+            std::string jsonStr;
+            stream.readUntil(jsonStr, nullptr);
+            core::data::JSONDeserializer deserializer{};
+
+            // New JSONDeserializer() receives a JSON object to deserialize from
+            nlohmann::json json = nlohmann::json::parse(jsonStr);
```
RiscadoA, line 43: Be careful, we need to handle invalid JSON files gracefully. You should catch the thrown exception and return false (check `engine/src/cubos/engine/scene/bridge.cpp:44` for an example).

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -26,32 +26,26 @@ namespace cubos::engine
...
+            deserializer.feed(json);
 
             // Deserialize the asset and store it in the asset manager.
             T data{};
             deserializer.read(data);
-            if (deserializer.failed())
-            {
-                CUBOS_ERROR("Could not deserialize asset from JSON file");
-                return false;
-            }
```
RiscadoA, line 49: This deserializer can still fail, but you check it through the read's return value.

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -60,21 +54,16 @@ namespace cubos::engine
...
 
             // Read the asset from the asset manager and serialize it to the file stream.
             auto data = assets.read<T>(handle);
-            serializer.write(*data, nullptr); // JSON serializer does not use names.
-            if (serializer.failed())
-            {
-                CUBOS_ERROR("Could not serialize asset to JSON file");
-                return false;
-            }
+            serializer.write(*data);
```
RiscadoA, line 62: This serializer can still fail,but you check it through the write's return value.

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -60,21 +54,16 @@ namespace cubos::engine
...
 
+            // new JSONSerializer() does not receive a stream to write to, need to write to it manually
```
RiscadoA, line 63: You can remove those comments: future readers won't know about the old serialization system :relieved:

RiscadoA (changes requested):
Thanks for working on this! :) Overall LGTM, just needs some error handling. Merry christmas! :santa:

engine/include/cubos/engine/assets/bridges/json.hpp
RiscadoA: You forgot to remove it!

engine/include/cubos/engine/assets/bridges/json.hpp
```diff
@@ -26,28 +26,36 @@ namespace cubos::engine
...
+            try
+            {
+                json = nlohmann::json::parse(jsonStr);
+            }
+            catch (nlohmann::json::parse_error& e)
+            {
+                CUBOS_ERROR("{}", e.what());
+                return false;
+            }
+            deserializer.feed(json);
```
RiscadoA, line 54:
```suggestion
            stream.readUntil(jsonStr, nullptr);

            // Parse JSON from the string.
            nlohmann::json json{};
            try
            {
                json = nlohmann::json::parse(jsonStr);
            }
            catch (nlohmann::json::parse_error& e)
            {
                CUBOS_ERROR("{}", e.what());
                return false;
            }
            
            // Feed the parsed JSON to a JSON deserializer.
            core::data::JSONDeserializer deserializer{};
            deserializer.feed(json);
```

RiscadoA (changes requested):

core/lib/glad
RiscadoA, line 1: Wait just noticed glad changed, why?

core/lib/json
RiscadoA, line 1: Same thing for JSON

RiscadoA (approved):

PR #874, opened 2023-12-20
Add screenPickerPlugin

tomas7770:
# Description

Adds a new plugin in `cubos::engine`, which adds the resource `ScreenPicker`. This resource provides a texture with entity/gizmo IDs for mouse picking.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/src/cubos/engine/screenpicker/plugin.cpp
+++ b/engine/src/cubos/engine/screenpicker/plugin.cpp
@@ -40,7 +40,7 @@ void cubos::engine::screenPickerPlugin(Cubos& cubos)
 
     cubos.startupSystem(initScreenPickerSystem).tagged("cubos.screenpicker.init").after("cubos.window.init");
 
-    cubos.system(clearTexture).tagged("cubos.screenpicker.clear").before("cubos.renderer.draw");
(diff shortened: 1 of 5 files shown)


engine/src/cubos/engine/screenpicker/plugin.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void cubos::engine::screenPickerPlugin(Cubos& cubos)
+{
+    cubos.addPlugin(cubos::engine::windowPlugin);
+
+    cubos.addResource<ScreenPicker>();
+
+    cubos.startupSystem(initScreenPickerSystem).tagged("cubos.screenpicker.init").after("cubos.window.init");
+
+    cubos.system(clearTexture).tagged("cubos.screenpicker.clear").before("cubos.renderer.draw");
```
RiscadoA, line 43: Hm, I think I wouldn't specify `.before("cubos.renderer.draw")` here, since this plugin shouldn't be aware about the renderer at all. Instead, the renderer should specify something like `.after("cubos.screenpicker.clear")`

RiscadoA (approved):
LGTM! Thanks for working on this! When you fix what I pointed out feel free to merge.

PR #897, opened 2024-01-24
Integrate screenPickerPlugin with rendererPlugin

tomas7770:
# Description

Draws entity indexes to the screenPickerPlugin framebuffer, allowing the plugin to actually work (for entities).

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/renderer/renderer.hpp
+++ b/engine/include/cubos/engine/renderer/renderer.hpp
@@ -94,10 +94,11 @@ namespace cubos::engine
         /// @param camera Camera to use.
         /// @param frame Frame to draw.
         /// @param pickingBuffer Screen picking framebuffer.
+        /// @param enableScreenPicking Whether to draw to the screen picking buffer.
         /// @param usePostProcessing Whether to use post processing.
         /// @param target Target framebuffer to draw to.
         void render(const glm::mat4& view, const Viewport& viewport, const engine::Camera& camera,
-                    const RendererFrame& frame, const core::gl::Framebuffer& pickingBuffer,
+                    const RendererFrame& frame, const core::gl::Framebuffer& pickingBuffer, bool enableScreenPicking,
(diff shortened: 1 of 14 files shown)


engine/include/cubos/engine/renderer/renderer.hpp
```diff
@@ -93,11 +93,13 @@ namespace cubos::engine
...
         /// @param camera Camera to use.
         /// @param frame Frame to draw.
+        /// @param pickingBuffer Screen picking framebuffer.
+        /// @param enableScreenPicking Whether to draw to the screen picking buffer.
         /// @param usePostProcessing Whether to use post processing.
         /// @param target Target framebuffer to draw to.
         void render(const glm::mat4& view, const Viewport& viewport, const engine::Camera& camera,
-                    const RendererFrame& frame, bool usePostProcessing = true,
-                    const core::gl::Framebuffer& target = nullptr);
+                    const RendererFrame& frame, const core::gl::Framebuffer& pickingBuffer, bool enableScreenPicking,
```
RiscadoA, line 101: Instead of passing a boolean `enableScreenPicking`, you can set `pickingBuffer` to `nullptr` to disable it, as we do with `target`.

engine/src/cubos/engine/screenpicker/screenpicker.cpp
RiscadoA, line 1: Just now noticed that we're using `screenpicker` as a path name, but shouldn't we use `screen_picker`, for consistency? Usually we add a underscore between words.

RiscadoA (changes requested):
Thanks for working on this! Overall LGTM, good work!

engine/src/cubos/engine/screenpicker/screenpicker.cpp
tomas7770: What about tags and settings? E.g. `cubos.screenpicker.init`, `cubos.renderer.screenpicking.enabled`. Underscore too?

PR #911, opened 2024-01-29
Replace Broadphase resource with a relations solution

fallenatlas:
# Description

Replaces broad phase resource which kept possibly colliding pairs of entities with a relation between these entities.

## Checklist

- [x] Self-review changes.


engine/samples/collisions/main.cpp
```diff
@@ -103,18 +103,13 @@ static void updateTransform(State& state, const Input& input, Query<Position&, R
     bPos.vec += glm::vec3{0.0F, 0.0F, -0.01F};
 }
 
-static void updateCollided(Query<Entity, const Collider&> query, State& state, const BroadPhaseCandidates& candidates)
+static void updateCollided(Query<Entity, PotentiallyCollidingWith&, Entity> query, State& state)
 {
-    for (auto [entity, collider] : query)
+    for (auto [ent, colliding, ent1] : query)
```
RiscadoA, line 108:
```suggestion
    for (auto [ent1, colliding, ent2] : query)
```

PR #912, opened 2024-01-29
Fix cmake 3.5 version warnings

RiscadoA:
# Description

Updated both the glad and the json dependencies to their latest versions, should have fixed the problem.


RiscadoA:
When you're able to, please check if you still get the CMake warnings on this branch @joaomanita.

PR #937, opened 2024-02-01
Box to Box Collisions

fallenatlas:
# Description

Implement narrow phase collision detection for box vs box.

## Checklist

- [x] Self-review changes.
- [x] Update sample.


core/include/cubos/core/geom/box.hpp
```diff
@@ -52,6 +52,16 @@ namespace cubos::core::geom
             corners[6] = {-halfSize.x, halfSize.y, halfSize.z};
             corners[7] = {halfSize.x, -halfSize.y, -halfSize.z};
         }
+
+        void normals(glm::vec3 normals[6]) const
```
RiscadoA, line 56:
```suggestion
        static void normals(glm::vec3 normals[6]) const
```

Since this doesn't use the object's data, I think you can make it static.

core/include/cubos/core/geom/intersections.hpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        CUBOS_REFLECT;
+
+        bool collided;      ///< If there was a collision
+        float penetration;  ///< Penetration depth of the collision.
+        glm::vec3 position; ///< Position of contact on the surface of the entity.
+        glm::vec3 normal;   ///< Normal of contact on the surface of the entity.
+    };
+
+    CollisionInfo BoxBoxIntersection(const Box& box1, const glm::mat4& localToWorld1, const Box& box2,
```
RiscadoA, line 28:
```suggestion
    /// docs here...
    CollisionInfo intersects(const Box& box1, const glm::mat4& localToWorld1, const Box& box2,
```

No need to add BoxBox to the name: C++ allows us to have multiple functions with the same name as long as they receive different argument types.

core/include/cubos/core/geom/intersections.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <glm/glm.hpp>
+
+#include <cubos/core/geom/box.hpp>
+#include <cubos/core/reflection/reflect.hpp>
+
+namespace cubos::core::geom
+{
+    /// @brief Contains info regarding a collision.
+    /// @ingroup core-geom
+    struct CollisionInfo
```
RiscadoA, line 18: I would call this intersect info instead of CollisionInfo

core/include/cubos/core/geom/intersections.hpp
RiscadoA: Also, I think it would be best to receive a `CollisionInfo&` by reference, where data is written to if there is a collision. Then you could return a boolean in the function, indicatign if there was indeed a collision. Since the struct may get quite big, we avoid unnecessary moving around of data this way.

core/include/cubos/core/geom/intersections.hpp
```diff
@@ -0,0 +1,39 @@
...
+    CollisionInfo BoxBoxIntersection(const Box& box1, const glm::mat4& localToWorld1, const Box& box2,
+                                     const glm::mat4& localToWorld2);
+
+    std::vector<glm::vec3> getAxes(const Box& box, const glm::mat4& localToWorld);
+
+    glm::vec2 project(const Box& box, glm::vec3 axis, const glm::mat4& localToWorld);
+
+    bool overlap(glm::vec2 p1, glm::vec2 p2);
+
+    float getOverlap(glm::vec2 p1, glm::vec2 p2);
```
RiscadoA, line 37: Do those functions need to be in the header, or are they only used in the source file?
Does `getAxes` really need to be an std::vector? Using a fixed size array here, as you did for the normals, would be more efficient to avoid memory allocation.

engine/src/collisions/narrow_phase/colliding_with.cpp
```diff
@@ -0,0 +1,16 @@
+#include <cubos/core/ecs/reflection.hpp>
+#include <cubos/core/reflection/external/glm.hpp>
+#include <cubos/core/reflection/external/primitives.hpp>
```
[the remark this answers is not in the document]
fallenatlas: It gives me error if I don't include it, I think it's because of the float.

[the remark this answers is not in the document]
RiscadoA: Whoops, my bad, didn't notice you had a float.

RiscadoA (approved):

PR #949, opened 2024-02-05
Implement save for scenebridge

DiogoMendonc-a:
# Description

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


core/include/cubos/core/reflection/comparison.hpp
```diff
@@ -0,0 +1,6 @@
+#include <cubos/core/reflection/type.hpp>
```
RiscadoA, line 1: Missing docs (both header and func)

core/src/data/ser/json.cpp
```diff
@@ -94,6 +94,18 @@ bool JSONSerializer::decompose(const Type& type, const void* value)
 
     if (type.has<FieldsTrait>())
     {
+        if (type.get<FieldsTrait>().size() == 1)
+        {
+            // If there's a single field, write it directly.
+            if (!this->write(type.get<FieldsTrait>().begin()->type(),
+                             type.get<FieldsTrait>().view(value).begin()->value))
+            {
+                CUBOS_WARN("Couldn't deserialize wrapped field '{}'", type.get<FieldsTrait>().begin()->name());
```
RiscadoA, line 103: Serialize*

core/src/reflection/comparison.cpp
```diff
@@ -0,0 +1,92 @@
...
+        return true;
+    }
+
+    if (type.has<StringConversionTrait>())
+    {
+        const auto& trait = type.get<StringConversionTrait>();
+        return trait.into(a) == trait.into(b);
+    }
+
+    CUBOS_CRITICAL("Type cannot be compared!");
```
RiscadoA, line 90: Remove return and replace the critical by CUBOS_FAIL, which aborts the program

engine/src/scene/bridge.cpp
```diff
@@ -244,12 +207,168 @@ bool SceneBridge::loadFromFile(Assets& assets, const AnyAsset& handle, Stream& s
...
+            }
+            if (dirty)
+            {
+                entitiesJson.emplace(name.c_str(), entityJson);
+            }
+        }
+    }
+    json.push_back({"entities", entitiesJson});
+
+    stream.print(json.dump(2));
```
RiscadoA, line 372: You are dumping json formatted with 2 spaces indentation, but in this PR you formatted the JSON files to be tab-indented.  Make sure to revert the formatting on the json files so that we have something consistent.

RiscadoA (approved):
Thanks for working on this! Overall LGTM, will approve, but make sure to resolve the comments I wrote before merging. Also: regarding the blueprint realation serialization change (to use '@'), make sure that docs match the new format. IIRC i wrote some stuff explaining the previous format, not sure where (perhaps in the bridge docs or sample)

Issue #1003, opened 2024-02-24
Transform gizmo uses relative position instead of world position 


RiscadoA:
Blocked by #1002.

# Description
If you select a child entity, the transform gizmo will appear not in its position, but where it would be if its parent was at the origin. 

# Expected behaviour
It should use the world position.

PR #1031, opened 2024-02-26
1002 add world positionrotationscale getters to localtoworld

### LINKED ISSUE #1002: Add world position/rotation/scale getters to LocalToWorld
# Problem
Currently the only way to get the world position of an object is to extract the fourth column of the `LocalToWorld` matrix.

# Solution
We should provide getters to the `LocalToWorld` component which hide the necessary logic to calculate it.
Would also be cool to include rotation and scale.

DiogoMendonc-a:
closes #1002 and #1003

### DIFF
--- a/engine/include/cubos/engine/transform/local_to_world.hpp
+++ b/engine/include/cubos/engine/transform/local_to_world.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <glm/gtc/quaternion.hpp>
 #include <glm/mat4x4.hpp>
 
 #include <cubos/core/reflection/reflect.hpp>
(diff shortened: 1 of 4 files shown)


RiscadoA:
Don't forget to update the changelog! (you shouldn't erase the checklist :triumph:)

engine/include/cubos/engine/transform/local_to_world.hpp
```diff
@@ -24,6 +25,18 @@ namespace cubos::engine
...
+        /// @return Position vector in world space.
+        glm::vec3 worldPosition();
+
+        /// @brief Gets global rotation of the entity.
+        /// @return Rotation quaternion in world space.
+        glm::quat worldRotation();
+
+        /// @brief Gets global scale of the entity.
+        /// @return Scale value in world space.
+        float worldScale();
```
RiscadoA, line 39:
```suggestion
        glm::vec3 worldPosition() const;

        /// @brief Gets global rotation of the entity.
        /// @return Rotation quaternion in world space.
        glm::quat worldRotation() const;

        /// @brief Gets global scale of the entity.
        /// @return Scale value in world space.
        float worldScale() const;
```

RiscadoA (changes requested):
Also should add an entry to the changelog

RiscadoA (approved):

PR #1079, opened 2024-03-16
572 stop using serialization for grids and palettes

Dacops:


engine/include/cubos/engine/voxels/grid.hpp
```diff
@@ -8,22 +8,16 @@
 
 #include <glm/glm.hpp>
 
-#include <cubos/core/data/old/deserializer.hpp>
-#include <cubos/core/data/old/serializer.hpp>
 #include <cubos/core/reflection/reflect.hpp>
 
+#include <cubos/engine/assets/bridges/file.hpp>
```
RiscadoA, line 13: You don't need this anymore!
Since you use `cubos::core::memory::Stream&` down there, you can forward declare it instead of including it. Saves compiling time.

```cpp
namespace cubos::core::memory
{
    class Stream;
} // namespace cubos::core::memory

File: engine/include/cubos/engine/voxels/grid.hpp [hunk 2]
```diff
@@ -91,11 +85,10 @@ namespace cubos::engine
         /// @return Whether the conversion was successful.
         bool convert(const VoxelPalette& src, const VoxelPalette& dst, float minSimilarity);
 
-    private:
-        friend void core::data::old::serialize(core::data::old::Serializer& /*serializer*/, const VoxelGrid& /*grid*/,
-                                               const char* /*name*/);
-        friend void core::data::old::deserialize(core::data::old::Deserializer& /*deserializer*/, VoxelGrid& /*grid*/);
+        bool loadFrom(core::memory::Stream& stream);
+        bool writeTo(core::memory::Stream& stream) const;
```
RiscadoA, line 89: Missing docs. Explain briefly the format used to store/save them here.
I.e.

```suggestion
        /// @brief Loads the grid's data from the given stream.
        ///
        /// Assumes the data is stored in big-endian (network byte order).
        /// The first bytes correspond to three uint32_t, which represent the size of the grid (x, y, z).
        /// The next bytes correspond to `size.x * size.y * size.z` uint16_t, which represent the actual voxel materials.
        /// The voxel data is indexed by `x + y * size.x + z * size.x * size.y`.
        ///
        /// @param stream Stream to read from.
        /// @return Whether the stream contained valid data.
        bool loadFrom(core::memory::Stream& stream);
        
        /// @brief Writs the grid's data to the given stream.
        ///
        /// Writes in the format specified in @ref loadFrom.
        ///
        /// @param stream Stream to write to.
        /// @return Whether the write was successful.
        bool writeTo(core::memory::Stream& stream) const;
```

engine/include/cubos/engine/voxels/palette.hpp
```diff
@@ -104,12 +99,10 @@ namespace cubos::engine
         /// @return Iterator.
         Iterator end();
 
-    private:
-        friend void core::data::old::serialize(core::data::old::Serializer& /*serializer*/,
-                                               const VoxelPalette& /*palette*/, const char* /*name*/);
-        friend void core::data::old::deserialize(core::data::old::Deserializer& /*deserializer*/,
-                                                 VoxelPalette& /*palette*/);
+        bool loadFrom(core::memory::Stream& stream);
+        bool writeTo(core::memory::Stream& stream) const;
```
RiscadoA, line 103: Should also be documented, as I suggested doing for the VoxelGrid.

engine/include/cubos/engine/voxels/palette_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <cubos/engine/assets/bridges/file.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Palette assets.
+    ///
+    /// Palettes are composed by n materials.
+    /// They're stored as an array of floats (each 4 floats equal a material).
```
RiscadoA, line 14: Here you can just refer that it uses the same format as `@ref VoxelPalette::loadFrom and @ref VoxelPalette::writeTo`.

engine/include/cubos/engine/voxels/palette_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+    ///
+    /// @ingroup scene-plugin
+    class PaletteBridge : public FileBridge
```
RiscadoA, line 17: I think we should call this `VoxelPaletteBridge` for consistency.

engine/src/voxels/grid.cpp
```diff
@@ -152,27 +157,40 @@ bool VoxelGrid::convert(const VoxelPalette& src, const VoxelPalette& dst, float
...
+bool VoxelGrid::loadFrom(Stream& stream)
 {
-    serializer.beginObject(name);
-    serializer.write(grid.mSize, "size");
-    serializer.write(grid.mIndices, "data");
-    serializer.endObject();
+    uint32_t x, y, z;
+    stream.read(&x, sizeof(uint32_t));
+    stream.read(&y, sizeof(uint32_t));
+    stream.read(&z, sizeof(uint32_t));
```
RiscadoA, line 165: Check if return value of read is really `sizeof(uint32_t)`, and if it isn't, log an error (something like unexpected end of file) and return false.

engine/src/voxels/grid.cpp
```diff
@@ -152,27 +157,40 @@ bool VoxelGrid::convert(const VoxelPalette& src, const VoxelPalette& dst, float
...
+    x = fromBigEndian(x);
+    y = fromBigEndian(y);
+    z = fromBigEndian(z);
+    mSize = glm::uvec3(x, y, z);
+
+    while (!stream.eof())
+    {
+        uint16_t idx;
+        stream.read(&idx, sizeof(uint16_t));
```
RiscadoA, line 174: Ditto, check if `== sizeof(uint16_t)`

engine/src/voxels/grid.cpp
```diff
@@ -152,27 +157,40 @@ bool VoxelGrid::convert(const VoxelPalette& src, const VoxelPalette& dst, float
...
-    deserializer.read(grid.mIndices);
-    deserializer.endObject();
-
-    if (grid.mSize.x * grid.mSize.y * grid.mSize.z != static_cast<unsigned int>(grid.mIndices.size()))
+    uint32_t x = toBigEndian(mSize.x);
+    uint32_t y = toBigEndian(mSize.y);
+    uint32_t z = toBigEndian(mSize.z);
+    stream.write(&x, sizeof(uint32_t));
+    stream.write(&y, sizeof(uint32_t));
+    stream.write(&z, sizeof(uint32_t));
```
RiscadoA, line 188: Ditto, check if the written amount checks out.

engine/src/voxels/grid.cpp
```diff
@@ -152,27 +157,40 @@ bool VoxelGrid::convert(const VoxelPalette& src, const VoxelPalette& dst, float
...
+
+    for (const auto& indice : mIndices)
     {
-        CUBOS_WARN("Grid size and indices size mismatch: was {}, indices size is {}.", grid.mSize,
-                   grid.mIndices.size());
-        grid.mSize = {1, 1, 1};
-        grid.mIndices.clear();
-        grid.mIndices.resize(1, 0);
+        uint16_t idx = toBigEndian(indice);
+        stream.write(&idx, sizeof(uint16_t));
```
RiscadoA, line 193: Ditto

engine/include/cubos/engine/voxels/grid_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Grid assets.
+    ///
+    /// Grids are composed by their size and their indices.
+    /// They're stored as an array of unsigned integers (first 3 equal the grid size, the remainder the grid indexes)
+    ///
+    /// @ingroup scene-plugin
+    class GridBridge : public FileBridge
```
RiscadoA, line 17: Should also be `VoxelGridBridge` for consistency.

engine/include/cubos/engine/voxels/grid_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <cubos/engine/assets/bridges/file.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Grid assets.
+    ///
+    /// Grids are composed by their size and their indices.
+    /// They're stored as an array of unsigned integers (first 3 equal the grid size, the remainder the grid indexes)
+    ///
+    /// @ingroup scene-plugin
```
RiscadoA, line 16:
```suggestion
    /// @ingroup voxels-plugin
```

engine/include/cubos/engine/voxels/grid_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <cubos/engine/assets/bridges/file.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Grid assets.
+    ///
+    /// Grids are composed by their size and their indices.
+    /// They're stored as an array of unsigned integers (first 3 equal the grid size, the remainder the grid indexes)
```
RiscadoA, line 14: Ditto, same as mentioned for `VoxelPaletteBridge`

engine/include/cubos/engine/voxels/palette_bridge.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <cubos/engine/assets/bridges/file.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Palette assets.
+    ///
+    /// Palettes are composed by n materials.
+    /// They're stored as an array of floats (each 4 floats equal a material).
+    ///
+    /// @ingroup scene-plugin
```
RiscadoA, line 16:
```suggestion
    /// @ingroup voxels-plugin
```

engine/src/voxels/palette.cpp
```diff
@@ -151,53 +156,36 @@ VoxelPalette::Iterator VoxelPalette::end()
...
-        if (palette.mMaterials[i].similarity(VoxelMaterial::Empty) < 1.0F)
-        {
-            serializer.write<uint16_t>(i + 1, nullptr);
-            serializer.write(palette.mMaterials[i], nullptr);
-        }
+        float r, g, b, a;
+        stream.read(&r, sizeof(float));
+        stream.read(&g, sizeof(float));
+        stream.read(&b, sizeof(float));
+        stream.read(&a, sizeof(float));
```
RiscadoA, line 167: Ditto

engine/src/voxels/palette.cpp
```diff
@@ -151,53 +156,36 @@ VoxelPalette::Iterator VoxelPalette::end()
...
-    {
-        if (material.similarity(VoxelMaterial::Empty) < 1.0F)
-        {
-            count++;
-        }
-    }
-
-    serializer.beginDictionary(count, name);
-    for (uint16_t i = 0; i < static_cast<uint16_t>(palette.mMaterials.size()); i++)
+    while (!stream.eof())
```
RiscadoA, line 161: IMO, it's best to store an integer at the start of the file which indicates how many materials the palette will have, instead of expecting the whole stream to have just this palette.
In the future someone using this method may want to store multiple palettes in a single stream, and it should be possible.

RiscadoA (changes requested):
Thanks for working on this! So close to getting rid of the serialization :eyes:
Don't forget to add an entry to the CHANGELOG.md!

Dacops:
TODO: Figure out why engine-samples.voxels started blocking

RiscadoA (changes requested):
Almost ready to go! Also, don't use the commit scope `(serialization)`, use `(voxels)`. I.e. `feat(voxels)` blabla (thats the name of the label).

Another thing to take into account is that commits should be readable even without their PR - that info will be lost after we merge them. So, avoid stuff like 'resolve conversations' in the future (no need to change that now though, would be too much work).

One more thing: start commit descs with a lowercase character. E.g. 'fix(...): resolve' instead of 'fix(...): Resolve'

RiscadoA (approved):
LGTM!

PR #1083, opened 2024-03-18
Add GBuffer plugin

RiscadoA:
# Description

Adds the `gBufferPlugin`, which adds and manages the `GBuffer` component.
Also adds a placeholder `RenderTarget` component to make this implementable (still missing the actual plugin and logic).

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/src/render/g_buffer/plugin.cpp
+++ b/engine/src/render/g_buffer/plugin.cpp
@@ -1,0 +1,54 @@
+#include <cubos/core/io/window.hpp>
+#include <cubos/core/reflection/external/primitives.hpp>
+
+#include <cubos/engine/render/g_buffer/g_buffer.hpp>
+#include <cubos/engine/render/g_buffer/plugin.hpp>
+#include <cubos/engine/render/target/target.hpp>
+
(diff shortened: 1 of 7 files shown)


engine/src/render/g_buffer/plugin.cpp
```diff
@@ -0,0 +1,53 @@
...
+using cubos::core::gl::Texture2DDesc;
+using cubos::core::gl::TextureFormat;
+using cubos::core::gl::Usage;
+using cubos::core::io::Window;
+
+void cubos::engine::gBufferPlugin(Cubos& cubos)
+{
+    cubos.addComponent<GBuffer>();
+
+    cubos.tag("CreateGBufferTag");
```
tomas7770, line 17: Tag naming looks different from usual. Does this have anything to do with the upcoming typed tags?

RiscadoA: Yeah, kept it like that so that @joaomanita knows what it should be named like if we merge this first.

tomas7770 (approved):
LGTM!

PR #1104, opened 2024-03-21
Add Render Target plugin

tomas7770:
# Description

Adds a new `renderTargetPlugin` in the `cubos/engine/render` directory, which registers a new component named `RenderTarget`, and a system which automatically resizes render targets with a null framebuffer to the current window size.

Also changes window `framebufferSize()` and `size()` return types to unsigned values, because the window size shouldn't be negative.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


RiscadoA (approved):
LGTM, ty for working on this :)

PR #1113, opened 2024-03-31
Add `InheritsTrait`

roby2014:
# Description

Implemented `InheritsTest` which allows exposing the inheritance relationship between types.
## Checklist

- [X] Self-review changes.
- [X] Ensure test coverage.
- [X] Write new samples.


core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
+#include <string>
```
RiscadoA, line 1: Missing file level docs (see other files)

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <unordered_set>
```
RiscadoA, line 2: Unnecessary includes

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
...
+            return std::move(*this);
+        }
+
+        /// @brief Checks if the type inherits from the specified parent type.
+        /// @tparam T The parent type to check for inheritance.
+        /// @return Whether type inherits from the specified parent type, false otherwise.
+        template <typename T>
+        bool inherits() const
+        {
+            return mType->is<T>();
```
RiscadoA, line 35: We should also check if the inherit T indirectly.
```suggestion
            return mType->is<T>() || (mType->has<InheritsTrait>() && mType->get<InheritsTrait>().inherits<T>());
```

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
...
+        }
+
```
RiscadoA, line 37: Missing a getter `base()` to get `mType`.

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
...
+    /// @brief Provides inheritance relationship between types.
+    /// @see See @ref examples-core-reflection-traits-inherits for an example of using this trait.
+    /// @ingroup core-reflection
+    class InheritsTrait
+    {
+    public:
+        ~InheritsTrait() = default;
+
+        /// @brief Constructs.
+        InheritsTrait() = default;
```
RiscadoA, line 17: I would make the constructor receive a `const Type&`, as it never makes sense to add this trait without setting a type.

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+        /// @brief Adds an inheritance relationship to the type.
+        /// @tparam T The parent type from which this type inherits.
+        /// @return Trait.
+        template <typename T>
+        InheritsTrait&& from() &&
```
RiscadoA, line 23: I would change this to create a trait, instead of changing the existing one.
```suggestion
        static InheritsTrait from()
```

core/tests/reflection/traits/inherits.cpp
```diff
@@ -0,0 +1,38 @@
+#include <doctest/doctest.h>
+
+#include <cubos/core/reflection/traits/inherits.hpp>
+#include <cubos/core/reflection/type.hpp>
+
+using cubos::core::reflection::InheritsTrait;
+using cubos::core::reflection::reflect;
+using cubos::core::reflection::Type;
+
+struct Parent
```
RiscadoA, line 10: Add GrandParent type and check if son inherits from grand parent (it does).

core/tests/reflection/traits/inherits.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+CUBOS_REFLECT_IMPL(Son)
+{
+    return Type::create("Son").with(InheritsTrait().from<Parent>());
+}
+
+TEST_CASE("reflection::InheritsTrait")
+{
+    SUBCASE("")
+    {
```
RiscadoA, line 33: Since there's a single case, you can remove this.
```suggestion
```

RiscadoA (changes requested):
Thanks for working on this! :)

core/include/cubos/core/reflection/traits/inherits.hpp
```diff
@@ -0,0 +1,48 @@
...
+        /// @return Whether type inherits from the specified parent type, false otherwise.
+        template <typename T>
+        bool inherits() const
+        {
+            return mType->is<T>() || (mType->has<InheritsTrait>() && mType->get<InheritsTrait>().inherits<T>());
+        }
+
+        /// @brief Gets the base type of the inherited type.
+        /// @return Type.
+        const Type* base();
```
RiscadoA, line 43:
```suggestion
        const Type& base();
```
We usually always use references in the public facing methods as these types may never be null.

RiscadoA (approved):
LGTM! Just one tiny thing to change before merging :)

RiscadoA (approved):

PR #1115, opened 2024-04-04
Add Render Picker plugin

tomas7770:
# Description

Adds a `renderPickerPlugin` in the `cubos/engine/render` directory, which registers a component named `RenderPicker`, and a system which automatically resizes render picker textures when the corresponding render target is resized.

This should do the same as the existing `screenPickerPlugin` and `ScreenPicker`, but uses the new `RenderTarget`.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


RiscadoA (approved):
LGTM! Ty for working on this :)

PR #1129, opened 2024-04-22
Allow filtering by components and name on world inspector

### LINKED ISSUE #1105: Add search bar to World Inspector which allows filtering by components

diogomsmiranda:
# Description

closes #1105 & #1106.

## Checklist

### DIFF
--- a/tools/tesseratos/src/tesseratos/world_inspector/plugin.cpp
+++ b/tools/tesseratos/src/tesseratos/world_inspector/plugin.cpp
@@ -1,3 +1,7 @@
+#include <iostream>
+#include <sstream>
+#include <string>
+
 #include <imgui.h>
 
 #include <cubos/core/ecs/name.hpp>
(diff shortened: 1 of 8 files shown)


tools/tesseratos/src/tesseratos/world_inspector/plugin.cpp
```diff
@@ -20,8 +24,112 @@ using cubos::engine::Entity;
...
+static void parse(World& world, char* searchBuffer, Query<Entity, Opt<const Name&>> all,
+                  tesseratos::EntitySelector& selector)
+{
+    std::stringstream ss(searchBuffer);
+    std::string query;      // query string
+    char delimiter = ',';   // separator for queries
+    cubos::core::reflection::TypeRegistry worldComponents = world.types().components();
+    std::vector<const cubos::core::reflection::Type*> components;
+    std::string nameQuery;
+    bool limit = false;     // limit the filtering by name query to 1 name
```
RiscadoA, line 97: Unnecessary variable, you can just check if `nameQuery` is empty.

tools/tesseratos/src/tesseratos/world_inspector/plugin.cpp
```diff
@@ -20,8 +24,112 @@ using cubos::engine::Entity;
...
+
+    for (auto [entity, name] : all)
+    {
+        for (auto component : components)
+        {
+            if (world.components(entity).has(*component))
+            {
+                hasComponent = true;
+                break;
+            }
```
RiscadoA, line 67: This means that the entity only needs one of the requested components to appear in the result, which is wrong.
It should only appear if all of the components are present in it.
What you want is `if (!has) { hasComponent = false; break; }`, and by default, `hasComponent = true;`

RiscadoA (changes requested):
Thanks for working on this!

RiscadoA (changes requested):

tools/tesseratos/src/tesseratos/world_inspector/plugin.cpp
```diff
@@ -20,8 +24,107 @@ using cubos::engine::Entity;
...
+    auto worldComponents = world.types().components();
+    std::vector<const cubos::core::reflection::Type*> components;
+    std::string nameQuery;
+
+    while (getline(ss, query, delimiter))
+    {
+        if (isValid(query)) // TODO: ADD future checks for queries here
+        {
+            query.erase(std::remove(query.begin(), query.end(), ' '),
+                        query.end()); // remove all white-spaces from the query
```
RiscadoA, line 100: Why the query.erase? The std::remove is enough.

RiscadoA (changes requested):

RiscadoA (approved):
LGTM!

PR #1153, opened 2024-05-01
Implement metrics

roby2014:
# Description

**Check the sample for usage**

TLDR: The goal is to have a simple but extendible core class to register, search and debug metrics, with a compiler option to enable it

## Checklist

- [X] Self-review changes.
- [x] Ensure test coverage.
- [X] Write new samples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reset method to the Cubos class (#1225, **@RiscadoA**).
 - Plugin injection to the Cubos class (#1214, **@RiscadoA**).
 - Utility class used to load plugins from shared libraries (#1035, **@RiscadoA**).
+- Implemented metrics and profiling tools (#1150, **@roby2014**).
 
 ## [v0.2.0] - 2024-05-07
 
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -17,6 +17,9 @@ option(BUILD_CORE_SAMPLES "Build cubos core samples" OFF)
 option(BUILD_CORE_TESTS "Build cubos core tests?" OFF)
 option(BUILD_CORE_SHARED "Build cubos core as shared library?" ON)
 
+option(ENABLE_PROFILING "Should profiling macros be compiled?" ON)
+set(CUBOS_METRIC_MAX_ENTRIES "1024" CACHE STRING "Maximum number of entries each metric can store")
+
 message("# Building core samples: " ${BUILD_CORE_SAMPLES})
 message("# Building core tests: " ${BUILD_CORE_TESTS})
 message("# Building core as shared library: " ${BUILD_CORE_SHARED})
@@ -25,6 +28,7 @@ message("# Building core as shared library: " ${BUILD_CORE_SHARED})
 set(CUBOS_CORE_SOURCE
 	"src/api.cpp"
 	"src/log.cpp"
+	"src/metrics.cpp"
 
 	"src/thread/pool.cpp"
 	"src/thread/task.cpp"
@@ -214,4 +218,12 @@ endif()
 # Add core samples
 if(BUILD_CORE_SAMPLES)
 	add_subdirectory(samples)
+endif()
+
+if(ENABLE_PROFILING)
+	target_compile_definitions(cubos-core
+		PUBLIC 
+		-DCUBOS_PROFILING=${ENABLE_PROFILING}
+		-DCUBOS_METRIC_MAX_ENTRIES=${CUBOS_METRIC_MAX_ENTRIES}
+	)
 endif()
--- a/core/include/cubos/core/metrics.hpp
+++ b/core/include/cubos/core/metrics.hpp
@@ -1,0 +1,105 @@
+/// @file
+/// @brief Metrics and profiling utilities and macros.
+/// @ingroup core
+
+#pragma once
+
+#include <chrono>
+#include <string>
+
+#include <cubos/core/api.hpp>
+#include <cubos/core/reflection/reflect.hpp>
+
+namespace cubos::core
+{
+    /// @brief Singleton class that manages a collection of metrics.
+    /// @ingroup core
+    class CUBOS_CORE_API Metrics
+    {
+    public:
+        /// @brief Deleted constructor.
+        Metrics() = delete;
+
+        /// @brief Add a metric in the pool.
+        /// @param metric The metric to add.
+        static void metric(const std::string& name, double metric);
+
+        /// @brief Size of the metric pool.
+        /// @return Metrics count.
+        static std::size_t size();
+
+        /// @brief Size of metrics associated with a given name.
+        /// @param name Metric name.
+        /// @return Metrics count.
+        static std::size_t sizeByName(const std::string& name);
+
+        /// @brief Clears the metric pool
+        static void clear();
+
+        /// @brief Sets maximum metric entries.
+        /// @param n The new maximum value.
+        static void setMaxEntries(std::size_t n);
+
+        /// @brief Read a metric value by its name.
+        /// @param name Metric name.
+        /// @param[out] value Buffer to store metric value.
+        /// @param offset Read count per metric.
+        /// @return Whether the metric was found.
+        static bool readValue(const std::string& name, double& value, size_t& offset);
+
+        /// @brief Search for a new unique metric name.
+        /// @param[out] name Buffer to store metric name.
+        /// @param seenCount Seen metrics count.
+        /// @return Whether a new metric was found.
+        static bool readName(std::string& name, size_t& seenCount);
+    };
+
+    /// @brief Utility to measure the execution time of a scope.
+    ///
+    /// This class measures the time between its construction and destruction,
+    /// providing a way to profile code execution time.
+    class CUBOS_CORE_API ScopeProfiler
+    {
+    public:
+        /// @brief Construct a ScopeProfiler with a given name.
+        ScopeProfiler(std::string name, std::string file, std::size_t line);
+
+        /// @brief Destruct the ScopeProfiler and calculate the elapsed time.
+        ~ScopeProfiler();
+
+    private:
+        std::string mName;                                                  ///< The name of the scope.
+        std::string mFile;                                                  ///< The file from where it was invoked.
+        std::size_t mLine;                                                  ///< The line from where it was invoked.
+        std::chrono::time_point<std::chrono::high_resolution_clock> mStart; ///< Start time when constructed.
+    };
+} // namespace cubos::core
+
+/// @def CUBOS_METRIC
+/// @ingroup core
+/// @brief Macro to add a metric.
+/// @param name The name of the metric.
+/// @param val The value to set for the metric.
+
+#ifdef CUBOS_PROFILING
+#define CUBOS_METRIC(name, val) ::cubos::core::Metrics::metric(name, val)
+#else
+#define CUBOS_METRIC(...)                                                                                              \
+    do                                                                                                                 \
+    {                                                                                                                  \
+    } while (false)
+#endif
+
+/// @def CUBOS_PROFILE
+/// @ingroup core
+/// @brief Macro to measure the execution time of a function/scope, registering it as a metric.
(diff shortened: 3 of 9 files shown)


RiscadoA:
Since this PR is still blocked on the other one, I'm marking it as a draft. Does it really depend on the other PR though? They seem mostly unrelated.

roby2014:
It does not depend anymore yeah, you're right.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - UI canvas, elements, and color rects. (#1116, **@DiogoMendonc-a**)
+- Implemented file-based logging (#1154, **@roby2014**).
```
RiscadoA, line 13: This shouldn't be here, right?

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+        std::string mName; ///< The name of the metric.
+        double mValue;     ///< The value of the metric (as a double).
+        std::string mFile; ///< The source file where the metric was created.
+        int mLine;         ///< The line number where the metric was created.
+    };
+
+    /// @brief Singleton class that manages a collection of metrics.
+    /// @ingroup core
+    class Metrics
```
RiscadoA, line 25: Missing CUBOS_CORE_API

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,79 @@
...
+        /// @param seenCount Seen metrics count.
+        /// @return Whether a new metric was found.
+        static bool readName(std::string& name, size_t& seenCount);
+    };
+
+    /// @brief Utility to measure the execution time of a scope.
+    ///
+    /// This class measures the time between its construction and destruction,
+    /// providing a way to profile code execution time.
+    class TimeTracker
```
RiscadoA, line 57: ScopeProfiler perhaps would be a better name? TimeTracker seems unrelated to the metrics system

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    private:
+        std::string mName;                                 ///< The name of the TimeTracker.
+        std::chrono::_V2::system_clock::time_point mStart; ///< Start time when constructed.
+    };
+} // namespace cubos::core
+
+/// @brief Macro to add a metric.
+/// @param name The name of the metric.
+/// @param val The value to set for the metric.
```
RiscadoA, line 74: Missing @ ingroup core

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,79 @@
...
+#define CUBOS_METRIC(name, val) ::cubos::core::Metrics::metric({name, val, __FILE__, __LINE__})
+
+/// @brief Macro to measure the execution time of a function/scope, registering it as a metric.
+/// @note The macro uses `__FUNCTION__` to register the metric name.
```
RiscadoA, line 78: Missing in group

core/samples/metrics/main.cpp
```diff
@@ -0,0 +1,75 @@
+/// [Include metrics]
+#include <cubos/core/metrics.hpp>
+/// [Include metrics]
+#include <chrono>
+#include <thread>
```
RiscadoA, line 5: I dont think you need those includes

core/src/metrics.cpp
```diff
@@ -0,0 +1,97 @@
...
+size_t Metrics::size()
+{
+    return state().mMetrics.size();
+}
+
+bool Metrics::readValue(const std::string& name, double& value, size_t& offset)
+{
+    std::lock_guard<std::mutex> lock(state().mMutex);
+
+    std::vector<Metric> filteredMetricsByName;
```
RiscadoA, line 47: Why not store a map from metric names to vectors in the singleton?
This way you wouldnt need this expensive operation here

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,79 @@
...
+#define CUBOS_PROFILE() ::cubos::core::TimeTracker _profilingTimeTracker(__FUNCTION__)
```
RiscadoA, line 79: Probably a good idea to also add the file and line of code.

RiscadoA (changes requested):
Thanks for working on this :)

core/samples/metrics/main.cpp
roby2014: I do: `std::this_thread::sleep_for(std::chrono::seconds(2));`

roby2014: at least, the `thread` one, `chrono` can be removed

RiscadoA: Ah right, mb

core/samples/metrics/main.cpp
```diff
@@ -0,0 +1,86 @@
...
+    CUBOS_ASSERT(offset == 1);
+
+    // etc..
+    /// [Getting values]
+
+    /// [Profiling metrics 2]
+    myFunction();
+
+    offset = 0;
+    CUBOS_ASSERT(Metrics::readValue("myFunction", value, offset));
```
RiscadoA, line 72: Won't this fail now, due to __FILE__ + __LINE__? Maybe you can get the name through the `readName` function.

core/src/metrics.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::lock_guard<std::mutex> lock(state().mMutex);
+
+    if (seenCount >= state().mMetrics.size())
+    {
+        return false;
+    }
+
+    // std::map does not support get by index
+    // so we manually loop it like this...
+    for (std::size_t idx = 0; const auto& n : state().mMetrics)
```
RiscadoA, line 126: This is a problem, as the map doesn't have a fixed order. It sorts its keys by lexicographical roder.
Thus, you could end up missing new names and getting old ones repeated when calling this function.
Instead of doing this, keep a separate `vector` for the known metric names.
Each time you add a metric in `::metric` with a name which isn't on the map, you push it to the vector.
Then, here you just access that vector.

RiscadoA (changes requested):

core/src/metrics.cpp
```diff
@@ -0,0 +1,163 @@
+#include <deque>
+#include <map>
+#include <utility>
+
+#include <cubos/core/log.hpp>
+#include <cubos/core/metrics.hpp>
+#include <cubos/core/reflection/external/primitives.hpp>
```
RiscadoA, line 7:
```suggestion
#include <cubos/core/reflection/external/primitives.hpp>
#include <cubos/core/reflection/external/string.hpp>
```

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace cubos::core
+{
+    /// @brief Represents a metric.
+    struct Metric
+    {
+        CUBOS_REFLECT;
+        std::string mName; ///< The name of the metric.
+        double mValue;     ///< The value of the metric (as a double).
+        std::string mFile; ///< The source file where the metric was created.
+        std::size_t mLine; ///< The line number where the metric was created.
```
RiscadoA, line 24: I don't think this matters for generic metrics. Just the name should be enough.
On the profiler side, you can add this information to the name of the metric - this is necessary as to not have conflicts between functions with the same name.

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include <cubos/core/api.hpp>
+#include <cubos/core/reflection/reflect.hpp>
+
+namespace cubos::core
+{
+    /// @brief Represents a metric.
+    struct Metric
+    {
+        CUBOS_REFLECT;
+        std::string mName; ///< The name of the metric.
```
RiscadoA, line 21: At this point you don't need to store this anymore, as the values are stored in a map with names as keys anyways.
Thus, you can just keep a `std::deque` of doubles, which should be faster. Which means that, maybe there's no need at all to have this type.

RiscadoA (changes requested):
Sorry for being so nitpicky with these reviews! Since this is something we'll start using everywhere I really want to get the details right. Ty for the all the work :pray:

core/include/cubos/core/metrics.hpp
roby2014: Hmmm, like `myFunction:54:main.cpp` ? The problem will be reading the metric...

RiscadoA: You would never want to do that though, you would find its name through `readName`.
Btw, we need a way to distinguish profiling metrics from other metrics, maybe we should prefix their name?

roby2014: `prof__myFunction` and `metric__a` looks good?

roby2014: IDK, thinking about it later on the UI, not sure how it should look.

RiscadoA: On the UI we would probably filter out those and give them special treatment. Maybe we should some special character which doesn't appear in function names? For example, `profiling:<file>:<function name:/line number`

roby2014: so for normal metrics, no prefix, correct?

RiscadoA (approved):
LGTM! :ship:

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,107 @@
+/// @file
+/// @brief Metrics and profiling utilities and macros.
+/// @ingroup core
+
+#pragma once
+
+#include <algorithm>
```
RiscadoA, line 7:
```suggestion
```

Just noticed this, do not include this in headers! It eats up compile times because it includes pretty much the whole STL

core/include/cubos/core/metrics.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include <chrono>
+#include <mutex>
```
RiscadoA, line 9:
```suggestion
```
You only use this in the source file, move it there

roby2014:
Eh no idea why static analysis is failing

RiscadoA:
> Eh no idea why static analysis is failing

It's an error in the action. It's trying to dismiss a review which no longer exists and crashes.
TL;DR: not a problem in your code. @joaomanita @diogomsmiranda pls review when you have the chance

core/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ option(BUILD_CORE_SAMPLES "Build cubos core samples" OFF)
 option(BUILD_CORE_TESTS "Build cubos core tests?" OFF)
 option(BUILD_CORE_SHARED "Build cubos core as shared library?" ON)
 
+option(ENABLE_PROFILING "Should profiling macros be compiled?" ON)
+set(CUBOS_METRIC_MAX_ENTRIES "1024" CACHE STRING "Maximum number of entries each metric can store")
```
diogomsmiranda, line 21: why was 1024 chosen as the max number of entries? Is this arbitrary?

diogomsmiranda (approved):

joaomanita (approved):

RiscadoA:
> I also didn't understand the need to make such a sample basically copying from the test. Shouldn't we make something more in tune with what would be a real example?

Agree, we should avoid this, could you improve it @roby2014?

core/CMakeLists.txt
roby2014: Just a dummy value. It can be changed easily through the method `setMaxEntries`

diogomsmiranda:
I would rather have the metrics to be explicitly generated with different names and contexts. But if @RiscadoA is ok with a sample like this It's not a deal breaker for me either.

roby2014:
> I would rather have the metrics to be explicitly generated with different names and contexts. But if @RiscadoA is ok with a sample like this It's not a deal breaker for me either.

Can you give me an example? What should I sample?

diogomsmiranda:
> I also didn't understand the need to make such a sample basically copying from the test. Shouldn't we make something more in tune with what would be a real example?

@roby2014 thank you for addressing this! LGTM

RiscadoA:
You can ignore the failing lint check btw

PR #1155, opened 2024-05-04
Add split screen to new renderer

tomas7770:
# Description

Ports the splitscreen plugin to the new renderer. For each `DrawsTo` relation with a target that has a `SplitScreen` component, the plugin sets the viewport in the relation. Plugins which draw to the target should use this viewport.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - GBuffer Rasterizer plugin (#1062, **@RiscadoA**).
 - Deferred Shading plugin (#1086, **@RiscadoA**).
 - SSAO plugin (#1088, **@RiscadoA**).
+- Split screen for the new renderer (#1149, **@tomas7770**).
 
 ### Changed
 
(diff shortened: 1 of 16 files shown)


engine/include/cubos/engine/render/splitscreen/plugin.hpp
RiscadoA, line 1: Considering the plugin is now called `splitScreen` we should rename the files to `split_screen` for consistency.

engine/src/render/splitscreen/plugin.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    int cameraCount = 0;
+
+                    // Find the cameras that draw to the target.
+                    for (auto [localToWorld, camera, drawsTo, splitScreen] : perspectiveCameras.pin(1, targetEnt))
+                    {
+                        // Ignore unused argument warnings
+                        (void)splitScreen;
+
+                        cameraCount += 1;
+                    }
```
RiscadoA, line 79: You should skip non-active cameras here and on the loop below.

RiscadoA (approved):
Overall LGTM! Ty very much for working on this :relieved: just two things to change before merging:
- The split screen plugin shouldn't reserve screen space for inactive cameras.
- The folder and files should be renamed to `split_screen` for consistency with the `splitScreen` name.

PR #1156, opened 2024-05-04
Implement file logging

roby2014:
# Description

The implementation is: when logging (with cubos_debug, etc..), if the file was set, it also prints to the file stream.
this means: data logged before setting the file, wont be logged

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.

### DIFF
--- a/core/include/cubos/core/log.hpp
+++ b/core/include/cubos/core/log.hpp
@@ -319,6 +319,18 @@ namespace cubos::core
         /// @return Log level.
         static Level level();
 
+        /// @brief Sets a file path where logs will be saved.
+        /// @note Previous logs are also dumped to the file.
+        /// @param filePath Path to file in the virtual file system.
+        /// @return Whether the file could be opened for logging.
+        static bool logToFile(const std::string& filePath);
+
+        /// @brief Mounts a standard archive on the `/logs/` directory, and calls `logToFile` to a timestamped log file
+        /// on that directory.
+        /// @note Previous logs are also dumped to the file.
+        /// @return Whether the file could be mounted and opened for logging.
+        static bool logToFile();
+
         /// @brief Creates a new entry in the logs, as long as the log level is high enough.
         /// @param level Log level.
         /// @param location Code location.
--- a/core/samples/logging/page.md
+++ b/core/samples/logging/page.md
@@ -19,6 +19,53 @@ this sample, to demonstrate @ref CUBOS_TRACE, we set it to @ref cubos::core::Log
 
 @snippet logging/main.cpp Set logging level
 
+The logger also supports logging to a file, which will by default be named to `cubos_<timestamp>.log`.
+
+@snippet logging/main.cpp Log to file
+
+@note The logger will by default store logs inside `/logs`.
+
+Or, if you want, you can set the file manually:
+
+@snippet logging/main.cpp Set logging file
+
+@note The logger will **always** dump previous logs to the file.
+
+```sh
+$ cat /home/cubos/build/logs/cubos_15:25:30.191.log
+[15:25:30.192] [file.cpp:124 mount] info: Mounted archive at ""/"logs"
+[15:25:30.192] [file.cpp:322 create] trace: Created "file" "/logs/cubos_15:25:30.191"
+[15:25:30.192] [main.cpp:29 main] trace: Trace message
+[15:25:30.192] [main.cpp:30 main] debug: Debug message
+[15:25:30.192] [main.cpp:31 main] info: Info message
+[15:25:30.192] [main.cpp:32 main] warn: Warning message
+[15:25:30.193] [main.cpp:33 main] error: Error message
+[15:25:30.193] [main.cpp:34 main] critical: Critical message
+[15:25:30.193] [file.cpp:85 mount] error: Could not mount archive at "/logs"/"": "/logs" is already part of an archive
+[15:25:30.193] [main.cpp:42 main] info: An integer: 1
+[15:25:30.193] [main.cpp:43 main] info: A glm::vec3: (x: 0.0000, y: 1.0000, z: 2.0000)
+[15:25:30.194] [main.cpp:44 main] info: An std::unordered_map: {2: "two", 1: "one"}
+[15:25:30.194] [log.cpp:248 streamFormat] warn: You tried to print a type ("unnamed102322537939356") which doesn't implement reflection. Did you forget to include its reflection definition?
+[15:25:30.194] [main.cpp:45 main] info: A type without reflection: (no reflection)
+```
+
+```sh
+$ cat /home/cubos/build/logs/sample_logs.txt 
+[15:23:03.737] [file.cpp:124 mount] info: Mounted archive at ""/"logs"
+[15:23:03.737] [main.cpp:29 main] trace: Trace message
+[15:23:03.737] [main.cpp:30 main] debug: Debug message
+[15:23:03.737] [main.cpp:31 main] info: Info message
+[15:23:03.737] [main.cpp:32 main] warn: Warning message
+[15:23:03.737] [main.cpp:33 main] error: Error message
+[15:23:03.738] [main.cpp:34 main] critical: Critical message
+[15:23:03.738] [main.cpp:38 main] info: An integer: 1
+[15:23:03.738] [main.cpp:39 main] info: A glm::vec3: (x: 0.0000, y: 1.0000, z: 2.0000)
+[15:23:03.738] [main.cpp:40 main] info: An std::unordered_map: {2: "two", 1: "one"}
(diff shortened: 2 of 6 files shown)


core/include/cubos/core/log.hpp
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstdlib>
+#include <memory>
```
RiscadoA, line 8: Avoid including this in the header file - these STL headers are really heavy on the compile times, this logging header particularly is included pretty much everywhere should it be as lean as possible.

core/include/cubos/core/log.hpp
```diff
@@ -319,6 +320,12 @@ namespace cubos::core
         /// @return Log level.
         static Level level();
 
+        /// @brief Sets a file path where logs will be saved.
+        /// @note Logs before setting the file will be dumped!
+        /// @param filePath Path to file (inside `/logs`)
```
RiscadoA, line 325:
```suggestion
        /// @param filePath Path to file in the virtual file system.
```

core/include/cubos/core/log.hpp
```diff
@@ -319,6 +320,12 @@ namespace cubos::core
         /// @return Log level.
         static Level level();
 
+        /// @brief Sets a file path where logs will be saved.
+        /// @note Logs before setting the file will be dumped!
```
RiscadoA, line 324:
```suggestion
        /// @note Previous logs are also dumped to the file.
```

core/include/cubos/core/log.hpp
```diff
@@ -319,6 +320,12 @@ namespace cubos::core
...
+        /// @param filePath Path to file (inside `/logs`)
+        /// @return Whether setting the file occurred successfully.
```
RiscadoA, line 326:
```suggestion
        /// @return Whether the file could be opened for writing.
```

core/include/cubos/core/log.hpp
```diff
@@ -353,6 +360,8 @@ namespace cubos::core
         static bool read(std::size_t& cursor, Entry& entry);
 
     private:
+        std::unique_ptr<memory::Stream> mLogFileStream;
```
RiscadoA, line 363: This is a singleton, you can/should just use the global static `State` structure defined in the `log.cpp` file to store this stream.

core/src/log.cpp
```diff
@@ -139,6 +147,33 @@ auto Logger::level() -> Logger::Level
...
 
+bool Logger::setFilePath(const std::string& filePath)
+{
+    if (!FileSystem::mount("/logs", std::make_unique<cubos::core::data::StandardArchive>("./logs", /*isDirectory=*/true,
+                                                                                         /*readOnly=*/false)))
+    {
+        return false;
+    }
+
+    auto fullPath = "/logs/" + filePath;
```
RiscadoA, line 158: I prefer to would leave this responsibility to allow more flexibility regarding the logs locations. Maybe rename this to `logToFile`, and provide another overload method `logToFile` which takes no arguments. That file could perform the base archive setup, choose a filename with the current timestamp (date and time), and then call the other `logToFile` overload.

You should also then add a call to `Logger::logToFile()` on the `Cubos` constructor.

What do you think?

core/src/log.cpp
```diff
@@ -139,6 +147,33 @@ auto Logger::level() -> Logger::Level
...
+    auto file = FileSystem::create(fullPath, /*directory=*/false);
+    if (file == nullptr)
+    {
+        return false;
+    }
```
RiscadoA, line 163: Here, you should only create the file if it doesn't already exist.

core/src/log.cpp
```diff
@@ -139,6 +147,33 @@ auto Logger::level() -> Logger::Level
...
+
+    state().logFileStream = file->open(File::OpenMode::ReadWrite);
```
RiscadoA, line 165:
```suggestion
    state().logFileStream = file->open(File::OpenMode::Write);
```

RiscadoA (changes requested):
Thanks very much for working on this! Left some comments :relieved:

core/src/log.cpp
roby2014: Does this mean we would need to have 2 different streams? What if both functions are called?

core/src/log.cpp
roby2014: Yeah, but `create` returns the file if alr exists (https://gamedevtecnico.github.io/cubos/docs/classcubos_1_1core_1_1data_1_1FileSystem.html#a4a5c6b0cb8c33de83caa6cdaf4147f87). I thought this way would be cleaner instead of chaining
```
get file
if file does not exist
   create
       if create fails
            return false
```

core/src/log.cpp
RiscadoA: No, one function is just a short-hand for calling the other. I think you should just `ASSERT` that a stream isn't already open - it doesn't make sense to call it twice.

core/src/log.cpp
RiscadoA: Whoops, forgot about that property, mb.

core/include/cubos/core/log.hpp
```diff
@@ -319,6 +319,17 @@ namespace cubos::core
         /// @return Log level.
         static Level level();
 
+        /// @brief Sets a file path where logs will be saved.
+        /// @note Previous logs are also dumped to the file.
+        /// @param filePath Path to file in the virtual file system.
+        /// @return Whether the file could be opened for logging.
+        static bool logToFile(const std::string& filePath); // print to filePath?=
```
RiscadoA, line 326:
```suggestion
        static bool logToFile(const std::string& filePath);
```
?

core/include/cubos/core/log.hpp
```diff
@@ -319,6 +319,17 @@ namespace cubos::core
...
+
+        /// @brief Sets a file based on current timestamp where logs will be saved.
```
RiscadoA, line 328:
```suggestion
        /// @brief Mounts a standard archive on the `/logs/` directory, and calls `logToFile` to a timestamped log file on that directory.
```

core/src/log.cpp
```diff
@@ -139,6 +147,40 @@ auto Logger::level() -> Logger::Level
...
 
+bool Logger::logToFile(const std::string& filePath)
+{
+    if (!FileSystem::mount("/logs", std::make_unique<cubos::core::data::StandardArchive>("./logs", /*isDirectory=*/true,
+                                                                                         /*readOnly=*/false)))
+    {
+        return false;
+    }
+
+    auto fullPath = "/logs/" + filePath;
```
RiscadoA, line 158: What I was saying is that we should allow users to choose any file to log to with this function.
Thus, you shouldn't mount the archive here, nor prefix the path with `/logs/`. This code should be moved to the other `logToFile` overload.

core/src/log.cpp
```diff
@@ -139,6 +147,40 @@ auto Logger::level() -> Logger::Level
...
+                                      toLower(EnumTrait::toString(e.level)), e.message);
+    }
+
+    return true;
+}
+
+bool Logger::logToFile()
+{
+    auto timestamp = Timestamp::now();
+    auto filePath = "cubos_" + timestamp.string();
```
RiscadoA, line 180:
```suggestion
    auto filePath = "cubos_" + timestamp.string() + ".log";
```

RiscadoA (changes requested):
Almost good to go, ty for working on this!

RiscadoA (approved):

core/samples/logging/page.md
```diff
@@ -19,6 +19,53 @@ this sample, to demonstrate @ref CUBOS_TRACE, we set it to @ref cubos::core::Log
 
 @snippet logging/main.cpp Set logging level
 
+The logger also supports logging to a file, which will be named to `cubos_<timestamp>.log`.
```
DiogoMendonc-a, line 22:
```suggestion
The logger also supports logging to a file, which will by default be named `cubos_<timestamp>.log`.
```

core/samples/logging/page.md
```diff
@@ -19,6 +19,53 @@ this sample, to demonstrate @ref CUBOS_TRACE, we set it to @ref cubos::core::Log
...
+
+@snippet logging/main.cpp Log to file
+
+@note The logger will store logs inside `/logs`.
```
DiogoMendonc-a, line 26:
```suggestion
@note The logger will by default store logs inside `/logs`.
```

DiogoMendonc-a (approved):
Looks good, just have minor nitpicky suggestions

fallenatlas (approved):
LGTM!

RiscadoA:
You can merge this btw @roby2014

PR #1189, opened 2024-05-23
feat(ui): draw ui through commands

DiogoMendonc-a:
# Description

The two main UI components are UIElement and UICanvas.
For it to work properly, each UIElement must be a child of a UICanvas, or of another UIElement.

Each UIElement generates a list of drawing commands, UIDrawList, that are then pooled together by the UICanvas at the root of the hierarchy.

Each entry in a draw list contains a UIDrawList::Type and a UIDrawList::Command. The Type holds information that is common for all similar commands, and the Command contains the specifics to that one command.
Each entry also has a raw data buffer to hold the data the command needs to pass to the shaders.

After aggregating the draw commands, the UICanvas draws them, grouped by type.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.
- [x] Add entry to the changelog's unreleased section. **No need, this is altering stuff that will be added this version**

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -87,6 +87,7 @@ set(CUBOS_ENGINE_SOURCE
 
 	"src/ui/canvas/plugin.cpp"
 	"src/ui/canvas/canvas.cpp"
+	"src/ui/canvas/draw_list.cpp"
 	"src/ui/canvas/element.cpp"
 	"src/ui/canvas/vertical_stretch.cpp"
 	"src/ui/canvas/horizontal_stretch.cpp"
--- a/engine/assets/ui/color_rect.fs
+++ b/engine/assets/ui/color_rect.fs
@@ -1,6 +1,13 @@
 #version 330 core
 
-uniform vec4 color;
+layout(std140) uniform PerElement
+{
+    vec2 xRange;
+    vec2 yRange;
+    vec4 color;
+    int depth;
+};
+
 out vec4 out_color;
 
 void main()
--- a/engine/assets/ui/element.vs
+++ b/engine/assets/ui/element.vs
@@ -2,10 +2,11 @@
 
 in vec3 position;
 
-uniform PerElement
+layout(std140) uniform PerElement
 {
     vec2 xRange;
     vec2 yRange;
+    vec4 color;
     int depth;
 };
 
--- a/engine/include/cubos/engine/ui/canvas/draw_list.hpp
+++ b/engine/include/cubos/engine/ui/canvas/draw_list.hpp
@@ -1,0 +1,142 @@
+/// @file
+/// @brief Class @ref cubos::engine::UIDrawList.
+/// @ingroup ui-canvas-plugin
+
+#pragma once
+#include <vector>
+
+#include <cubos/core/gl/render_device.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Holds a collections of draw commands and their data.
+    /// @note See the @ref ui-canvas-plugin documentation page for more information.
+    /// @ingroup ui-canvas-plugin
+    class CUBOS_ENGINE_API UIDrawList final
+    {
+    public:
+        /// @brief Describes a UI draw call type.
+        /// @ingroup ui-canvas-plugin
+        struct CUBOS_ENGINE_API Type
+        {
+            static constexpr int MaxTextures = 4;
+
+            cubos::core::gl::ShaderPipeline pipeline; ///< Handle to shader pipeline.
+            cubos::core::gl::ConstantBuffer
+                constantBuffer; ///< Handle to the constant buffer to hold the per element data.
+            cubos::core::gl::ShaderBindingPoint constantBufferBindingPoint; ///< Binding point of the element data.
+            size_t perElementSize; ///< Size of the struct holding the element data.
+            cubos::core::gl::ShaderBindingPoint texBindingPoint[MaxTextures]; ///< Textures binding points.
+
+            /// @brief Constructs.
+            Type();
+        };
+
+        /// @brief Describes a UI draw instruction.
+        /// @ingroup ui-canvas-plugin
+        struct CUBOS_ENGINE_API Command
+        {
+            cubos::core::gl::VertexArray vertexArray;                    ///< Vertex array handle.
+            size_t vertexOffset;                                         ///< Index of the first vertex to be drawn.
+            size_t vertexCount;                                          ///< Number of vertexes to draw.
+            cubos::core::gl::Sampler samplers[Type::MaxTextures] = {};   ///< Samplers.
+            cubos::core::gl::Texture2D textures[Type::MaxTextures] = {}; ///< Textures.
+        };
+
+        /// @brief Describes a draw commands and how it fits into the draw list carrying it.
+        /// @ingroup ui-canvas-plugin
+        struct CUBOS_ENGINE_API Entry
+        {
+            /// @brief Constructs.
+            /// @param commandType Command type.
+            explicit Entry(const Type& commandType);
+
+            /// @brief Constructs.
+            /// @param commandType Command type.
+            /// @param drawCommand UI draw command.
+            /// @param dataOffset Position of the command data in the draw list.
+            Entry(const Type& commandType, Command drawCommand, size_t dataOffset);
+
+            const Type& type; ///< Command type.
+            Command command;  ///< UI draw command.
+            size_t offset;    ///< Position of the command data in the draw list.
+        };
+
+        /// @brief Builds an Entry.
+        /// @ingroup ui-canvas-plugin
+        class CUBOS_ENGINE_API EntryBuilder
+        {
+        public:
+            /// @brief Constructs builder with required data.
+            /// @param entry Reference to the associated entry.
+            EntryBuilder(Entry& entry);
+
+            /// @brief Adds a texture to the entry.
+            /// @param texIndex Index of the texture to add.
+            /// @param texture Texture to add.
+            /// @param sampler Sampler to be used with the texture.
+            /// @returns The builder.
+            EntryBuilder withTexture(int texIndex, cubos::core::gl::Texture2D texture,
+                                     cubos::core::gl::Sampler sampler = nullptr);
+
+        private:
+            Entry& mEntry;
+        };
+
+        /// @brief Constructs a draw list.
+        UIDrawList();
+
+        /// @brief Move constructs.
+        /// @param other Other draw list.
+        UIDrawList(UIDrawList&& other) noexcept;
+
+        /// @brief Copy constructs.
+        /// @param other Other draw list.
+        UIDrawList(const UIDrawList& other) noexcept;
+
+        ~UIDrawList();
+
+        /// @brief Adds entry to the draw list.
+        /// @param type Command type.
+        /// @param command UI draw command.
+        /// @param data Command data.
+        /// @return A builder for the created entry.
+        EntryBuilder push(const Type& type, const Command& command, const void* data);
+
+        /// @brief Creates a new entry in the draw list.
+        /// @param type Command type.
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Command data.
+        /// @return A builder for the created entry.
+        EntryBuilder push(const Type& type, core::gl::VertexArray vertexArray, size_t vertexOffset, size_t vertexCount,
+                          const void* data);
+
+        /// @brief Empties draw list.
+        void clear();
+
+        /// @brief Gets list size.
+        /// @return Number of elements in the list.
+        std::size_t size() const;
+
+        /// @brief Gets an entry from the list.
+        /// @param i Entry index.
+        /// @return A list entry.
+        const Entry& entry(std::size_t i) const;
+
+        /// @brief Gets the data of an entry.
+        /// @param i Entry index.
+        /// @return A pointer to the entry's data.
+        const void* data(std::size_t i) const;
+
+    private:
+        std::vector<Entry> mEntries; ///< List holding the draw commands of this element and all its children.
+
+        void* mData;                 ///< Buffer holding the data for all the draw commands.
+        size_t mDataCapacity = 1024; ///< How much data the data buffer can hold.
+        size_t mDataSize = 0;        ///< How much data the data buffer is holding.
+    };
+} // namespace cubos::engine
--- a/engine/include/cubos/engine/ui/canvas/element.hpp
+++ b/engine/include/cubos/engine/ui/canvas/element.hpp
@@ -10,6 +10,7 @@
 #include <cubos/core/reflection/reflect.hpp>
 
 #include <cubos/engine/api.hpp>
+#include <cubos/engine/ui/canvas/draw_list.hpp>
 
 using namespace cubos::core::gl;
 
@@ -20,7 +21,6 @@ namespace cubos::engine
     struct CUBOS_ENGINE_API UIElement
     {
         CUBOS_REFLECT;
-
         glm::vec2 offset = {0.0F, 0.0F}; ///< Offset of the pivot.
         glm::vec2 size = {0.0F, 0.0F};   ///< Size of the element.
         glm::vec2 pivot = {0.5F, 0.5F};  ///< Relative position of the pivot to the element itself.
@@ -28,9 +28,10 @@ namespace cubos::engine
         int layer = 0;                   ///< Rendering order layer.
 
         glm::vec2 position = {0.0F, 0.0F}; ///< Global position of the pivot.
-        glm::mat4 vp = {};                 ///< View matrix provided by the canvas.
 
         int hierarchyDepth = 0; ///< Depth in hierarchy. Used for depth sorting.
+
+        UIDrawList drawList = {}; ///< Draw list holding draw commands from this element.
 
         /// @brief Gets global position of the bottom left corner of the element.
         /// @return Position value in global coordinates.
@@ -39,5 +40,30 @@ namespace cubos::engine
         /// @brief Gets global position of the top right corner of the element.
         /// @return Position value in global coordinates.
         glm::vec2 topRightCorner() const;
+
+        /// @brief Adds new entry to the draw list, with undefined data type.
+        /// @param type Command type.
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Untyped command data.
+        /// @return A builder for the created entry.
+        UIDrawList::EntryBuilder drawUntyped(const UIDrawList::Type& type, const core::gl::VertexArray& vertexArray,
+                                             size_t vertexOffset, size_t vertexCount, const void* data);
+
+        /// @brief Adds new entry to the draw list, with a defined data type.
+        /// @tparam T Command data type.
+        /// @param type Command type.
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Typed command data.
+        /// @return A builder for the created entry.
+        template <typename T>
+        UIDrawList::EntryBuilder draw(const UIDrawList::Type& type, const core::gl::VertexArray& vertexArray,
+                                      size_t vertexOffset, size_t vertexCount, const T& data)
+        {
+            return this->drawUntyped(type, vertexArray, vertexOffset, vertexCount, &data);
+        }
     };
 } // namespace cubos::engine
--- a/engine/include/cubos/engine/ui/canvas/plugin.hpp
+++ b/engine/include/cubos/engine/ui/canvas/plugin.hpp
@@ -12,16 +12,32 @@
 namespace cubos::engine
 {
     /// @defgroup ui-canvas-plugin Canvas
-    /// @ingroup engine
-    /// @brief Adds UI canvas and elements.
+    /// @ingroup ui-plugins
+    /// @brief Adds the core UI functionality and rendering logic.
+    ///
+    /// # Architecture
+    ///
+    /// The two main UI components are @ref UIElement and @ref UICanvas.
+    /// For it to work properly, each UIElement must be a child of a UICanvas, or of another UIElement.
+    ///
+    /// Each UIElement generates a list of drawing commands, @ref UIDrawList, that are then pooled together by the
+    /// UICanvas at the root of the hierarchy.
+    ///
+    /// Each entry in a draw list contains a @ref UIDrawList::Type and a @ref UIDrawList::Command. The Type holds
+    /// information that is common for all similar commands, and the Command contains the specifics to that one command.
+    /// Each entry also has a raw data buffer to hold the data the command needs to pass to the shaders.
+    ///
+    /// After aggregating the draw commands, the UICanvas draws them, grouped by type.
+    ///
+    /// Check out @ref examples-engine-ui for the an example of the UI plugin in use.
 
     /// @brief @ref UICanvas components' information is passed to related @ref UIElement components.
     /// @ingroup canvas-target-plugin
-    CUBOS_ENGINE_API extern Tag elementVPUpdateTag;
+    CUBOS_ENGINE_API extern Tag uiCanvasChildrenUpdateTag;
 
     /// @brief The @ref UIElement components are updated with information from their parents.
     /// @ingroup canvas-target-plugin
-    CUBOS_ENGINE_API extern Tag elementPropagateTag;
+    CUBOS_ENGINE_API extern Tag uiElementPropagateTag;
 
     /// @brief Readies canvas for drawing.
     /// @ingroup canvas-target-plugin
@@ -30,6 +46,10 @@ namespace cubos::engine
     /// @brief Systems which draw to canvas should be tagged with this.
     /// @ingroup canvas-target-plugin
     CUBOS_ENGINE_API extern Tag uiDrawTag;
+
+    /// @brief All queued draw commands are collected and executed to the render targets.
+    /// @ingroup canvas-target-plugin
+    CUBOS_ENGINE_API extern Tag uiEndTag;
 
     /// @brief Plugin entry function.
     /// @param cubos @b CUBOS. main class
--- a/engine/samples/ui/main.cpp
+++ b/engine/samples/ui/main.cpp
@@ -43,12 +43,12 @@ int main(int argc, char** argv)
                              .entity();
         commands.relate(elementBg, canvas, ChildOf{});
         /// [Set up Background]
-
         /// [Set up Panel]
-        auto elementPanel = commands.create()
-                                .add(UIElement{{0, -50}, {200, 600}, {1, 0.5F}, {1, 0.5F}})
-                                .add(UIColorRect{{1, 0, 0, 1}})
-                                .entity();
+        auto elementPanel =
+            commands.create()
+                .add(UIElement{.offset = {-50, 0}, .size = {200, 600}, .pivot = {1, 0.5F}, .anchor = {1, 0.5F}})
+                .add(UIColorRect{{1, 0, 0, 1}})
+                .entity();
         commands.relate(elementPanel, elementBg, ChildOf{});
         /// [Set up Panel]
     });
--- a/engine/src/ui/canvas/draw_list.cpp
+++ b/engine/src/ui/canvas/draw_list.cpp
@@ -1,0 +1,125 @@
+#include <cstring>
+#include <utility>
+
+#include <cubos/core/log.hpp>
+
+#include <cubos/engine/ui/canvas/draw_list.hpp>
+
+cubos::engine::UIDrawList::Type::Type()
+{
+    for (auto& i : texBindingPoint)
+    {
+        i = nullptr;
+    }
+}
+
+cubos::engine::UIDrawList::Entry::Entry(const Type& commandType)
+    : type(commandType)
+{
+}
+
+cubos::engine::UIDrawList::Entry::Entry(const Type& commandType, Command drawCommand, size_t dataOffset)
+    : type(commandType)
+    , command(std::move(drawCommand))
+    , offset(dataOffset)
+{
+}
+
+cubos::engine::UIDrawList::EntryBuilder::EntryBuilder(Entry& entry)
+    : mEntry(entry)
+{
+}
+
+cubos::engine::UIDrawList::EntryBuilder cubos::engine::UIDrawList::EntryBuilder::withTexture(
+    int texIndex, cubos::core::gl::Texture2D texture, cubos::core::gl::Sampler sampler)
+{
+    if (texIndex >= Type::MaxTextures)
+    {
+        CUBOS_WARN("Invalid texture index, texture will be dropped");
+        return *this;
+    }
+    mEntry.command.samplers[texIndex] = std::move(sampler);
+    mEntry.command.textures[texIndex] = std::move(texture);
+    return *this;
+}
+
+cubos::engine::UIDrawList::UIDrawList()
+{
+    mData = operator new(sizeof(char) * mDataCapacity);
+}
+
+cubos::engine::UIDrawList::UIDrawList(UIDrawList&& other) noexcept
+    : mEntries(std::move(other.mEntries))
+    , mData(other.mData)
+    , mDataCapacity(other.mDataCapacity)
+    , mDataSize(other.mDataSize)
+{
+    other.mData = nullptr;
+    other.mDataSize = 0;
+    other.mDataCapacity = 0;
+}
+
+cubos::engine::UIDrawList::UIDrawList(const UIDrawList& other) noexcept
+    : mEntries(other.mEntries)
+    , mDataCapacity(other.mDataSize)
+    , mDataSize(other.mDataSize)
+{
+    mData = operator new(sizeof(char) * other.mDataSize);
+    std::memcpy(mData, other.mData, sizeof(char) * other.mDataSize);
+}
+
+cubos::engine::UIDrawList::~UIDrawList()
+{
+    operator delete(mData);
+}
+
+cubos::engine::UIDrawList::EntryBuilder cubos::engine::UIDrawList::push(const Type& type, const Command& command,
+                                                                        const void* data)
+{
+    auto entry = Entry{type, command, mDataSize};
+    entry.offset = mDataSize;
+    mEntries.push_back(entry);
+    if (mDataSize + entry.type.perElementSize > mDataCapacity)
+    {
+        mDataCapacity += entry.type.perElementSize;
+        mDataCapacity *= 2;
+        void* temp = operator new(sizeof(char) * mDataCapacity);
+        std::memcpy(temp, mData, mDataSize);
+        operator delete(mData);
+        mData = temp;
+    }
+    std::memcpy(static_cast<char*>(mData) + mDataSize, data, entry.type.perElementSize);
+    mDataSize += entry.type.perElementSize;
+    return {mEntries[mEntries.size() - 1]};
+}
+
+cubos::engine::UIDrawList::EntryBuilder cubos::engine::UIDrawList::push(const Type& type,
+                                                                        core::gl::VertexArray vertexArray,
+                                                                        size_t vertexOffset, size_t vertexCount,
+                                                                        const void* data)
+{
+    auto command =
+        Command{.vertexArray = std::move(vertexArray), .vertexOffset = vertexOffset, .vertexCount = vertexCount};
+    return push(type, command, data);
+}
+
+void cubos::engine::UIDrawList::clear()
+{
+    mDataSize = 0;
+    mEntries.clear();
+}
+
+std::size_t cubos::engine::UIDrawList::size() const
+{
+    return mEntries.size();
+}
+
+const cubos::engine::UIDrawList::Entry& cubos::engine::UIDrawList::entry(std::size_t i) const
+{
+    return mEntries[i];
+}
+
+const void* cubos::engine::UIDrawList::data(std::size_t i) const
+{
+    return static_cast<void*>(static_cast<char*>(mData) + mEntries[i].offset);
+}
--- a/engine/src/ui/canvas/plugin.cpp
+++ b/engine/src/ui/canvas/plugin.cpp
@@ -1,3 +1,6 @@
+#include <map>
+#include <utility>
+
 #include <glm/gtc/matrix_transform.hpp>
 
 #include <cubos/core/io/window.hpp>
@@ -7,18 +10,49 @@
 #include <cubos/engine/transform/child_of.hpp>
 #include <cubos/engine/transform/plugin.hpp>
 #include <cubos/engine/ui/canvas/canvas.hpp>
+#include <cubos/engine/ui/canvas/draw_list.hpp>
 #include <cubos/engine/ui/canvas/element.hpp>
 #include <cubos/engine/ui/canvas/horizontal_stretch.hpp>
 #include <cubos/engine/ui/canvas/plugin.hpp>
 #include <cubos/engine/ui/canvas/vertical_stretch.hpp>
 #include <cubos/engine/window/plugin.hpp>
 
-CUBOS_DEFINE_TAG(cubos::engine::elementVPUpdateTag);
-CUBOS_DEFINE_TAG(cubos::engine::elementPropagateTag)
+CUBOS_DEFINE_TAG(cubos::engine::uiCanvasChildrenUpdateTag);
+CUBOS_DEFINE_TAG(cubos::engine::uiElementPropagateTag)
 CUBOS_DEFINE_TAG(cubos::engine::uiBeginTag)
 CUBOS_DEFINE_TAG(cubos::engine::uiDrawTag)
+CUBOS_DEFINE_TAG(cubos::engine::uiEndTag)
 
 using namespace cubos::engine;
+
+namespace
+{
+    struct State
+    {
+        cubos::core::gl::DepthStencilState dss;
+        cubos::core::gl::ConstantBuffer mvpBuffer;
+        State(cubos::core::gl::DepthStencilState depthStencilState, cubos::core::gl::ConstantBuffer mvpConstantBuffer)
+            : dss(std::move(depthStencilState))
+            , mvpBuffer(std::move(mvpConstantBuffer))
+        {
+        }
+    };
+} // namespace
+
+static void copyCommands(Query<Entity, UIElement&, UIElement&> query, UIElement& element, Entity entity,
+                         std::map<const UIDrawList::Type*, UIDrawList>& commands)
+{
+    for (size_t i = 0; i < element.drawList.size(); i++)
+    {
+        auto entry = element.drawList.entry(i);
+        commands[&entry.type].push(entry.type, entry.command, element.drawList.data(i));
+    }
+    element.drawList.clear();
+    for (auto [child, childElement, _] : query.pin(1, entity))
+    {
+        copyCommands(query, childElement, child, commands);
+    }
+}
 
 void cubos::engine::uiCanvasPlugin(Cubos& cubos)
 {
@@ -30,11 +64,13 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
     cubos.component<UICanvas>();
     cubos.component<UIHorizontalStretch>();
     cubos.component<UIVerticalStretch>();
+    cubos.uninitResource<State>();
 
-    cubos.tag(elementVPUpdateTag);
-    cubos.tag(elementPropagateTag).after(elementVPUpdateTag);
-    cubos.tag(uiBeginTag).after(elementPropagateTag);
-    cubos.tag(uiDrawTag).after(uiBeginTag).tagged(drawToRenderTargetTag);
+    cubos.tag(uiCanvasChildrenUpdateTag);
+    cubos.tag(uiElementPropagateTag).after(uiCanvasChildrenUpdateTag);
+    cubos.tag(uiBeginTag).after(uiElementPropagateTag);
+    cubos.tag(uiDrawTag).after(uiBeginTag);
+    cubos.tag(uiEndTag).after(uiDrawTag).tagged(drawToRenderTargetTag);
 
     cubos.observer("initialize Canvas").onAdd<UICanvas>().call([](Query<UICanvas&> query) {
         for (auto [canvas] : query)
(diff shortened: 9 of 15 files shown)


engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +138,52 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                for (auto [element, _] : query.pin(1, entity))
+                {
+                    for (auto entry : element.drawList)
+                    {
+                        commands[&entry.type].push_back(entry.command);
+                    }
+                }
+
+                window->renderDevice().setViewport(0, 0, static_cast<int>(window->framebufferSize()[0]),
+                                                   static_cast<int>(window->framebufferSize()[1]));
```
RiscadoA, line 158: Here you should access the render target's size instead of the window's. Just add the `RenderTarget` component to the canvasQuery, as they're part of the same entity.

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +138,52 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                auto mvpBuffer = window->renderDevice().createConstantBuffer(sizeof(glm::mat4), &canvas.mat,
+                                                                             cubos::core::gl::Usage::Static);
+                for (auto type : std::ranges::views::keys(commands))
+                {
+                    window->renderDevice().setShaderPipeline(type->pipeline);
+                    window->renderDevice().setDepthStencilState(type->depthStencilState);
```
RiscadoA, line 164: Do we really need a per-command type depth-stencil state?

engine/include/cubos/engine/ui/canvas/draw_type.hpp
```diff
@@ -0,0 +1,24 @@
...
+    /// @brief Describes a UI draw call type.
+    /// @ingroup ui-canvas-plugin
+    struct DrawType
+    {
+        cubos::core::gl::ShaderPipeline pipeline;       ///< Handle to shader pipeline.
+        cubos::core::gl::ConstantBuffer constantBuffer; ///< Handle to the constant buffer to hold the per element data.
+        cubos::core::gl::ShaderBindingPoint constantBufferBindingPoint; ///< Binding point of the element data.
+        size_t perElementSize;                                  ///< Size of the struct holding the element data.
+        cubos::core::gl::DepthStencilState depthStencilState;   ///< Depth stencil state to use when drawing.
+        cubos::core::gl::ShaderBindingPoint texBindingPoint[4]; ///< Textures binding points.
```
RiscadoA, line 22:
```suggestion
        /// @brief Maximum texture binding point count.
        static constexpr int MaxTextures = 4;

        cubos::core::gl::ShaderPipeline pipeline;       ///< Handle to shader pipeline.
        cubos::core::gl::ConstantBuffer constantBuffer; ///< Handle to the constant buffer to hold the per element data.
        cubos::core::gl::ShaderBindingPoint constantBufferBindingPoint{nullptr}; ///< Binding point of the element data.
        size_t perElementSize{0};                                  ///< Size of the struct holding the element data.
        cubos::core::gl::DepthStencilState depthStencilState{nullptr};   ///< Depth stencil state to use when drawing.
        cubos::core::gl::ShaderBindingPoint texBindingPoint[MaxTextures]{}; ///< Textures binding points.
```

engine/samples/ui/main.cpp
```diff
@@ -43,12 +43,13 @@ int main(int argc, char** argv)
...
-        auto elementPanel = commands.create()
-                                .add(UIElement{{0, -50}, {200, 600}, {1, 0.5F}, {1, 0.5F}})
-                                .add(UIColorRect{{1, 0, 0, 1}})
-                                .entity();
+        UIElement element;
+        element.offset = {-50, 0};
+        element.size = {200, 600};
+        element.anchor = {1, 0.5F};
+        element.pivot = {1, 0.5F};
+        auto elementPanel = commands.create().add(element).add(UIColorRect{{1, 0, 0, 1}}).entity();
```
RiscadoA, line 52:
```suggestion
        auto elementPanel = commands.create().add(UIElement{.offset = {-50, 0}, ...}).add(UIColorRect{{1, 0, 0, 1}}).entity();
```

RiscadoA (changes requested):
Thanks for working on this! Made a first pass on the higher level details.

engine/CMakeLists.txt
RiscadoA, line 1: Can you revert the formatting change? It dirties the file history :pensive:

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+        {
+            other.perElementData = nullptr;
+        }
+
+        /// @brief Copy constructs.
+        /// @param other Other draw list.
+        UIDrawList(const UIDrawList& other) noexcept
+            : entries(other.entries)
+        {
+            perElementData = operator new(sizeof(char) * 1024);
```
RiscadoA, line 41:
```suggestion
            perElementData = operator new(sizeof(char) * other.totalSize);
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+            std::memcpy(perElementData, other.perElementData, sizeof(char) * 1024);
```
RiscadoA, line 42:
```suggestion
            std::memcpy(perElementData, other.perElementData, sizeof(char) * other.totalSize);
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+    /// @ingroup ui-canvas-plugin
+    class UIDrawList
+    {
+    public:
+        /// @brief Constructs a draw list.
+        UIDrawList();
+
+        /// @brief Move constructs.
+        /// @param other Other draw list.
+        UIDrawList(UIDrawList&& other) noexcept
```
RiscadoA, line 29: Move this and the constructors below to the .cpp

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+    struct UIDrawListEntry
+    {
+        UIDrawType& type;
+        UIDrawCommand command;
+        size_t offset;
+    };
+
+    /// @brief Holds a collections of draw commands and their data.
+    /// @ingroup ui-canvas-plugin
+    class UIDrawList
```
RiscadoA, line 21:
```suggestion
    class CUBOS_ENGINE_API UIDrawList
```
Also missing in other struct and class definitions.

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+        /// @return The created command.
+        UIDrawCommand& emplaceCommand(UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset,
+                                      size_t vertexCount, void* data);
+
+        /// @brief Empties draw list.
+        void clear();
+
+        std::vector<UIDrawListEntry> entries; ///< List holding the draw commands of this element and all its children.
+
+        void* perElementData; ///< Buffer holding the data for all the draw commands.
```
RiscadoA, line 69: These should be private, if they're not meant to be modified directly.
If they need to be accessed, add `const` getters:

```cpp
std::size_t size() const;
const UIDrawListEntry& entry(std::size_t i) const;
const void* data(std::size_t i) const;
```

engine/include/cubos/engine/ui/canvas/element.hpp
```diff
@@ -39,5 +40,25 @@ namespace cubos::engine
...
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Pointer to per element data.
+        /// @return The created command.
+        UIDrawCommand& emplaceCommand(UIDrawType& type, VertexArray vertexArray, size_t vertexOffset,
+                                      size_t vertexCount, void* data);
+
+        /// @brief Empties draw list.
+        void clearCommands();
+
+        UIDrawList drawList; ///< Draw list holding draw commands from this element.
```
RiscadoA, line 62: Move this above, next to the other fields.

engine/include/cubos/engine/ui/canvas/draw_command.hpp
```diff
@@ -0,0 +1,35 @@
...
+    struct UIDrawCommand
+    {
+        static constexpr int MaxTextures = 4;
+
+        cubos::core::gl::VertexArray vertexArray;         ///< Vertex array handle.
+        size_t vertexOffset;                              ///< Index of the first vertex to be drawn.
+        size_t vertexCount;                               ///< Number of vertexes to draw.
+        cubos::core::gl::Sampler samplers[MaxTextures];   ///< Samplers.
+        cubos::core::gl::Texture2D textures[MaxTextures]; ///< Textures.
+        void* perElementData;                             ///< Element raw data.
```
RiscadoA, line 24: Why is this here if you're already storing the offset in the UIDrawListEntry?

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+    private:
+        size_t current_capacity = 1024;
```
RiscadoA, line 71:
```suggestion
        size_t dataCapacity = 1024;
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+        size_t totalSize = 0;
```
RiscadoA, line 72:
```suggestion
        size_t dataSize = 0;
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
RiscadoA, line 69:
```suggestion
        void* data; ///< Buffer holding the per element data for all the draw commands.
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+        }
+        ~UIDrawList()
+        {
+            operator delete(perElementData);
+        }
+
+        /// @brief Adds entry to draw list.
+        /// @param entry Entry to add.
+        /// @param data Pointer to per element data.
+        void pushEntry(UIDrawListEntry entry, const void* data);
```
RiscadoA, line 52:
```suggestion
        void push(UIDrawListEntry entry, const void* data);
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+        /// @brief Creates new draw command and adds it to draw list.
+        /// @param type Type of the command.
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Pointer to per element data.
+        /// @return The created command.
+        UIDrawCommand& emplaceCommand(UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset,
+                                      size_t vertexCount, void* data);
```
RiscadoA, line 62:
```suggestion
        UIDrawCommand& draw(const UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset,
                                      size_t vertexCount, const void* data);
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+            operator delete(perElementData);
+        }
+
+        /// @brief Adds entry to draw list.
+        /// @param entry Entry to add.
+        /// @param data Pointer to per element data.
+        void pushEntry(UIDrawListEntry entry, const void* data);
+
+        /// @brief Creates new draw command and adds it to draw list.
+        /// @param type Type of the command.
```
RiscadoA, line 55:
```suggestion
        /// @param type Command type.
```
We usually avoid 'of the'

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,74 @@
...
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Pointer to per element data.
+        /// @return The created command.
```
RiscadoA, line 60: Why do you return a reference? This is dangerous, as the reference may become invalid the next time the method is called (due to std::vector moving the memory around).
If you just want the user to be able to customize the command further, I would return a builder with helper methods. This makes it more clear that its supposed to be used immediately after and not kept around in a variable.

engine/include/cubos/engine/ui/canvas/draw_command.hpp
```diff
@@ -0,0 +1,31 @@
...
+        cubos::core::gl::Sampler samplers[4];     ///< Samplers.
+        cubos::core::gl::Texture2D textures[4];   ///< Textures.
+        void* perElementData;                     ///< Element raw data.
+        int numTextures = 0;                      ///< Number of textures added to command.
+
+        /// @brief Adds a texture to the command.
+        /// @param texture Texture to add.
+        /// @param sampler Sampler to be used with the texture.
+        /// @returns Reference to the draw command.
+        DrawCommand& AddTexture(cubos::core::gl::Texture2D texture, cubos::core::gl::Sampler sampler = nullptr);
```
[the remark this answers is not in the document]
RiscadoA: You kept this, was it by mistake?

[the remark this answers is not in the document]
DiogoMendonc-a: nah, changed it to include the index as a parameter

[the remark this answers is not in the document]
RiscadoA: Why keep a `textureCount` variable then? Wouldn't this be a property of the command type and not of the command itself?

engine/include/cubos/engine/ui/canvas/element.hpp
```diff
@@ -20,24 +23,32 @@ namespace cubos::engine
...
         glm::vec2 bottomLeftCorner() const;
 
         /// @brief Gets global position of the top right corner of the element.
         /// @return Position value in global coordinates.
         glm::vec2 topRightCorner() const;
+
+        /// @brief Adds entry to draw list.
+        /// @param entry Entry to add.
+        /// @param data Pointer to per element data.
+        void pushEntry(UIDrawListEntry entry, const void* data);
```
RiscadoA, line 49: Wouldn't it be more friendly to return a builder here to modify the added entry?
Since this is what the user will interact with when making custom widgets and stuff it should look as good as possible.
I.e. `Builder draw(const UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset, size_t vertexCount, const void* data)`

This was what I was suggesting before, sorry for not making it clearer
What I was imagining was being able to do something like:
```
element.draw(type, va, 0, 6, perElementData).withTexture(0, myTexture);
```
Basically the builder would just be a wrapper around a reference to an UIDrawEntry.
What do you think?

---

As a sidenote, I would maybe name this `drawUntyped`, and add a wrapper templated method which calls it. For example,

```cpp
template <typename T>
Builder draw(const UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset, size_t vertexCount, const T& data)
{
  return this->drawUntyped(type, vertexArray, vertexOffset, vertexCount, &data);
}
```

Then the color rect code could become just:

```cpp
element.draw(state.drawType, state.va, 0, 6, PerElement{
  .xRange = {minX, maxX},
  .yRange = {minY, maxY},
  .color = colorRect.color,
  .depth = element.layer * 100 + element.hierarchyDepth
});
```

engine/include/cubos/engine/ui/canvas/element.hpp
```diff
@@ -20,24 +23,32 @@ namespace cubos::engine
...
+
+        /// @brief Empties draw list.
+        void clearCommands();
```
RiscadoA, line 52: No need to have this method as the canvas will access the drawList directly anyway I think.

engine/src/ui/canvas/plugin.cpp
```diff
@@ -7,19 +9,36 @@
...
-CUBOS_DEFINE_TAG(cubos::engine::elementPropagateTag)
+CUBOS_DEFINE_TAG(cubos::engine::uiElementVPUpdateTag);
+CUBOS_DEFINE_TAG(cubos::engine::uiElementPropagateTag)
 CUBOS_DEFINE_TAG(cubos::engine::uiBeginTag)
 CUBOS_DEFINE_TAG(cubos::engine::uiDrawTag)
+CUBOS_DEFINE_TAG(cubos::engine::uiEndTag)
 
 using namespace cubos::engine;
 
+void copyCommands(Query<Entity, UIElement&, UIElement&> query, UIElement& element, Entity entity,
```
RiscadoA, line 27:
```suggestion
static void copyCommands(Query<Entity, UIElement&, UIElement&> query, UIElement& element, Entity entity,
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +134,58 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+        .with<UIElement>()
+        .call([](Query<Entity, UIElement&, const UICanvas&> drawsToQuery,
+                 Query<Entity, UIElement&, UIElement&> childrenQuery,
+                 Query<Entity, UICanvas&, RenderTarget&> canvasQuery, const core::io::Window& window) {
+            cubos::core::gl::DepthStencilStateDesc dssd;
+            dssd.depth.enabled = true;
+            dssd.depth.writeEnabled = true;
+            dssd.depth.near = -1.0F;
+            dssd.depth.compare = Compare::LEqual;
+            window->renderDevice().setDepthStencilState(window->renderDevice().createDepthStencilState(dssd));
```
RiscadoA, line 155: Create this once in a startup system and just bind it here (keep it on a `State` resource ig).

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +134,58 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+
+            for (auto [entity, canvas, renderTarget] : canvasQuery)
+            {
+                window->renderDevice().setViewport(0, 0, static_cast<int>(renderTarget.size.x),
+                                                   static_cast<int>(renderTarget.size.y));
+                auto mvpBuffer = window->renderDevice().createConstantBuffer(sizeof(glm::mat4), &canvas.mat,
+                                                                             cubos::core::gl::Usage::Static);
```
RiscadoA, line 162: Same as above, this should only be created once as it is an expensive operation. You can reuse it every frame.

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +134,58 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                    window->renderDevice().setShaderPipeline(type->pipeline);
+                    type->pipeline->getBindingPoint("MVP")->bind(mvpBuffer);
+
+                    for (size_t i = 0; i < list.size(); i++)
+                    {
+                        auto entry = list.entry(i);
+                        window->renderDevice().setVertexArray(entry.command.vertexArray);
+
+                        type->constantBuffer->fill(list.data(i), type->perElementSize);
+                        type->constantBufferBindingPoint->bind(type->constantBuffer);
```
RiscadoA, line 179: You just need to bind it once, after you set the shader pipeline (before this inner loop).

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,91 @@
...
+    , mData(other.mData)
+{
+    other.mData = nullptr;
+}
+
+cubos::engine::UIDrawList::UIDrawList(const UIDrawList& other) noexcept
+    : mEntries(other.mEntries)
+{
+    mData = operator new(sizeof(char) * other.mDataSize);
+    std::memcpy(mData, other.mData, sizeof(char) * other.mDataSize);
```
RiscadoA, line 49:
```suggestion
    mData = operator new(sizeof(char) * other.mDataSize);
    mDataCapacity = other.mDataSize;
    mDataSize = other.mDataSize;
    std::memcpy(mData, other.mData, sizeof(char) * other.mDataSize);
```

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,91 @@
...
+cubos::engine::UIDrawList::UIDrawList()
+{
+    mData = operator new(sizeof(char) * mDataCapacity);
+}
+
+cubos::engine::UIDrawList::UIDrawList(UIDrawList&& other) noexcept
+    : mEntries(std::move(other.mEntries))
+    , mData(other.mData)
+{
+    other.mData = nullptr;
```
RiscadoA, line 42:
```suggestion
    other.mData = nullptr;
    other.mDataSize = 0;
    other.mDataCapacity = 0;
```

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+void cubos::engine::UIDrawList::push(UIDrawListEntry entry, const void* data)
+{
+    entry.offset = mDataSize;
+    mEntries.push_back(entry);
+    if (mDataSize + entry.type.perElementSize > mDataCapacity)
+    {
+        mDataCapacity += entry.type.perElementSize;
+        void* temp = operator new(sizeof(char) * mDataCapacity);
+        operator delete(mData);
```
RiscadoA, line 65:
```suggestion
        void* temp = operator new(sizeof(char) * mDataCapacity);
        std::memcpy(temp, mData, mDataSize);
        operator delete(mData);
```

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,91 @@
...
+    operator delete(mData);
+}
+
+void cubos::engine::UIDrawList::push(UIDrawListEntry entry, const void* data)
+{
+    entry.offset = mDataSize;
+    mEntries.push_back(entry);
+    if (mDataSize + entry.type.perElementSize > mDataCapacity)
+    {
+        mDataCapacity += entry.type.perElementSize;
```
RiscadoA, line 63:
```suggestion
        mDataCapacity += entry.type.perElementSize;
        mDataCapacity *= 2;
```
Better to also grow it exponentially to avoid having to make many small allocations.

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+cubos::engine::UIDrawList::~UIDrawList()
+{
+    operator delete(mData);
+}
+
+void cubos::engine::UIDrawList::push(UIDrawListEntry entry, const void* data)
+{
+    entry.offset = mDataSize;
```
RiscadoA, line 59:
```suggestion
void cubos::engine::UIDrawList::push(const UIDrawType& type, UIDrawCommand command, const void* data)
{
    auto entry = UIDrawEntry{.type = type, .command = command, .offset = mDataSize};
```
The only extra field the entry provides is `offset`, and it isn't supposed to be set by the caller, so may as well construct the entry within the function.

engine/include/cubos/engine/ui/canvas/draw_type.hpp
```diff
@@ -0,0 +1,28 @@
...
+    struct CUBOS_ENGINE_API UIDrawType
+    {
+        static constexpr int MaxTextures = 4;
+
+        cubos::core::gl::ShaderPipeline pipeline;       ///< Handle to shader pipeline.
+        cubos::core::gl::ConstantBuffer constantBuffer; ///< Handle to the constant buffer to hold the per element data.
+        cubos::core::gl::ShaderBindingPoint constantBufferBindingPoint; ///< Binding point of the element data.
+        size_t perElementSize; ///< Size of the struct holding the element data.
+        cubos::core::gl::ShaderBindingPoint texBindingPoint[MaxTextures]; ///< Textures binding points.
+        size_t numTextures = 0;                                           ///< Number of textures the shader requires.
```
RiscadoA, line 26:
```suggestion
```
You can replace this by just checking if the respective binding point is null on the canvas rendering.
Simplifies a bit the things on the user's side, as they don't have to manage the `numTextures`, at the cost of having to iterate over all of the 4 bindings points in the canvas side. The performance penalty shouldn't be a problem though.

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,92 @@
...
+        /// @param entry Entry to add.
+        /// @param data Pointer to per element data.
+        void push(UIDrawListEntry entry, const void* data);
+
+        /// @brief Empties draw list.
+        void clear();
+
+        std::size_t size() const;
+        const UIDrawListEntry& entry(std::size_t i) const;
+        const void* data(std::size_t i) const;
```
RiscadoA, line 83: Missing docs

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
RiscadoA, line 1: Missing file docs

RiscadoA (changes requested):
I think this is the last big request batch from me. Sorry for being so nit picky about the details!
Just trying to make sure that this core part of the UI remains as simple and lean as possible.
Ty for the good work :pray:

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+namespace cubos::engine
+{
+    /// @brief Holds a collections of draw commands and their data.
+    /// @ingroup ui-canvas-plugin
+    class CUBOS_ENGINE_API UIDrawList final
+    {
+    public:
+        /// @brief Describes a UI draw call type.
+        /// @ingroup ui-canvas-plugin
+        struct CUBOS_ENGINE_API UIDrawType
```
RiscadoA, line 21: Since these are now internal to the class, you should remove the `UIDraw` prefix.

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+                for (size_t i = 0; i < MaxTextures; i++)
+                {
+                    texBindingPoint[i] = nullptr;
+                }
+            }
+        };
+
+        /// @brief Describes a UI draw instruction.
+        /// @ingroup ui-canvas-plugin
+        struct CUBOS_ENGINE_API UIDrawCommand
```
RiscadoA, line 44: Same as above

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+            cubos::core::gl::VertexArray vertexArray;                     ///< Vertex array handle.
+            size_t vertexOffset;                                          ///< Index of the first vertex to be drawn.
+            size_t vertexCount;                                           ///< Number of vertexes to draw.
+            cubos::core::gl::Sampler samplers[UIDrawType::MaxTextures];   ///< Samplers.
+            cubos::core::gl::Texture2D textures[UIDrawType::MaxTextures]; ///< Textures.
+        };
+
+        /// @brief Describes a draw commands and how it fits into the draw list carrying it.
+        /// @ingroup ui-canvas-plugin
+        struct UIDrawListEntry
```
RiscadoA, line 55: Same as above

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+            cubos::core::gl::ShaderBindingPoint texBindingPoint[MaxTextures]; ///< Textures binding points.
+
+            /// @brief Constructs.
+            UIDrawType()
+            {
+                for (size_t i = 0; i < MaxTextures; i++)
+                {
+                    texBindingPoint[i] = nullptr;
+                }
+            }
```
RiscadoA, line 39: Move to .cpp

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+            /// @brief Constructs.
+            /// @param commandType Command type.
+            /// @param drawCommand UI draw command.
+            /// @param dataOffset Position of the command data in the draw list.
+            UIDrawListEntry(const UIDrawType& commandType, const UIDrawCommand& drawCommand, size_t dataOffset)
+                : type(commandType)
+                , command(drawCommand)
+                , offset(dataOffset)
+            {
+            }
```
RiscadoA, line 73: Move to .cpp

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+            const UIDrawType& type; ///< Command type.
+            UIDrawCommand command;  ///< UI draw command.
+            size_t offset;          ///< Position of the command data in the draw list.
+        };
+
+        /// @brief Builds a UIDrawListEntry.
+        /// @ingroup ui-canvas-plugin
+        class UIDrawListEntryBuilder
```
RiscadoA, line 82: Same as above

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+            /// @param texIndex Index of the texture to add.
+            /// @param texture Texture to add.
+            /// @param sampler Sampler to be used with the texture.
+            /// @returns The builder.
+            UIDrawListEntryBuilder withTexture(int texIndex, cubos::core::gl::Texture2D texture,
+                                               cubos::core::gl::Sampler sampler = nullptr);
+
+            /// @brief Gets the built entry.
+            /// @return Built UIDrawListEntry.
+            UIDrawListEntry entry() const;
```
RiscadoA, line 99: Not needed anymore I think?

engine/include/cubos/engine/ui/canvas/element.hpp
```diff
@@ -20,24 +21,49 @@ namespace cubos::engine
...
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Untyped command data.
+        /// @return A builder for the created entry.
+        UIDrawList::UIDrawListEntryBuilder drawUntyped(const UIDrawList::UIDrawType& type,
+                                                       core::gl::VertexArray vertexArray, size_t vertexOffset,
+                                                       size_t vertexCount, const void* data);
+
+        /// @brief Adds new entry to the draw list, with a defined data type.
+        /// @param type Command type.
```
RiscadoA, line 56:
```suggestion
        /// @tparam T Command data type.
        /// @param type Command type.
```

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,122 @@
...
+    {
+        mDataCapacity += entry.type.perElementSize;
+        mDataCapacity *= 2;
+        void* temp = operator new(sizeof(char) * mDataCapacity);
+        std::memcpy(temp, mData, mDataSize);
+        operator delete(mData);
+        mData = temp;
+    }
+    std::memcpy(static_cast<char*>(mData) + mDataSize, data, entry.type.perElementSize);
+    mDataSize += entry.type.perElementSize;
```
RiscadoA, line 75: Can't you make this return a builder too? So that in the function below you can just call this one, to avoid duplicating this allocation logic.

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+        /// @brief Creates a new entry in the draw list.
+        /// @param type Command type.
+        /// @param vertexArray Vertex array handle.
+        /// @param vertexOffset Index of the first vertex to be drawn.
+        /// @param vertexCount Number of vertexes to draw.
+        /// @param data Command data.
+        /// @return A builder for the created entry.
+        UIDrawListEntryBuilder draw(const UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset,
+                                    size_t vertexCount, const void* data);
```
RiscadoA, line 132:
```suggestion
        UIDrawListEntryBuilder push(const UIDrawType& type, core::gl::VertexArray vertexArray, size_t vertexOffset,
                                    size_t vertexCount, const void* data);
```

I think you can leave it as `push` here, since this should just be a wrapper around the function above.

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+        .with<UICanvas>()
+        .other()
+        .with<UIElement>()
+        .related<ChildOf>()
+        .with<UIElement>()
+        .call([](Query<Entity, UIElement&, const UICanvas&> drawsToQuery,
+                 Query<Entity, UIElement&, UIElement&> childrenQuery,
+                 Query<Entity, UICanvas&, RenderTarget&> canvasQuery, const core::io::Window& window,
+                 const State& state) {
+            window->renderDevice().setDepthStencilState(state.dss);
```
RiscadoA, line 180:
```suggestion
            // For each render target with a UI canvas.
            window->renderDevice().setDepthStencilState(state.dss);
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+            for (auto [entity, canvas, renderTarget] : canvasQuery)
+            {
+                window->renderDevice().setViewport(0, 0, static_cast<int>(renderTarget.size.x),
```
RiscadoA, line 183:
```suggestion
                // Prepare which is common to all draw command types.
                window->renderDevice().setViewport(0, 0, static_cast<int>(renderTarget.size.x),
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                                                   static_cast<int>(renderTarget.size.y));
+                state.mvpBuffer->fill(&canvas.mat, sizeof(glm::mat4));
+                std::map<const UIDrawList::UIDrawType*, UIDrawList> commands;
```
RiscadoA, line 186:
```suggestion

                // Extract draw commands from all children UI elements, recursively.
                std::map<const UIDrawList::UIDrawType*, UIDrawList> commands;
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                for (auto [child, childElement, _] : drawsToQuery.pin(1, entity))
+                {
+                    copyCommands(childrenQuery, childElement, child, commands);
+                }
+                for (const auto& [type, list] : commands)
```
RiscadoA, line 191:
```suggestion
                // For each draw command type.
                for (const auto& [type, list] : commands)
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                {
+                    window->renderDevice().setShaderPipeline(type->pipeline);
```
RiscadoA, line 193:
```suggestion
                    // Prepare state common to all instances of this draw command type.
                    window->renderDevice().setShaderPipeline(type->pipeline);
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                    type->pipeline->getBindingPoint("MVP")->bind(state.mvpBuffer);
+                    type->constantBufferBindingPoint->bind(type->constantBuffer);
+
+                    for (size_t i = 0; i < list.size(); i++)
```
RiscadoA, line 197:
```suggestion
                    // For all instances of this draw command type.
                    for (size_t i = 0; i < list.size(); i++)
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                    {
+                        auto entry = list.entry(i);
```
RiscadoA, line 199:
```suggestion
                        // Prepare command-specific state and issue draw call.
                        auto entry = list.entry(i);
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -116,4 +163,54 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+                        type->constantBuffer->fill(list.data(i), type->perElementSize);
+                        window->renderDevice().drawTriangles(entry.command.vertexOffset, entry.command.vertexCount);
+
+                        for (size_t j = 0; j < UIDrawList::UIDrawType::MaxTextures; j++)
+                        {
+                            if (type->texBindingPoint[j] == nullptr)
+                                continue;
+                            type->texBindingPoint[j]->bind(entry.command.textures[j]);
+                            type->texBindingPoint[j]->bind(entry.command.samplers[j]);
+                        }
```
RiscadoA, line 211: This needs to be above the draw call.

engine/src/ui/color_rect/plugin.cpp
```diff
@@ -16,23 +17,19 @@ namespace
...
     {
-        ShaderPipeline pipeline;
         ShaderBindingPoint color;
         VertexArray va;
         VertexArrayDesc vad;
-        DepthStencilState dss;
         ConstantBuffer perElement;
-
+        UIDrawList::UIDrawType drawType;
         State(RenderDevice& renderDevice, const ShaderPipeline& pipeline)
```
RiscadoA, line 31:
```suggestion
        UIDrawList::UIDrawType drawType;
 
        State(RenderDevice& renderDevice, const ShaderPipeline& pipeline)
```
lemme breath

engine/include/cubos/engine/ui/canvas/canvas.hpp
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <map>
+
```
RiscadoA, line 8:
```suggestion

```
You don't need this here right?

engine/assets/ui/color_rect.fs
```diff
@@ -1,6 +1,13 @@
 #version 330 core
 
-uniform vec4 color;
+uniform PerElement
```
RiscadoA, line 3:
```suggestion
layout(std140) uniform PerElement
```
Without this, there's no guarantee that the fields will be in the order you expect (this might have been what was causing the red tint now that I think of it)

engine/assets/ui/element.vs
```diff
@@ -6,6 +6,7 @@ uniform PerElement
 {
```
RiscadoA, line 6: Also add the layout(std140) here

RiscadoA (changes requested):

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,158 @@
...
+        /// @brief Gets list size.
+        /// @return Number of elements in the list.
+        std::size_t size() const;
+
+        /// @brief Gets an entry from the list.
+        /// @param i Entry index.
+        /// @return A list entry.
+        const UIDrawListEntry& entry(std::size_t i) const;
+
+        /// @brief Dets the data of an entry.
```
tomas7770, line 146:
```suggestion
        /// @brief Gets the data of an entry.
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,144 @@
...
+            cubos::core::gl::VertexArray vertexArray;               ///< Vertex array handle.
+            size_t vertexOffset;                                    ///< Index of the first vertex to be drawn.
+            size_t vertexCount;                                     ///< Number of vertexes to draw.
+            cubos::core::gl::Sampler samplers[Type::MaxTextures];   ///< Samplers.
+            cubos::core::gl::Texture2D textures[Type::MaxTextures]; ///< Textures.
+        };
+
+        /// @brief Describes a draw commands and how it fits into the draw list carrying it.
+        /// @ingroup ui-canvas-plugin
+        struct Entry
```
RiscadoA, line 49:
```suggestion
        struct CUBOS_ENGINE_API Entry
```

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            /// @brief Constructs.
+            /// @param commandType Command type.
+            explicit Entry(const Type& commandType)
+                : type(commandType)
+            {
+            }
```
RiscadoA, line 56: Move to .cpp

engine/include/cubos/engine/ui/canvas/draw_list.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include <vector>
+
+#include <cubos/core/gl/render_device.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Holds a collections of draw commands and their data.
+    /// @ingroup ui-canvas-plugin
```
RiscadoA, line 15: Since this part of the code is a bit confusing to people who are not familiar with it, we should explain it a lil more in depth.
I think you can explain the overall mechanism in the plugin page, and just a link to it here.
```suggestion
    /// @brief Holds a collections of draw commands and their data.
    /// @note See the @ref ui-canvas-plugin documentation page for more information.
    /// @ingroup ui-canvas-plugin
```

engine/include/cubos/engine/ui/canvas/plugin.hpp
RiscadoA, line 1: As I wrote in the draw list comment, I think it's a good idea to a short overview of this plugin's architecture in its documentation page. You can do this by just continuing writing on the plugins documentation block:

```
/// @defgroup ui-canvas-plugin Canvas
/// @ingroup engine
/// @brief Adds UI canvas and elements.
///
/// # Architecture
///
/// Sample text
/// more sample text
///
/// even more sample text
///
/// @note cool note
///
/// etc
///
/// Check out @ref sample-name! or something
```

The main points that should be here are:
- each UI element holds a draw list
- what is a draw list, what is a draw command, and a draw command type
- that the canvas aggregates those commands and draws them in a single step
- link to the color rect plugin for an example of basic usage

engine/include/cubos/engine/ui/canvas/plugin.hpp
```diff
@@ -30,6 +30,7 @@ namespace cubos::engine
     /// @brief Systems which draw to canvas should be tagged with this.
     /// @ingroup canvas-target-plugin
     CUBOS_ENGINE_API extern Tag uiDrawTag;
+    CUBOS_ENGINE_API extern Tag uiEndTag;
```
RiscadoA, line 33:
```suggestion

    /// @brief All queued draw commands are collected and executed to the render targets.
    CUBOS_ENGINE_API extern Tag uiEndTag;
```

engine/src/ui/canvas/plugin.cpp
```diff
@@ -30,20 +63,36 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
...
+            dssd.depth.compare = Compare::LEqual;
+            cmds.emplaceResource<State>(window->renderDevice().createDepthStencilState(dssd),
+                                        window->renderDevice().createConstantBuffer(sizeof(glm::mat4), nullptr,
+                                                                                    cubos::core::gl::Usage::Dynamic));
+        });
+
     cubos.system("set element matrix")
-        .tagged(elementVPUpdateTag)
+        .tagged(uiElementVPUpdateTag)
         .with<UIElement>()
```
RiscadoA, line 96: There's no longer a VP matrix in UI element, this system should be renamed.
The same applies to the tag.

RiscadoA (changes requested):

engine/include/cubos/engine/ui/canvas/plugin.hpp
```diff
@@ -13,15 +13,31 @@ namespace cubos::engine
...
+    ///
+    /// # Architecture
+    ///
+    /// The two main UI components are @ref UIElement and @ref UICanvas.
+    /// For it to work properly, each UIElement must be a child of a UICanvas, or of another UIElement.
+    ///
+    /// Each UIElement generates a list of drawing commands, @ref UIDrawList, that are then pooled together by the
+    /// UICanvas at the root of the hierarchy.
+    ///
+    /// Each entry in a draw list contains a @ref Type and a @ref Command. The Type hold information that is common for
```
RiscadoA, line 26:
```suggestion
    /// Each entry in a draw list contains a @ref UIDrawList::Type and a @ref UIDrawList::Command. The Type holds information that is common for
```

engine/src/ui/canvas/draw_list.cpp
```diff
@@ -0,0 +1,123 @@
...
+    , mDataSize(other.mDataSize)
+{
+    other.mData = nullptr;
+    other.mDataSize = 0;
+    other.mDataCapacity = 0;
+}
+
+cubos::engine::UIDrawList::UIDrawList(const UIDrawList& other) noexcept
+    : mEntries(other.mEntries)
+    , mDataCapacity(other.mDataCapacity)
```
RiscadoA, line 63:
```suggestion
    , mDataCapacity(other.mDataSize)
```

engine/src/ui/color_rect/plugin.cpp
```diff
@@ -16,23 +17,19 @@ namespace
...
     };
 
     struct State
     {
-        ShaderPipeline pipeline;
-        ShaderBindingPoint color;
         VertexArray va;
         VertexArrayDesc vad;
-        DepthStencilState dss;
         ConstantBuffer perElement;
```
RiscadoA, line 28:
```suggestion
```
This isn't needed anymore right? It is already part of the draw type.

engine/src/ui/color_rect/plugin.cpp
```diff
@@ -16,23 +17,19 @@ namespace
...
+        glm::vec4 color;
         int depth;
     };
 
     struct State
     {
-        ShaderPipeline pipeline;
-        ShaderBindingPoint color;
         VertexArray va;
         VertexArrayDesc vad;
```
RiscadoA, line 27:
```suggestion
```
No need to store this, you only use this in the constructor

engine/src/ui/color_rect/plugin.cpp
```diff
@@ -88,28 +84,18 @@ void cubos::engine::colorRectPlugin(Cubos& cubos)
...
-                PerElement perElement{{minX, maxX}, {minY, maxY}, element.layer * 100 + element.hierarchyDepth};
-                state.perElement->fill(&perElement, sizeof(perElement));
-                state.pipeline->getBindingPoint("PerElement")->bind(state.perElement);
-
-                auto mvpBuffer = window->renderDevice().createConstantBuffer(sizeof(glm::mat4), &element.vp,
-                                                                             cubos::core::gl::Usage::Static);
-                state.pipeline->getBindingPoint("MVP")->bind(mvpBuffer);
-                state.color->setConstant(colorRect.color);
-                window->renderDevice().drawTriangles(0, 6);
+                element.draw<PerElement>(
```
RiscadoA, line 95:
```suggestion
                element.draw(
```
Isn't the compiler able to infer the type?

engine/src/ui/color_rect/plugin.cpp
```diff
@@ -88,28 +84,18 @@ void cubos::engine::colorRectPlugin(Cubos& cubos)
...
-                                               static_cast<int>(window->framebufferSize()[1]));
-            window->renderDevice().setVertexArray(state.va);
-            window->renderDevice().setShaderPipeline(state.pipeline);
+        .call([](State& state, Query<UIElement&, const UIColorRect&> query) {
             for (auto [element, colorRect] : query)
             {
                 float minX = element.position.x - element.size.x * (element.pivot.x);
                 float minY = element.position.y - element.size.y * (element.pivot.y);
                 float maxX = element.position.x + element.size.x * (1.0F - element.pivot.x);
                 float maxY = element.position.y + element.size.y * (1.0F - element.pivot.y);
```
RiscadoA, line 93:
```suggestion
                glm::vec2 min = element.position - element.size * element.pivot;
                glm::vec2 max = element.position + element.size * (glm::vec2(1.0F, 1.0F) - element.pivot);
```

RiscadoA (approved):
That's all from me :relieved:

tomas7770 (approved):

roby2014 (approved):

RiscadoA:
Feel free to merge @DiogoMendonc-a! The CI is failing but it isn't your fault; lets hope it starts working on other PRs.

PR #1203, opened 2024-05-27
Switch to new logo and use Cubos instead of CUBOS.

RiscadoA:
# Description

Changed the README and the docs to use the new logos.
Rename every occurrence of CUBOS./CUBOS by Cubos.


RiscadoA:
YOLO - merging this as is, it doesn't really need a review

PR #1222, opened 2024-06-04
Allow importing plugins from shared libraries

RiscadoA:
# Description

This is the first step towards having a standalone editor application!
This PR just adds a new class, `DynamicPlugin`, which is used to load shared libraries and import plugins from them, in a cross-platform manner.

For example, you could do something like:
```cpp
DynamicPlugin plugin{};
plugin.load("game.dll"); // would have a different name in linux, probably "game.so" or smth
cubos.plugin(plugin.function());
```

This isn't meant to be used directly by the user - it will be mostly used internally by the Tesseratos application.

Also included a small commit adding some missing `@ingroup` tags.

## Checklist

- [x] Self-review changes.
- [x] ~~Ensure test coverage.~~ Cumbersome to do it in a cross-platform way; I think we can leave this to a later stage.
- [x] ~~Write new samples.~~ Same as above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Start and update methods to the Cubos class (#1213, **@RiscadoA**).
 - Reset method to the Cubos class (#1225, **@RiscadoA**).
 - Plugin injection to the Cubos class (#1214, **@RiscadoA**).
+- Utility class used to load plugins from shared libraries (#1035, **@RiscadoA**).
 
 ## [v0.2.0] - 2024-05-07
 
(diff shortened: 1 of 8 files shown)


diogomsmiranda:
Is this supposed to have no test coverage?

RiscadoA:
> Is this supposed to have no test coverage?

Take a look at the checklist, I added a note there.

diogomsmiranda:
> > Is this supposed to have no test coverage?
> 
> Take a look at the checklist, I added a note there. Will probably be easier to cover it on an upper layer.

Got it!! thanks

diogomsmiranda (approved):

Dacops (approved):
lgtm

teres4 (approved):

PR #1227, opened 2024-06-04
Make Tesseratos an executable

RiscadoA:
# Description

This PR looks really big but there wasn't really any other way to tackle this.
It simply moves stuff around so that Tesseratos becomes an executable, instead of a library.
Thus, this PR:
- removes the Tesseratos samples directory;
- replaces the base Tesseratos plugin.cpp by a main.cpp with a main function.
- moves all headers in the `include` directory to their respective folders in the `src` directory.

## Checklist

- [x] Self-review changes.
- [x] ~~Write new samples.~~
- [x] ~~Add entry to the changelog's unreleased section.~~


tools/tesseratos/CMakeLists.txt
```diff
@@ -22,12 +20,7 @@ set(TESSERATOS_SOURCE
 	"src/tesseratos/ecs_statistics/plugin.cpp"
 )
 
-add_library(tesseratos ${TESSERATOS_SOURCE})
-target_include_directories(tesseratos PUBLIC "include")
```
DiogoMendonc-a, line 26: Shouldn't we want the individual tools to be able to be added to the game itself? A lot of games have a debug mode that toggles a lot of developer tools. It might be useful for devs if they can test stuff in the build itself while still having access to some of the tools.

DiogoMendonc-a (commented):
Shouldn't we want the individual tools to be able to be added to the game itself? A lot of games have a debug mode that toggles a lot of developer tools. It might be useful for devs if they can test stuff in the build itself while still having access to some of the tools.

tools/tesseratos/CMakeLists.txt
RiscadoA: Agree, I think we should gradually move stuff to maybe a `engine/tools` directory, but that would be a big change to include in this PR. We need to decide which of tools we want to include.

DiogoMendonc-a (approved):

Scarface1809 (approved):

teres4 (approved):

PR #1235, opened 2024-06-05
Add a Menu Bar to Tesseratos

RiscadoA:
# Description

Adds a menu bar plugin to Tesseratos. This plugin acts somewhat like the Toolbox plugin, except it interacts directly with the tools themselves.

## Checklist

- [x] Self-review changes.
- [x] ~~Write new samples.~~

### DIFF
--- a/tools/tesseratos/src/tesseratos/main.cpp
+++ b/tools/tesseratos/src/tesseratos/main.cpp
@@ -6,16 +6,9 @@
 #include <cubos/engine/settings/plugin.hpp>
 #include <cubos/engine/settings/settings.hpp>
 #include <cubos/engine/tools/plugin.hpp>
-#include <cubos/engine/tools/selection/plugin.hpp>
 #include <cubos/engine/utils/free_camera/plugin.hpp>
 
 #include "menu_bar/plugin.hpp"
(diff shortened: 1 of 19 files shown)


joaomanita (approved):

tools/tesseratos/src/tesseratos/main.cpp
```diff
@@ -68,5 +71,20 @@ int main(int argc, char** argv)
             input.bind(*bindings);
         });
 
+    cubos.startupSystem("add some menu items just for testing purposes").call([](Commands cmds) {
```
diogomsmiranda, line 74: I would maybe change the name of this system besides that lgtm

diogomsmiranda (approved):

tools/tesseratos/src/tesseratos/main.cpp
RiscadoA: It's not really important, this is code that will be erased as soon as actual menus are added. What do you suggest though?

diogomsmiranda: It's not a deal breaker to me, it can stay the same. I just thought it was something that went unnoticed

PR #1257, opened 2024-06-09
Add network UdpSocket class

roby2014:
blocked by #1258 

# Description

Add a UdpSocket class with the implementation proposed on the issue.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] Write new samples.


core/include/cubos/core/net/udp_socket.hpp
```diff
@@ -0,0 +1,79 @@
...
+    public:
+        /// @brief Represents a result code.
+        enum class ResultCode
+        {
+            Success,
+            BindError,
+            SendError,
+            ReceiveError,
+            UnknownError
+        };
```
RiscadoA, line 29: Since you basically only have one result code for each operation, you might as well remove this enum and just return a boolean in each method, indicating whether the operation was successful or not.

core/include/cubos/core/net/udp_socket.hpp
```diff
@@ -0,0 +1,79 @@
...
+        /// @param size Maximum buffer size.
+        /// @param[out] received Number of bytes read.
+        /// @param[out] address Origin address.
+        /// @return Result code.
+        ResultCode receive(char* data, size_t size, int& received, Address& address) const;
+
+        /// @brief Control `blocking` mode of inner socket.
+        /// @param blocking Whether to block or not.
+        /// @note A blocking socket does not return control until it has sent (or received) some or all data specified
+        /// for the operation
```
RiscadoA, line 71:
```suggestion
        /// for the operation.
```

RiscadoA (changes requested):
Changed it to draft again as it's still depending on the other PR, no need to get reviews from everyone yet.

core/include/cubos/core/net/udp_socket.hpp
roby2014: Ok. Just wanted a review so I can start implementing the other network utilities (`TcpSocket`, etc..) which should be similar to this one.

EdSwordsmith (approved):

core/include/cubos/core/net/udp_socket.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+
+#include <cubos/core/api.hpp>
+#include <cubos/core/net/address.hpp>
+
+#ifdef _WIN32
+#include <vcruntime.h>
+#include <winsock2.h>
```
RiscadoA, line 16: Are you sure you need to include this here? I'm afraid of macro troubles down the line.
Is the type of `SOCKET` always the same? If so, you can just use that type down below.
Also, if you change this, avoid the `#define`s. Having them on the .cpp is fine though.

core/src/net/udp_socket.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+bool UdpSocket::bind(uint16_t port, const Address& address)
+{
+    mSock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (mSock == INVALID_SOCKET)
+    {
+        CUBOS_ERROR("Failed to allocate UDP socket: {}", error());
+        return false;
+    }
```
RiscadoA, line 59: This should be done on the constructor, as for clients, you don't need to call `bind`, you can just `send`.

core/samples/net/udp/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+        CUBOS_ERROR("Receive error");
+    }
+}
+/// [Server function]
+
+/// [Client function]
+void runClient()
+{
+    UdpSocket client;
+    CUBOS_ASSERT(client.bind(CLIENT_PORT, CLIENT_ADDRESS), "Failed to bind client socket");
```
RiscadoA, line 52: This line should not be necessary. You don't need to bind a UDP socket to send a packet, you only need to bind it to receive one.

core/samples/net/udp/page.md
RiscadoA, line 1: Ty for maintaining the docs :pray: Don't forget to link this in `page.md` files in directories above, otherwise we won't get a nice index in the website. In this case, you need a `net/page.md`, and there you can link this page. Check out other modules for examples.

core/samples/net/udp/page.md
```diff
@@ -0,0 +1,35 @@
+# Udp Sockets {#examples-core-udp-sockets}
```
RiscadoA, line 1:
```suggestion
# Udp Sockets {#examples-core-net-udp-sockets}
```

core/src/net/udp_socket.cpp
```diff
@@ -0,0 +1,186 @@
...
+#else
+    auto res = sendto(mSock, data, size, 0, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr));
+#endif
+    if (res == INVALID_SOCKET)
+    {
+        CUBOS_ERROR("Failed to send UDP data: {}", error());
+        return false;
+    }
+
+    CUBOS_INFO("Sent UDP message to socket '{}' at address {} and port '{}'", mSock, address.toString(), port);
```
RiscadoA, line 118: This will be very spammy, I would keep it `CUBOS_TRACE` if you find it necessary.

core/src/net/udp_socket.cpp
```diff
@@ -0,0 +1,186 @@
...
+    if (len == INVALID_SOCKET)
+    {
+        CUBOS_ERROR("Failed to receive UDP data: {}", error());
+        return false;
+    }
+
+    received = static_cast<std::size_t>(len);
+    address = Address::from(std::string(inet_ntoa(addr.sin_addr))).value();
+
+    CUBOS_INFO("Received UDP message at socket '{}' from address '{}'", mSock, address.toString());
```
RiscadoA, line 149: Same as above.

core/src/net/udp_socket.cpp
```diff
@@ -0,0 +1,186 @@
...
+    addr.sin_addr.s_addr = inet_addr(address.toString().c_str());
+
+    if (::bind(mSock, reinterpret_cast<struct sockaddr*>(&addr), sizeof(addr)) == -1)
+    {
+        CUBOS_ERROR("Failed to bind UDP socket: {}", error());
+        unbind();
+        return false;
+    }
+
+    CUBOS_INFO("Bind UDP socket '{}' at address {} with port '{}'", mSock, address.toString(), port);
```
RiscadoA, line 75:
```suggestion
    CUBOS_DEBUG("Bind UDP socket '{}' at address {} with port '{}'", mSock, address.toString(), port);
```

core/src/net/udp_socket.cpp
```diff
@@ -0,0 +1,186 @@
...
+    success = fcntl(mSock, F_SETFL, flags) == 0;
+#endif
+    if (!success)
+    {
+        CUBOS_ERROR("Failed to change block flag of socket '{}'", mSock);
+    }
+    else
+    {
+        CUBOS_INFO("UDP socket {} is now '{}'", mSock, blocking ? "blocking" : "non-blocking");
+    }
```
RiscadoA, line 183: This shouldn't be necessary, as this is part of the core functionality and users should just expect it to work.

core/tests/net/udp_socket.cpp
```diff
@@ -0,0 +1,75 @@
...
+        REQUIRE(recvAddress.toString() == sendAddress.toString());
+        REQUIRE(*buffer == *msg);
+    }
+
+    SUBCASE("blocking")
+    {
+        UdpSocket socket;
+        REQUIRE(!socket.setBlocking(false));
+        REQUIRE(socket.bind(8080));
+        REQUIRE(socket.setBlocking(false));
```
RiscadoA, line 57: Why is this expected behavior? Shouldn't the order not matter?

core/tests/net/udp_socket.cpp
```diff
@@ -0,0 +1,75 @@
...
+    using cubos::core::net::Address;
+    using cubos::core::net::UdpSocket;
+
+    SUBCASE("bind")
+    {
+        UdpSocket socket;
+        REQUIRE(socket.bind(8080));
+    }
+
+    SUBCASE("double bind")
```
RiscadoA, line 22: What about binding the same socket twice? What happens?

RiscadoA (changes requested):
Thanks for working on this! :tada: this will be very useful, good to finally have some networking utils on Cubos, good work!

roby2014:
last quick review @RiscadoA 🙏🏻

core/include/cubos/core/net/udp_socket.hpp
```diff
@@ -0,0 +1,72 @@
...
+        /// @return Whether data was sent sucessfully.
+        bool send(const void* data, std::size_t size, const Address& address, uint16_t port) const;
+
+        /// @brief Receives a single datagram message on the socket.
+        /// @param data Buffer to store the datagram message.
+        /// @param size Maximum buffer size.
+        /// @param[out] received Number of bytes read.
+        /// @param[out] address Origin address.
+        /// @return Whether data was received sucessfully.
+        /// @note Depending on the OS, this method may not block. See @ref setBlocking.
```
RiscadoA, line 57: Depending on the OS? The behavior should be consistent in all implementations :thinking: are you saying that different OSes have their default sockets set to different blocking states? In that case, we should add a call to the `.setBlocking(true)` in the constructor.

RiscadoA (changes requested):

core/include/cubos/core/memory/stream.hpp
```diff
@@ -58,12 +58,26 @@ namespace cubos::core::memory
         /// @return Number of bytes read.
         virtual std::size_t read(void* data, std::size_t size) = 0;
 
+        /// @brief Reads data from the stream with the exact specified size.
+        /// @param data Buffer to read data into.
+        /// @param size Size of the buffer.
+        /// @note This function will only return after buffer is completed.
+        /// @return Whether reading was sucessful.
+        bool readExact(void* data, std::size_t size);
```
diogomsmiranda, line 66: I'm having a tough time understanding differences between this method and the 

`         virtual std::size_t read(void* data, std::size_t size) = 0;`

Besides what it returns what changes?

core/include/cubos/core/memory/stream.hpp
```diff
@@ -58,12 +58,26 @@ namespace cubos::core::memory
...
         /// @param size Size of the buffer.
         /// @return Number of bytes written.
         virtual std::size_t write(const void* data, std::size_t size) = 0;
 
+        /// @brief Writes data to the stream with the exact specified size.
+        /// @param data Buffer to write data from.
+        /// @param size Size of the buffer.
+        /// @note This function will only return after whole buffer is written.
+        /// @return Whether writing was sucessful.
+        bool writeExact(const void* data, std::size_t size);
```
diogomsmiranda, line 79: Same for here.

diogomsmiranda (approved):

diogomsmiranda:
Approved this PR supposing changes requested by @RiscadoA are addressed.

core/include/cubos/core/memory/stream.hpp
roby2014: It was because of this;:

image

but I am removing it, since Ive talked with @RiscadoA and other Stream implementations are assuming that `read` handles the whole buffer.

core/samples/net/page.md
```diff
@@ -0,0 +1,6 @@
+# Networking {#examples-net-networking}
```
RiscadoA, line 1:
```suggestion
# Networking {#examples-core-networking}
```

core/samples/net/page.md
```diff
@@ -0,0 +1,6 @@
...
+
+@brief Using the @ref core-net module.
+
+- @subpage examples-core-net-udp-sockets - @copybrief examples-core-net-udp-sockets
+- @subpage examples-core-net-tcp-sockets - @copybrief examples-core-net-tcp-sockets
```
RiscadoA, line 6: In the TCP sample `page.md` you used `examples-core-tcp-sockets`, make sure to change it to `examples-core-net-tcp-sockets` otherwise the link won't work.

docs/pages/3_examples/1_core/main.md
```diff
@@ -11,3 +11,4 @@ The following examples have fully documented tutorials:
 - @subpage examples-core-logging - @copybrief examples-core-logging
```
RiscadoA, line 11: Can you also add the metrics page here? Just noticed it is also missing

RiscadoA (approved):
:rocket:

PR #1258, opened 2024-06-09
Add network Tcp utilities

roby2014:
blocked by #1256 

# Description

Implemented network TcpStream and TcpListener classes.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] Write new samples.


core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <cubos/core/reflection/external/primitives.hpp>
+#include <cubos/core/reflection/external/string.hpp>
+
+using cubos::core::net::Address;
+using cubos::core::net::TcpListener;
+using cubos::core::net::TcpStream;
+
+/// @brief Cross-platform utility to get last error message.
+/// @return Message.
+/// @todo Move this to a higher module for general purposes?
```
RiscadoA, line 30: Yeah that might be a good idea, maybe a `logSystemError()` function in `log.hpp`?

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    int clientSocket = ::accept(mSock, nullptr, nullptr);
+    if (clientSocket == INVALID_SOCKET)
+    {
+        CUBOS_ERROR("Failed to accept TCP connection: {}", error());
+        return false;
+    }
+
+    CUBOS_INFO("Connecting stream to socket");
+    stream.connect(clientSocket);
```
RiscadoA, line 98: H'm I don't like this 'connect' wording, because you're not connecting the stream to a socket, you're just replacing the socket of the stream with a connected one, right?
I think you could just add a `.inner(socket)` method which replaces the inner socket with the given one, destroying the old one.

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,183 @@
...
+    CUBOS_INFO("New TCP stream at address {} and port '{}'", address.toString(), port);
+    return true;
+}
+
+bool TcpStream::connect(SOCKET socket)
+{
+    mSock = socket;
+    CUBOS_INFO("New TCP stream at socket {}", mSock);
+    return mSock != INVALID_SOCKET;
+}
```
RiscadoA, line 71: As I said above, I would prefer this to be renamed to `inner(SOCKET socket)`, as it is a setter.
Also, don't forged to close the old `mSock`, if it wasn't -1.

Scarface1809 (approved):

EdSwordsmith (approved):

core/include/cubos/core/log.hpp
```diff
@@ -375,6 +375,11 @@ namespace cubos::core
         static const char* streamFormat(memory::Stream& stream, const char* format, const reflection::Type& type,
                                         const void* value);
     };
+
+    /// @brief Cross-platform utility to get last error message.
+    /// @return Message.
+    /// @todo Move this to a higher module for general purposes?
```
RiscadoA, line 381:
```suggestion
```

core/include/cubos/core/log.hpp
```diff
@@ -375,6 +375,11 @@ namespace cubos::core
...
+    std::string logSystemError();
```
RiscadoA, line 382: Hm, maybe you should rebase the UDP PR on this one and merge it only after this one is merged, as it would also benefit from this function.

core/include/cubos/core/net/tcp_listener.hpp
```diff
@@ -0,0 +1,72 @@
...
+#ifdef _WIN32
+#include <winsock2.h>
+#pragma comment(lib, "Ws2_32.lib")
+#else
+#include <sys/socket.h>
+// Socket type. Differs from Windows (SOCKET) and POSIX (int).
+#define SOCKET int
+// Represents an invalid socket.
+#define INVALID_SOCKET (-1)
+#endif
```
RiscadoA, line 20: Are you sure we need to include this in the .hpp? Read my similar comment in #1257.

core/include/cubos/core/net/tcp_listener.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+        /// @brief Close the listener socket.
+        void close();
+
+        /// @brief Check if connection is valid.
+        /// @return Whether inner sock is connected.
+        inline bool connected() const
+        {
+            return mSock != INVALID_SOCKET;
+        }
```
RiscadoA, line 60: Unsure `connected` is the right work here. Maybe `valid`? As there isn't really a connection.

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        char peek() const override;
+
+        void seek(ptrdiff_t offset, memory::SeekOrigin origin) override
+        {
+            (void)offset;
+            (void)origin;
+            CUBOS_FAIL("Not implemented");
+        };
```
RiscadoA, line 84:
```suggestion
        }
```

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#ifdef _WIN32
+#include <winsock2.h>
+#pragma comment(lib, "Ws2_32.lib")
+#else
+// Socket type. Differs from Windows (SOCKET) and POSIX (int).
+#define SOCKET int
+// Represents an invalid socket.
+#define INVALID_SOCKET (-1)
+#endif
```
RiscadoA, line 15: Same as above and as in #1257.

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+        /// @brief Connects to a TCP listener, possibly on a remote machine.
+        /// @param address Destination address.
+        /// @param port Destination port.
+        /// @return Whether connecting was successful.
+        bool connect(const Address& address, uint16_t port);
+
+        /// @brief Connects to an existing connection, replacing the inner socket.
+        /// @param socket Socket.
+        /// @return Whether connecting was successful.
```
RiscadoA, line 48:
```suggestion
        /// @brief Replaces the inner socket with the given one.
        /// @param socket Socket.
        /// @return Whether connecting was successful.
```

core/src/log.cpp
```diff
@@ -263,4 +268,20 @@ const char* Logger::streamFormat(memory::Stream& stream, const char* format, con
     }
 
     return format;
+}
+
+std::string cubos::core::logSystemError()
+{
+#ifdef _WIN32
+    static char message[256] = {0};
+    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, 0, WSAGetLastError(), 0, message, 256, 0);
```
RiscadoA, line 277: `WSAGetLastError` is specific to `winsock2`, so it doesn't make sense after all for this to be in `log.hpp` :pensive: maybe just roll back the change then? You can also just add this utility in a private header in `core/src/net/log.hpp` or smth.

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,95 @@
...
+        return false;
+    }
+
+    CUBOS_INFO("TCP listener socket '{}' at address {} with port '{}'", mSock, address.toString(), port);
+    return true;
+}
+
+bool TcpListener::accept(TcpStream& stream) const
+{
+    auto clientSocket = ::accept(mSock, nullptr, nullptr);
```
RiscadoA, line 70:
```suggestion
    CUBOS_ASSERT(mSock != INVALID_SOCKET);
  
    auto clientSocket = ::accept(mSock, nullptr, nullptr);
```

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+TcpListener::~TcpListener()
+{
+    close();
+}
+
+bool TcpListener::listen(const Address& address, uint16_t port, int maxQueueLength)
+{
+    mSock = socket(AF_INET, SOCK_STREAM, 0);
```
RiscadoA, line 40: What if `listen` is called twice? The previous `mSock` should be destroyed.

core/tests/net/tcp.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+        std::thread serverThread{[&svAddr, &svPort]() {
+            TcpListener listener;
+            REQUIRE(listener.listen(svAddr, svPort, 1));
+
+            TcpStream stream;
+            REQUIRE(listener.accept(stream));
+
+            char readBuffer[4 + 1];
+            REQUIRE(stream.read(readBuffer, sizeof(readBuffer)) > 0);
```
RiscadoA, line 29: Two issues here:
- you're reading 5 bytes, `sizeof(readBuffer)`, while only 4 are being sent. 
- sockets reads are not guaranteed to return the entire requested data (I might be wrong though, if you can, check if that's really the case for blocking sockets). You should repeat `read` calls until all of the buffer has been received. Btw, this is usually necessary because if a signal is triggered, or the OS steals the app context, your read is interrupted. If you're unlucky with the timing, you might read only part of the data.

core/tests/net/tcp.cpp
```diff
@@ -0,0 +1,98 @@
...
+        TcpStream stream;
+        REQUIRE(stream.connect(svAddr, svPort));
+
+        const char* msg = "ping";
+        REQUIRE(stream.write(msg, std::strlen(msg)) > 0);
+        REQUIRE(stream.tell() == std::strlen(msg));
+
+        char readBuffer[4 + 1];
+        readBuffer[4] = '\0';
+        REQUIRE(stream.read(readBuffer, 4) > 0);
```
RiscadoA, line 52: Same as above!

core/tests/net/tcp.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+            TcpStream stream;
+            REQUIRE(listener.accept(stream));
+
+            char readBuffer[4 + 1];
+            REQUIRE(stream.read(readBuffer, sizeof(readBuffer)) > 0);
+            readBuffer[4] = '\0';
+            REQUIRE(std::string(readBuffer) == "ping");
+
+            REQUIRE(stream.write("pong", 4) > 0);
```
RiscadoA, line 33: Same as above, I think you also need to repeat this call until everything has been written.

core/tests/net/tcp.cpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+        TcpStream stream;
+        REQUIRE(!stream.inner(INVALID_SOCKET));
+        REQUIRE(!stream.connected());
+    }
+
+    SUBCASE("stream disconnect")
+    {
+        TcpStream stream;
+        stream.inner(1336);
```
RiscadoA, line 86:
```suggestion
        stream.inner(1336); // dummy value, not a real socket
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+bool TcpStream::inner(SOCKET socket)
+{
+    disconnect();
+
+    mSock = socket;
+    CUBOS_INFO("New TCP stream at socket {}", mSock);
+    return mSock != INVALID_SOCKET;
+}
```
RiscadoA, line 53: Why check if it is invalid here? I think this function should return `void`. The user is responsible for whatever they throw inside it, its not necessary to do the check for them.

RiscadoA (changes requested):
Ty for working on this! Basically I added similar comments to what I added to the UDP PR. Probably best to start there.

tomas7770 (approved):

core/src/log.cpp
roby2014: Can I add a `utils.hpp` for stuff like `SOCKET` and this function? Or how should I organize this

RiscadoA: Sure yeah, you can add a `net/utils.hpp`.

core/tests/net/tcp.cpp
roby2014: - How should I handle the null termination byte? Declare a buffer of 5 bytes and read 4, append 0 to the last one?
- Is this the client's job to do? Or should we do it in `TcpStream::read`

RiscadoA: - Yes, that's way to do it.
- Yes it is, although it becomes a bit repetitive, but that's the `Stream::read` interface. If you want to simplify the user code, we can add a `Stream::readAtLeast` (if you have a better name idea feel free to pick something else), that calls `Stream::read` until the requested bytes are received.

roby2014: `readExact` sounds good?

RiscadoA: LGTM

roby2014:
last quick review @RiscadoA 🙏🏻

core/include/cubos/core/memory/stream.hpp
```diff
@@ -58,12 +58,26 @@ namespace cubos::core::memory
         /// @return Number of bytes read.
         virtual std::size_t read(void* data, std::size_t size) = 0;
 
+        /// @brief Reads data from the stream with the exact specified size.
+        /// @param data Buffer to read data into.
+        /// @param size Size of the buffer.
+        /// @note This function will only return after buffer is completed.
```
RiscadoA, line 64:
```suggestion
        /// @note This function will only return after buffer is completed, or an error occurs.
```

core/include/cubos/core/memory/stream.hpp
```diff
@@ -58,12 +58,26 @@ namespace cubos::core::memory
...
         /// @brief Writes data to the stream.
         /// @param data Buffer to write data from.
         /// @param size Size of the buffer.
         /// @return Number of bytes written.
         virtual std::size_t write(const void* data, std::size_t size) = 0;
 
+        /// @brief Writes data to the stream with the exact specified size.
+        /// @param data Buffer to write data from.
+        /// @param size Size of the buffer.
+        /// @note This function will only return after whole buffer is written.
```
RiscadoA, line 77:
```suggestion
        /// @note This function will only return after whole buffer is written, or an error occurs.
```

core/include/cubos/core/net/tcp_listener.hpp
```diff
@@ -0,0 +1,62 @@
...
+#pragma once
+
+#include <cubos/core/api.hpp>
+#include <cubos/core/net/address.hpp>
+#include <cubos/core/net/tcp_stream.hpp>
+#include <cubos/core/net/utils.hpp>
+
+namespace cubos::core::net
+{
+    class CUBOS_CORE_API TcpListener final
```
RiscadoA, line 14: Missing documentation.

core/include/cubos/core/net/utils.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
RiscadoA, line 1: Missing file docs

core/include/cubos/core/net/utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <sys/socket.h>
+#endif
+
+#include <string>
+
+// Socket type. Differs from Windows (SOCKET) and POSIX (int).
+using InnerSocket = decltype(socket(0, 0, 0));
+
+// Represents an invalid socket.
+constexpr InnerSocket InnerInvalidSocket = static_cast<InnerSocket>(-1);
```
RiscadoA, line 15: Should both be documented, and marked as being for internal use. They should also be within the namespace.

core/include/cubos/core/net/utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace cubos::core::net
+{
+    /// @brief Cross-platform utility to get last error message.
```
RiscadoA, line 19:
```suggestion
    /// @brief Cross-platform utility to get last networking error message.
```

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,82 @@
...
+            return mEof;
+        }
+
+        char peek() const override;
+
+        void seek(ptrdiff_t offset, memory::SeekOrigin origin) override
+        {
+            (void)offset;
+            (void)origin;
+            CUBOS_FAIL("Not implemented");
```
RiscadoA, line 74: You need to include `log.hpp` for this. To avoid including it in the header, I would suggest moving this method's definition to `tcp_stream.cpp`.

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,82 @@
...
+        std::size_t read(void* buffer, std::size_t size) override;
+
+        std::size_t write(const void* buffer, std::size_t size) override;
+
+        inline std::size_t tell() const override
+        {
+            return mPos;
+        }
+
+        inline bool eof() const override
```
RiscadoA, line 63: Inline is redundant, you can remove the keyword - it is automatically added by the compiler if you're defining a method in the class definition itself.

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,82 @@
...
+        /// @param socket Socket.
+        /// @return Whether connecting was successful.
+        void inner(InnerSocket socket);
+
+        /// @brief Disconnects inner socket.
+        void disconnect();
+
+        /// @brief Checks if inner socket is connected.
+        /// @return Whether the buffer holds a valid connection.
+        inline bool valid() const
```
RiscadoA, line 47: Redundant inline keyword

core/samples/net/tcp/main.cpp
```diff
@@ -0,0 +1,67 @@
...
+    TcpListener listener;
+    listener.listen(SERVER_ADDRESS, SERVER_PORT, 1);
+
+    CUBOS_DEBUG("Server is listening");
+
+    TcpStream clientStream;
+    while (listener.accept(clientStream))
+    {
+        char buffer[1024];
+        if (std::size_t bytesRead = clientStream.read(buffer, sizeof(buffer)); bytesRead > 0)
```
RiscadoA, line 34: There's no guarantee that you will read the entire message at once. You should repeat read until `bytesRead = 0`.

core/samples/net/tcp/page.md
RiscadoA, line 1: You need to add a `net/page.md` (you can copy `reflection/page.md`), and add a link to that sample here.
Additionally, you should also add a link to `net/page.md` on `docs/pages/3_examples/1_core/main.md`

Check out the generated docs page index to see why this is a problem: https://docs.cubosengine.org/preview/pr-1258/pages.html
The TCP sample appears outside of the hierarchy, as it doesn't have a 'parent page'.

core/src/memory/stream.cpp
```diff
@@ -22,6 +22,44 @@ Stream& Stream::stdIn = stdInI;
...
+        totalBytesRead += bytesRead;
+    }
+
+    return true;
+}
+
+bool Stream::writeExact(const void* data, std::size_t size)
+{
+    const char* dataPtr = static_cast<const char*>(data);
+    std::size_t totalBytesRead = 0;
```
RiscadoA, line 47:
```suggestion
    std::size_t totalBytesWritten = 0;
```

core/src/net/tcp_listener.cpp
RiscadoA: Don't forget this! @roby2014

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,102 @@
...
+    CUBOS_INFO("Connecting stream to socket");
+    stream.inner(clientSocket);
+    return true;
+}
+
+void TcpListener::close()
+{
+    if (mSock != InnerInvalidSocket)
+    {
+        CUBOS_WARN("Closing TCP listener socket: '{}'", mSock);
```
RiscadoA, line 93:
```suggestion
```

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    CUBOS_INFO("TCP listener socket '{}' at address {} with port '{}'", mSock, address.toString(), port);
+    return true;
+}
+
+bool TcpListener::accept(TcpStream& stream) const
+{
+    if (mSock == InnerInvalidSocket)
+    {
+        CUBOS_ERROR("Invalid socket");
```
RiscadoA, line 73:
```suggestion
        CUBOS_ERROR("Listener hasn't been initialized correctly");
```

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    auto clientSocket = ::accept(mSock, nullptr, nullptr);
+    if (clientSocket == InnerInvalidSocket)
+    {
+        CUBOS_ERROR("Failed to accept TCP connection: {}", logSystemError());
+        return false;
+    }
+
+    CUBOS_INFO("Connecting stream to socket");
```
RiscadoA, line 84:
```suggestion
```

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,102 @@
...
+        return false;
+    }
+
+    if (::listen(mSock, maxQueueLength) < 0)
+    {
+        CUBOS_ERROR("Failed to listen to TCP socket: {}", logSystemError());
+        return false;
+    }
+
+    CUBOS_INFO("TCP listener socket '{}' at address {} with port '{}'", mSock, address.toString(), port);
```
RiscadoA, line 65:
```suggestion
    CUBOS_DEBUG("TCP listener socket '{}' bound to address {} with port '{}'", mSock, address.toString(), port);
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+    addr.sin_addr.s_addr = inet_addr(address.toString().c_str());
+    addr.sin_port = htons(port);
+
+    if (::connect(mSock, (struct sockaddr*)&addr, sizeof(addr)) < 0)
+    {
+        CUBOS_ERROR("Failed to connect TCP stream to {}:{}: {}", address.toString(), port, logSystemError());
+        return false;
+    }
+
+    CUBOS_DEBUG("New TCP stream at address {} and port '{}'", address.toString(), port);
```
RiscadoA, line 42:
```suggestion
    CUBOS_DEBUG("TCP stream connected to address {} and port '{}' successfully", address.toString(), port);
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+    disconnect();
+    mSock = socket;
+    CUBOS_DEBUG("New TCP stream at socket {}", mSock);
+}
+
+void TcpStream::disconnect()
+{
+    if (mSock != InnerInvalidSocket)
+    {
+        CUBOS_WARN("Closing TCP stream socket: '{}'", mSock);
```
RiscadoA, line 57:
```suggestion
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    CUBOS_DEBUG("New TCP stream at address {} and port '{}'", address.toString(), port);
+    return true;
+}
+
+void TcpStream::inner(InnerSocket socket)
+{
+    disconnect();
+    mSock = socket;
+    CUBOS_DEBUG("New TCP stream at socket {}", mSock);
```
RiscadoA, line 50:
```suggestion
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+        return 0;
+    }
+
+    if (bytesRead == 0)
+    {
+        CUBOS_WARN("TCP stream EOF");
+        mEof = true;
+    }
+
+    CUBOS_INFO("Incoming TCP message");
```
RiscadoA, line 106:
```suggestion
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+#endif
+    if (bytesRead < 0)
+    {
+        CUBOS_ERROR("Failed to read from TCP stream: {}", logSystemError());
+        return 0;
+    }
+
+    if (bytesRead == 0)
+    {
+        CUBOS_WARN("TCP stream EOF");
```
RiscadoA, line 102:
```suggestion
««
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+#else
+    auto bytesWritten = ::send(mSock, buffer, size, 0);
+#endif
+    if (bytesWritten < 0)
+    {
+        CUBOS_ERROR("Failed to write TCP buffer: {}", logSystemError());
+        return 0;
+    }
+
+    CUBOS_INFO("Sent TCP message");
```
RiscadoA, line 131:
```suggestion
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (bytesRead < 0)
+    {
+        CUBOS_ERROR("Failed to peek from TCP stream: {}", logSystemError());
+        return false;
+    }
+
+    if (bytesRead == 0)
+    {
+        CUBOS_WARN("TCP stream EOF");
+        // TODO: cant change mEof to true as its a const virtual method...
```
RiscadoA, line 161: It definitely should not be const, whoops. Feel free to change that!
Also, this peek implementation is a bit weird, as it erases data permanently.
Peek is supposed to be something like 'see what comes next without actually removing it from the stream, or advancing the cursor'.
You can implement it by storing an extra `char buffer` field or something, and a flag of whether that has been read or not before. If the flag is set when you call `peek`, you just return the existing buffer. Otherwise you write to it and set the flag. If the flag is set when you call `read`, you first copy the buffer to the output, and then continue reading if necessary; and you also should reset the flag.

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,165 @@
...
+    mPos += bytesWrittenCasted;
+    return bytesWrittenCasted;
+}
+
+char TcpStream::peek() const
+{
+    if (mSock == InnerInvalidSocket)
+    {
+        CUBOS_ERROR("Invalid socket");
+        return ' ';
```
RiscadoA, line 142:
```suggestion
        return ' \0';
```
Is probably a more fit 'default' character.

core/src/net/utils.cpp
```diff
@@ -0,0 +1,23 @@
+#include <cstring>
+
+#include <cubos/core/net/utils.hpp>
+
+#ifdef _WIN32
+#include <winsock2.h>
+#endif
+
```
RiscadoA, line 8: That has already been included in utils.hpp right?

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,82 @@
...
+        // Stream method implementations.
+
+        std::size_t read(void* buffer, std::size_t size) override;
+
+        std::size_t write(const void* buffer, std::size_t size) override;
+
+        inline std::size_t tell() const override
+        {
+            return mPos;
+        }
```
RiscadoA, line 61: I think it's best to not implement cursors at all in TCP streams, as seek isn't supported anyway.
The problem here is, what is this even counting? Both reads and writes? Just writes? I think to avoid the confusion you can just disable it.

RiscadoA (changes requested):
So that ended up to not be a small review at all, sorry for the nitpickiness! After these comments it should be good to go!

core/samples/net/tcp/page.md
roby2014: I added this in the UDP PR!!! Forgot to mention:
https://github.com/GameDevTecnico/cubos/pull/1257/commits/c94f5def4c29e427f9b3979d2daa027c69df30c9#diff-4f110d68eeb0c368c1fa8c90917c0d7d40969bc0f9dcfa2d52360747b916008c
Can I do it like this? easier to avoid rebasing stuff and such :V

core/samples/net/tcp/main.cpp
roby2014: Im confused, again. Will `read` handle this? Or does the client need to?

roby2014: Assuming your discord message was after this, I'm gonna assume `TcpStream::read` has to do it.

core/src/net/tcp_listener.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+TcpListener::~TcpListener()
+{
+    close();
+}
+
+bool TcpListener::listen(const Address& address, uint16_t port, int maxQueueLength)
+{
+    mSock = socket(AF_INET, SOCK_STREAM, 0);
```
RiscadoA, line 41: What if listen is called twice? The previous mSock should be destroyed.

core/samples/net/tcp/page.md
RiscadoA: Yeah that's okay, but I think you didn't the link to `net/page.md` there, on `docs/pages/3_examples/1_core/main.md`, right? Don't forget it!

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        return mPeekBuffer;
+    }
+
+    char c;
+
+#ifdef _WIN32
+    auto bytesRead = ::recv(mSock, &c, 1, MSG_PEEK);
+#else
+    auto bytesRead = ::recv(mSock, static_cast<void*>(&c), 1, MSG_PEEK);
```
RiscadoA, line 211:
```suggestion
    auto bytesRead = ::recv(mSock, &mPeekBuffer, 1, MSG_PEEK);
#else
    auto bytesRead = ::recv(mSock, static_cast<void*>(&mPeekBuffer), 1, MSG_PEEK);
```
The `c` variable isn't needed here right?

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    auto bytesWrittenCasted = static_cast<std::size_t>(bytesWritten);
+    mPos += bytesWrittenCasted;
+    return bytesWrittenCasted;
+}
+
+bool TcpStream::writeExact(const void* data, std::size_t size)
+{
+    const char* dataPtr = static_cast<const char*>(data);
+    std::size_t totalBytesRead = 0;
```
RiscadoA, line 172:
```suggestion
    std::size_t totalBytesWritten = 0;
```

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,266 @@
...
+        {
+            mEof = true;
+        }
+
+        auto bytesReadCasted = static_cast<std::size_t>(bytesRead + offset);
+        mPos += bytesReadCasted;
+        return bytesReadCasted;
+    }
+
+    mPos += offset;
```
RiscadoA, line 122: You no longer need `mPos` right?

core/src/net/tcp_stream.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+std::size_t TcpStream::read(void* buffer, std::size_t size)
+{
+    if (mSock == InnerInvalidSocket)
+    {
+        CUBOS_ERROR("Invalid socket");
+        return 0;
+    }
+
+    size_t offset = 0;
```
RiscadoA, line 88:
```suggestion
    std::size_t offset = 0;
```

RiscadoA (changes requested):

core/include/cubos/core/net/tcp_stream.hpp
```diff
@@ -0,0 +1,95 @@
...
+        /// @note This function will only return after buffer is completed or an error occurs.
+        /// @return Whether reading was sucessful.
+        bool readExact(void* data, std::size_t size);
+
+        /// @brief Writes data to the stream with the exact specified size.
+        /// @param data Buffer to write data from.
+        /// @param size Size of the buffer.
+        /// @note This function will only return after whole buffer is written or an error occurs.
+        /// @return Whether writing was sucessful.
+        bool writeExact(const void* data, std::size_t size);
```
RiscadoA, line 71: These utility methods should be part of the base class, no? Also, a note should be added to the base Stream::read/write method warning that the data may be written partially, and a recommendation to use the `*Exact` methods in case that's a problem.

RiscadoA (approved):
Other than the single comment I added, LGTM! Sorry for dragging this on for so long :pensive:

PR #1260, opened 2024-06-11
Add core::gl::Timer

tomas7770:
# Description

Adds a Timer to RenderDevice, which exposes OpenGL queries to allow profiling sections of rendering code. Also adds an higher-level PipelinedTimer for profiling every frame asynchronously.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/core/include/cubos/core/gl/render_device.hpp
+++ b/core/include/cubos/core/gl/render_device.hpp
@@ -77,6 +77,8 @@ namespace cubos::core::gl
         class ShaderStage;
         class ShaderPipeline;
         class ShaderBindingPoint;
+
+        class Timer;
     } // namespace impl
 
     /// @brief Handle to a framebuffer.
@@ -173,6 +175,11 @@ namespace cubos::core::gl
     /// @see @ref impl::ShaderBindingPoint - shader binding point interface.
     /// @ingroup core-gl
     using ShaderBindingPoint = impl::ShaderBindingPoint*;
+
+    /// @brief Handle to a timer.
+    /// @see @ref impl::Timer - timer interface.
+    /// @ingroup core-gl
+    using Timer = std::shared_ptr<impl::Timer>;
 
     /// @brief Render device properties that can be queried at runtime.
     /// @see @ref RenderDevice::getProperty().
@@ -876,6 +883,10 @@ namespace cubos::core::gl
         /// @param pipeline Shader pipeline handle.
         virtual void setShaderPipeline(ShaderPipeline pipeline) = 0;
 
+        /// @brief Creates a new timer.
+        /// @return Timer handle, or nullptr on failure.
+        virtual Timer createTimer() = 0;
+
         /// @brief Clears the color buffer bit on the current framebuffer to a specific color.
         /// @param r Red component.
         /// @param g Green component.
@@ -1453,6 +1464,34 @@ namespace cubos::core::gl
 
         protected:
             ShaderBindingPoint() = default;
+        };
+
+        /// @brief Abstract timer.
+        class CUBOS_CORE_API Timer
+        {
+        public:
+            virtual ~Timer() = default;
+
+            /// @brief Starts a region to be timed.
+            virtual void begin() = 0;
+
+            /// @brief Ends a region.
+            virtual void end() = 0;
+
+            /// @brief Checks if result is available.
+            /// @return Whether the result is available.
+            virtual bool done() = 0;
+
+            /// @brief Gets the result, in nanoseconds.
+            ///
+            /// Blocks until a result is available.
+            ///
+            /// @return Time spent in region, in nanoseconds, or -1 if an error
+            /// occurred while querying the time.
+            virtual int result() = 0;
+
+        protected:
+            Timer() = default;
         };
     } // namespace impl
 
--- a/core/include/cubos/core/gl/render_device.hpp
+++ b/core/include/cubos/core/gl/render_device.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <memory>
+#include <queue>
 #include <variant>
 
 #include <glm/glm.hpp>
@@ -79,6 +80,7 @@ namespace cubos::core::gl
         class ShaderBindingPoint;
 
         class Timer;
+        class PipelinedTimer;
     } // namespace impl
 
     /// @brief Handle to a framebuffer.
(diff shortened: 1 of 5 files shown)


DiogoMendonc-a (approved):
Not exactly sure what the pipelined timer is or what it's used for, but i don't see anything wrong with this

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+                while (!mTimers.empty() && mTimers.front()->done())
+                {
+                    mResult = mTimers.front()->result();
+                    mTimers.pop();
+                }
+
+                mNeedsEnd = false;
+
+                return {mResult, mTimers.size()};
+            }
```
RiscadoA, line 1562: Move definitions to .cpp (necessary due to CUBOS_CORE_API)

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+                int result;          ///< Time spent in region in the latest frame that finished rendering.
+                size_t framesBehind; ///< How many frames behind the CPU the latest rendered frame lags.
+            };
+
+            PipelinedTimer(RenderDevice* rd)
+                : mRd(rd)
+            {
+            }
+
+            ~PipelinedTimer() = default;
```
RiscadoA, line 1525: Unnecessary
```suggestion
```

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+            struct Result
+            {
+                int result;          ///< Time spent in region in the latest frame that finished rendering.
+                size_t framesBehind; ///< How many frames behind the CPU the latest rendered frame lags.
+            };
+
+            PipelinedTimer(RenderDevice* rd)
+                : mRd(rd)
+            {
+            }
```
RiscadoA, line 1523: Move to .cpp

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+
+            ~PipelinedTimer() = default;
+
+            /// @brief Starts a region to be timed.
+            void begin()
+            {
+                if (mNeedsEnd) // make sure it doesn't begin again before ending
+                {
+                    return;
+                }
```
RiscadoA, line 1533: I would `CUBOS_ASSERT` this, as when this happens it's always a mistake on the programmer's side right?
```suggestion
                CUBOS_ASSERT (!mNeedsEnd, "The previous call to begin() wasn't matched with a call to end()");
```

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+            /// @brief Ends a region.
+            /// @return Time spent in region in the latest frame that finished
+            /// rendering, in nanoseconds (-1 if no frame has finished yet),
+            /// and how many frames behind the CPU this frame lags.
+            Result end()
+            {
+                if (!mNeedsEnd) // make sure it doesn't end before beginning
+                {
+                    return {-1, static_cast<size_t>(-1)};
+                }
```
RiscadoA, line 1549: Same as above

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,94 @@ namespace cubos::core::gl
...
+
+                while (!mTimers.empty() && mTimers.front()->done())
+                {
+                    mResult = mTimers.front()->result();
+                    mTimers.pop();
+                }
+
+                mNeedsEnd = false;
+
+                return {mResult, mTimers.size()};
```
RiscadoA, line 1561: Why is `mResult` a member variable? Isn't it only used in this function?

RiscadoA (changes requested):
Thanks for working on this! This solution LGTM! Just some nitpicking :relieved:

core/include/cubos/core/gl/render_device.hpp
tomas7770: The idea was that it could persist after the function call, so that a valid value would still be returned in a future call if no timer had finished.

RiscadoA: Ah, makes sense. LGTM!

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,61 @@ namespace cubos::core::gl
...
+            virtual ~Timer() = default;
+
+            /// @brief Starts a region to be timed.
+            virtual void begin() = 0;
+
+            /// @brief Ends a region.
+            virtual void end() = 0;
+
+            /// @brief Returns whether the result is available.
+            /// @return True if the result is available, false otherwise.
```
roby2014, line 1496:
```cpp
/// @brief Checks if result is available.
/// @return Whether the result is available.
```

core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1454,6 +1479,61 @@ namespace cubos::core::gl
...
+            virtual bool done() = 0;
+
+            /// @brief Gets the result, in nanoseconds.
+            ///
+            /// Blocks until a result is available.
+            ///
+            /// @return Time spent in region, in nanoseconds.
+            virtual int result() = 0;
```
roby2014, line 1504: Maybe worth to say that you return `-1` if ...

core/src/gl/render_device.cpp
```diff
@@ -59,3 +60,34 @@ bool FramebufferDesc::FramebufferTarget::isSet() const
 {
     return mSet;
 }
+
+impl::PipelinedTimer::PipelinedTimer(RenderDevice* rd)
+    : mRd(rd)
```
RiscadoA, line 65:
```suggestion
impl::PipelinedTimer::PipelinedTimer(RenderDevice& rd)
    : mRd(rd)
```
Store a reference instead: it can never be null anyway, better make it clear.

RiscadoA (approved):
Looks good to me :partying_face: you can merge after resolving the single comment I added

roby2014 (approved):

PR #1263, opened 2024-06-13
Port to web

luishfonseca:
# Description

(Edited by @RiscadoA)

This PR ended up being absolutely gigantic.
I think the best way to review it is to give a quick look at the commits we made. They are very granular and all do pretty much a single thing, which makes it easier to follow the 'story' of what we were doing.
Make sure to at least give a look at the parts of the code which you're more familiar with/you authored, and see if stuff makes sense to you.

Web support was implemented with Emscripten, which is basically a C++ compiler which targets javascript and wasm.
Instead of outputting executables, this outputs an html page for each executable, i.e., all samples and also tesseratos.

We had to change lots of tiny nitpicks to make this work, particularly, on the rendering side.
A lot of our code was incompatible with WebGL. We reduced the interface of RenderDevice to contain only the interface supported by both OpenGL and WebGL.

We also added an action to build it here on GitHub. That action uploads the resulting artifacts, which you can download and try running on your machine. On a future PR we should improve this to automatically place the samples in their respective documentation pages. Will create issues for that soon

## Checklist

- [x] Self-review changes.


core/include/cubos/core/gl/render_device.hpp
```diff
@@ -1159,23 +1147,17 @@ namespace cubos::core::gl
...
+            // @brief Fills the buffer with the given data.
+            /// @param data Pointer to data.
+            /// @param size Data size in bytes.
+            virtual void fill(const void* data, std::size_t size) = 0;
+
+            // @brief Fills the buffer with the given data.
+            /// @param data Pointer to data.
+            /// @param size Data size in bytes.
+            /// @param offset Offset in the buffer where the data will be written.
+            /// @param synchronized TODO: IDK WHAT THIS DOES HELP RICARDO
```
kuukitenshi, line 1171: Ig that this TODO should be removed 🤣 
Sync here is when there are pending operations on the buffer, so you should wait until they are concluded and only then map it.

RiscadoA: LOL I reviewed the code but somehow let that slip, ty

kuukitenshi (approved):
LGTM 👍🏻

tomas7770 (approved):
LGTM!

SrGesus (approved):
LGTM

PR #1273, opened 2024-06-28
Transition to TGS Soft impulse solver

fallenatlas:
# Description

Oh boy...

Transition from XPBD solver to TGS solver.

Much of the solver for the penetration constraints was changed, we now make corrections based on impulses, based mainly on the equations for conservation of momentum with some bias proportional to the penetration which will work to separate the particles.

Collisions also work as soft constraints now, the values might need some tweaking in the future but they seem ok for now.

The integration also had to be changed and is now under the solver, because it just makes more sense.

The new loop is as follows:

- fixed step
  - collision stuff (unchanged)
  - prepare constraints (for now just create the penetration constraint for colliding pairs)
  - fixed substep
    - integrate velocity
    - solve velocity with bias
    - integrate position
    - solve velocity without bias (relax)
  - finalize position (apply position correction)

Currently the impulses are enough only to separate the particles, this means that some collisions may not look as expected because everything is perfectly inelastic. Elastic and inelastic collisions will come with restitution which should be the next PR.

Also of note that stacks of more than 2 particles aren't very good yet, so the complex physics sample won't look the best.

## Checklist

- [x] Self-review changes.


engine/include/cubos/engine/physics/plugin.hpp
```diff
@@ -39,24 +38,16 @@ namespace cubos::engine
...
     /// - @ref Impulse - holds impulses applied on a particle.
     /// - @ref Mass - holds the mass of an object.
-    /// - @ref PreviousPosition - holds the previous position of the entity in a substep.
     /// - @ref AccumulatedCorrection - holds the corrections accumulated from the constraints solving.
     ///
     /// ## Dependencies
     /// - @ref physics-gravity-plugin
+    ///
+    /// ## Tags
+    /// - physicsApplyForcesTag - tag for all systems that modify velocity or apply forces or impulses.
```
RiscadoA, line 48:
```suggestion

    /// @brief Tag which should be used on all systems that modify velocity or apply forces or impulses.
```

engine/src/physics/solver/integration/plugin.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <cubos/engine/transform/plugin.hpp>
+
+#include "../../fixed_substep/plugin.hpp"
+
+using namespace cubos::engine;
+
+CUBOS_DEFINE_TAG(cubos::engine::physicsApplyImpulsesTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsClearForcesTag);
+
+void cubos::engine::integrationPlugin(Cubos& cubos)
```
RiscadoA, line 18:
```suggestion
void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
```

engine/samples/complex-physics/main.cpp
RiscadoA, line 1: Just noticed the folder name doesn't follow the convention, it should be complex_physics.

engine/src/physics/plugin.cpp
```diff
@@ -76,9 +61,6 @@ CUBOS_REFLECT_IMPL(Damping)
 
 void cubos::engine::physicsPlugin(Cubos& cubos)
 {
-    cubos.depends(fixedStepPlugin);
-    cubos.depends(transformPlugin);
-
     cubos.plugin(fixedSubstepPlugin);
```
RiscadoA, line 64: This plugin should be prefixed with `physics`.

RiscadoA (changes requested):
Thanks for working on this! That was very quick :eyes: good job!

DiogoMendonc-a (approved):
LGTM, but to be fair, I don't really know what I'm looking at for most of these maths 😅

tomas7770 (approved):
LGTM as far as I can understand!

engine/include/cubos/engine/physics/plugin.hpp
```diff
@@ -39,24 +38,15 @@ namespace cubos::engine
     /// - @ref Force - holds forces applied on a particle.
     /// - @ref Impulse - holds impulses applied on a particle.
     /// - @ref Mass - holds the mass of an object.
-    /// - @ref PreviousPosition - holds the previous position of the entity in a substep.
     /// - @ref AccumulatedCorrection - holds the corrections accumulated from the constraints solving.
     ///
     /// ## Dependencies
     /// - @ref physics-gravity-plugin
+    ///
```
RiscadoA, line 45:
```suggestion
```

engine/src/physics/solver/plugin.cpp
```diff
@@ -1,22 +1,58 @@
...
+CUBOS_DEFINE_TAG(cubos::engine::physicsSolveContactTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsIntegratePositionTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsSolveRelaxConstraintTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsSolveRelaxUserConstraintTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsSolveRelaxContactTag);
+CUBOS_DEFINE_TAG(cubos::engine::physicsFinalizePositionTag);
 
 using namespace cubos::engine;
 
 void cubos::engine::solverPlugin(Cubos& cubos)
```
RiscadoA, line 24:
```suggestion
void cubos::engine::physicsSolverPlugin(Cubos& cubos)
```

RiscadoA (approved):
Other than the comments I added, LGTM!

PR #1277, opened 2024-07-03
Add friction and restitution on penetration constraint solving

fallenatlas:
# Description

- Adds friction impulse calculation, for now friction is the same for all particles.
- Adds restitutions to be able to customize the amount of energy lost/conserved on collision, changing the perceived bounciness of objects. For now the restitution coefficient is the same for all particles.

Note: I couldn't find how to apply restitution using accumulated impulses anywhere, which is ridiculous, so the implementation I have is based only on how Erin Catto does it in Box2D v3.0, which he says it's not the most accurate but it seems to work pretty well.

## Checklist

- [x] Self-review changes.


engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -50,15 +50,12 @@ void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection
...
-            // penetration is positive but we need negative value for separation
-            penetration = -penetration;
-
             float bias = 0.0F;
             float massScale = 1.0F;
             float impulseScale = 0.0F;
             if (penetration > 0.0F)
             {
-                bias = penetration * 1.0F / subDeltaTime;
+                bias = 0.2F * penetration * (1.0F / subDeltaTime);
```
RiscadoA, line 58: Why `0.2F`? Should this be a constant somewhere? And should it be configurable by the dev?

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -111,6 +101,68 @@ void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection
...
+            {
+                vr2 = velocity1.vec;
+                vr1 = velocity2.vec;
+            }
+
+            glm::vec3 vr = vr2 - vr1;
+            glm::vec3 tangent1 = vr - glm::dot(vr, constraint.normal) * constraint.normal;
+            glm::vec3 tangent2;
+            float tangentLenSq = glm::length2(tangent1);
+            if (tangentLenSq > 1e-6)
```
RiscadoA, line 129: Maybe we should also use a constant here? We might have some problems regarding floating point precision.
@luishfonseca what do you think?

RiscadoA (approved):
Overall LGTM! Thanks for working on this! I don't really understand the math, but codewise the only thing I have to point out is the magic number stuff

engine/src/physics/solver/penetration_constraint/plugin.cpp
luishfonseca: It would probably be a good idea to make all magic numbers configurable settings, not really for the user, but for our fine tuning work later on.

engine/src/physics/solver/penetration_constraint/plugin.cpp
luishfonseca: 1e-6 is huge, not really sure what is the impact of erroneously skipping this block

luishfonseca:
@fallenatlas please share your resources for the physics engine to ease reviewing. The math looks right, but i don't know if I'd catch errors without a reference

DiogoMendonc-a (approved):
Very hard to review these extremely niche PRs, but lgtm

engine/src/physics/solver/penetration_constraint/plugin.cpp
fallenatlas: I did leave the wrong number here, the place where I took this from suggests "epsilon * epsilon", they didn't say what epsilon should be but if I recall I saw that bullet had it at 1e-6 back in like 2007 so that's what I used and forgot to multiply. The impact of skipping would be that you wouldn't have friction at that step, from what I read it's better to have a lower epsilon the faster the objects can move, essencially we're just making friction more sensitive to tiny movements, gonna try and find where I read that from.

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -111,6 +101,68 @@ void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection
...
+                vr2 = velocity2.vec;
+                vr1 = velocity1.vec;
+            }
+            else
+            {
+                vr2 = velocity1.vec;
+                vr1 = velocity2.vec;
+            }
+
+            glm::vec3 vr = vr2 - vr1;
```
luishfonseca, line 125:
```suggestion
            glm::vec3 vr = velocity2.vec - velocity1.vec;
            
            if (ent1 != constraint.entity)
            {
                vr *= -1.0f;
            }
```

luishfonseca (changes requested):

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -50,15 +50,12 @@ void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection
                 penetration += glm::dot(deltaPenetration, constraint.normal);
             }
 
-            // penetration is positive but we need negative value for separation
```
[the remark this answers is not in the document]
fallenatlas: They are yeah, it's the same thing as with relative velocity, we just change the direction of the vector deltaPenetration.

engine/src/physics/solver/penetration_constraint/plugin.cpp
fallenatlas: Going to open an issue for this.

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -68,47 +77,90 @@ void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection
...
-                correction2.impulse = newImpulse;
+                vr *= -1.0F;
             }
 
-            glm::vec3 p = constraint.normal * impulse;
-            if (ent1 == constraint.entity)
+            glm::vec3 tangent1 = vr - glm::dot(vr, constraint.normal) * constraint.normal;
+            glm::vec3 tangent2;
+            float tangentLenSq = glm::length2(tangent1);
+            if (tangentLenSq > 1e-6 * 1e-6)
```
luishfonseca, line 126: I'm still a bit sketched out by these epsilon values, they're just guesses. But now the impact of incorrectly detecting it as 0 is much smaller, so lgtm

luishfonseca (approved):

PR #1282, opened 2024-07-07
994 detect collisions for voxel vs voxel and voxel vs box

joaomanita:
# Description
Added a voxel collision shape which decomposes a voxel grid into a set of boxes to be used in collision detection.
Added a greedy meshing like algorithm to do the decomposing into boxes.
Added systems to detect collisions between voxel collision shapes and box shapes, and voxel-voxel.

## Checklist

- [X] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


engine/src/collisions/plugin.cpp
```diff
@@ -42,4 +142,58 @@ void cubos::engine::collisionsPlugin(Cubos& cubos)
...
+                        }
+                        if (possibleSize.x > 0)
+                        {
+                            expandBoxYaxis(copiedGrid, currentMin, size, possibleSize);
+                            glm::vec3 boxCenter = glm::vec3(currentMin + static_cast<glm::ivec3>(possibleSize)) / 2.0F;
+                            cubos::core::geom::Box box;
+                            box.halfSize = boxCenter;
+                            glm::vec3 shift = center - boxCenter;
+                            shape.insertBox(box, shift);
+                            CUBOS_WARN("inserted pair");
```
luishfonseca, line 187: was this forgotten here?

luishfonseca (dismissed):
Needs a sample to be approved, I want to play a bit with it before.
The code overall looks quite nice, good job :+1:

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -65,4 +66,86 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
+                        if (intersects)
+                        {
+                            cmds.relate(ent1, ent2,
+                                        CollidingWith{.entity = ent1,
+                                                      .entity1OriginalPosition = position1.vec,
+                                                      .entity2OriginalPosition = position2.vec,
+                                                      .penetration = intersectionInfo.penetration,
+                                                      .position = {0.0F, 0.0F, 0.0F},
+                                                      .normal = intersectionInfo.normal});
+                            break;
```
RiscadoA, line 141: This is still missing the multiple intersection handling code. Maybe we should leave it to another PR, what do you think?

RiscadoA (changes requested):
Thanks for working on this! Also mostly looks good to me, just needs a bit clean up.

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -113,78 +198,229 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
+                        // Create a translation matrix for the shift
+                        glm::mat4 shiftMatrix = glm::translate(glm::mat4(1.0F), -box.shift);
+                        matchedLocalToWorld2.mat *= shiftMatrix;
 
-                    // If we flipped incident and reference planes, we will
-                    // need to flip it back before sending it to the manifold.
-                    if (flipped)
+                        auto newPoints = computePoints(matchedShape1, matchedLocalToWorld1, matchedShape2,
+                                                       &matchedLocalToWorld2, collidingWith);
+                        points.insert(points.end(), newPoints.begin(), newPoints.end());
```
fallenatlas, line 332: Correct me if I'm wrong, here you are going through all boxes in the voxel shape and computing the manifold with the box, right? However, the collidingWith was already performed between each of these, which means that we should know exactly which boxes withing the voxel shape are colliding with this one. I think that we could maybe keep like an index or whatever is best of the specific box in the voxel shape that we collided with and only do computePoints between that and the box. 

Extra, but for more context: The computation of the manifold assumes the shapes are colliding, the only sanity check it performs is with penetration in the direction of the normal, and we are computing it between boxes that may not be colliding, I think this can create incorrect forces in the solver in some occasions.

fallenatlas (changes requested):
Overall looks good to me, good job 👍 Just have a couple comments regarding the manifold part.

fallenatlas (commented):
Another thing, in the voxel collision sample it would be nice to add the manifold visualization like we have in the other sample, would help to make sure everything is correct.

engine/src/collisions/narrow_phase/plugin.cpp
fallenatlas: Disregard my comment here, I think it's something still worth thinking about but we have to change other things, before worrying about this.

engine/include/cubos/engine/collisions/shapes/voxel.hpp
```diff
@@ -0,0 +1,97 @@
+/// @file
+/// @brief Component @ref cubos::engine::VoxelCollisionShape.
+/// @ingroup collisions-plugin
+
+#pragma once
+
+#include <glm/glm.hpp>
+
+#include <cubos/core/geom/box.hpp>
+#include <cubos/core/reflection/external/glm.hpp>
```
RiscadoA, line 10: You only need this include in the .cpp where you define the reflection

engine/samples/voxel-shape-collisions/main.cpp
```diff
@@ -0,0 +1,223 @@
...
+            .add(Rotation::lookingAt({3.0F, -1.0F, 0.0F}, glm::vec3{0.0F, 1.0F, 0.0F}));
+    });
+
+    cubos.startupSystem("configure Assets").tagged(settingsTag).call([](Settings& settings) {
+        settings.setString("assets.io.path", SAMPLE_ASSETS_FOLDER);
+    });
+
+    /// [Set palette]
+    cubos.startupSystem("set palette").call([](RenderPalette& palette) { palette.asset = PaletteAsset; });
+    /// [Set palette]
```
RiscadoA, line 98:
```suggestion
    cubos.startupSystem("set palette").call([](RenderPalette& palette) { palette.asset = PaletteAsset; });
```
You aren't using these, since the sample doesn't have its own page anyway.

engine/samples/voxel-shape-collisions/main.cpp
```diff
@@ -0,0 +1,223 @@
...
+    Entity b;
+
+    glm::vec3 aRotationAxis;
+    glm::vec3 bRotationAxis;
+};
+
+/// [Get handles to assets]
+static const Asset<VoxelGrid> CarAsset = AnyAsset("059c16e7-a439-44c7-9bdc-6e069dba0c75");
+static const Asset<VoxelPalette> PaletteAsset = AnyAsset("1aa5e234-28cb-4386-99b4-39386b0fc215");
+/// [Get handles to assets]
```
RiscadoA, line 69:
```suggestion
static const Asset<VoxelGrid> CarAsset = AnyAsset("059c16e7-a439-44c7-9bdc-6e069dba0c75");
static const Asset<VoxelPalette> PaletteAsset = AnyAsset("1aa5e234-28cb-4386-99b4-39386b0fc215");
```

engine/src/collisions/plugin.cpp
```diff
@@ -1,19 +1,193 @@
...
+            }
+        }
+    }
+    max = glm::ivec3(max.x + 1, max.y + 1, max.z + 1);
+    glm::vec3 dist = glm::vec3(max) - glm::vec3(min);
+    aabb.min({-dist.x / 2.0F, -dist.y / 2.0F, -dist.z / 2.0F});
+    aabb.max({dist.x / 2.0F, dist.y / 2.0F, dist.z / 2.0F});
+}
+
+pair<glm::uvec3, glm::uvec3> findLargestBox(const cubos::engine::VoxelGrid& grid,
```
RiscadoA, line 48:
```suggestion
static pair<glm::uvec3, glm::uvec3> findLargestBox(const cubos::engine::VoxelGrid& grid,
```

engine/src/collisions/plugin.cpp
```diff
@@ -1,19 +1,193 @@
...
+            {
+                processed[x][y][z] = true;
+            }
+        }
+    }
+
+    return std::make_pair(start, end);
+}
+
+vector<pair<glm::uvec3, glm::uvec3>> greedy3dMeshing(const cubos::engine::VoxelGrid& grid)
```
RiscadoA, line 160:
```suggestion
static vector<pair<glm::uvec3, glm::uvec3>> greedy3dMeshing(const cubos::engine::VoxelGrid& grid)
```

engine/tests/settings.cpp
```diff
@@ -52,12 +52,21 @@ TEST_CASE("cubos::engine::settingsPlugin")
     };
 
     EmbeddedArchive::registerData("settings.json", settingsData);
-    FileSystem::mount("/settings.json", std::make_unique<EmbeddedArchive>("settings.json"));
+
+    if (!FileSystem::mount("/settings.json", std::make_unique<EmbeddedArchive>("settings.json")))
+    {
+        FileSystem::unmount("/settings.json");
+        FileSystem::mount("/settings.json", std::make_unique<EmbeddedArchive>("settings.json"));
+    }
```
RiscadoA, line 60: ? what is going on here

engine/tests/settings.cpp
```diff
@@ -52,12 +52,21 @@ TEST_CASE("cubos::engine::settingsPlugin")
...
 
     // Run the actual test.
     Cubos cubos{argc, argv};
     cubos.plugin(settingsPlugin);
 
+    cubos.startupSystem("configure Assets").tagged(settingsTag).call([](Settings& settings) {
+        settings.setBool("assets.io.enable", false);
+    });
```
RiscadoA, line 68: Why is this here?

RiscadoA (changes requested):

RiscadoA (changes requested):
You still haven't resolved my new comments :thinking:

engine/tests/settings.cpp
joaomanita: This test was failing because "/settings.json" was already mounted from executing the raycast test, getting a "could not mount archive error". I don't really know how the file system works so I just unmounted and remounted.

joaomanita:
> You still haven't resolved my new comments 🤔

Oops, I re-requested a review without noticing you had already reviewed it :innocent:

engine/include/cubos/engine/collisions/shapes/voxel.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <cubos/engine/assets/asset.hpp>
+#include <cubos/engine/voxels/grid.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component which adds a voxel collision shape to an entity, used with a @ref Collider component.
+    /// @ingroup collisions-plugin
+    class VoxelCollisionShape
+    {
+
```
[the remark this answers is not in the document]
RiscadoA: Don't forget this!

engine/tests/settings.cpp
RiscadoA: Don't mark stuff as resolved without answering them! Otherwise I might just forget about it.
This test doesn't even use assets so it really shouldn't need this

RiscadoA (changes requested):

RiscadoA (approved):
From my side, LGTM! Regarding the actual mechanics and design, I'll leave that to the physics team members which are way more knowledgeable in this than I am.

RiscadoA:
Don't forget to add an entry to the changelog @joaomanita!

GCeSilva (approved):
LGTM!

PR #1287, opened 2024-07-18
Add shadow mapping with support for spot shadow casters

tomas7770:
# Description

Implements shadows for spot lights, including support for multiple lights and soft shadows.

- `shadowAtlasPlugin` creates and manages the shadow atlas, a large texture that holds the shadow maps for each spot light.
- `shadowAtlasRasterizerPlugin` draws to the shadow maps.
- `deferredShadingPlugin` has been modified to read from the shadow atlas, in order to draw shadows.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.


engine/include/cubos/engine/render/shadow_atlas/shadow_atlas.hpp
diogomsmiranda, line 1: At a first sight LGTM, although I'm not really familiar with this part of Cubos, I see a lot of structs with no documentation, is this on purpose?

diogomsmiranda:
On another note, is there a way to test this functionality?

engine/include/cubos/engine/render/shadow_atlas/shadow_atlas.hpp
tomas7770: For most of them, yes. Other existing render plugins have similar undocumented structs.

As for the one on this file (`Slot`), seems reasonable to add documentation since the user can access it, and other nested structs in the engine also have docs.

tomas7770:
> On another note, is there a way to test this functionality?

I'm not sure how the testing framework works, but I figure the `shadowAtlasPlugin` should be testable. The rest probably has the risk of slightly different behavior on different GPUs.

engine/src/render/shadow_atlas/plugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+            atlas.atlas = rd.createTexture2D(desc);
+
+            CUBOS_INFO("Resized ShadowAtlas to {}x{}", atlas.size.x, atlas.size.y);
+        }
+
+        // New frame, hint that the shadow atlas texture needs to be cleared.
+        atlas.cleared = false;
+    });
+
+    cubos.system("reserve space for shadow casters")
```
RiscadoA, line 54: For now this algorithm works and LGTM! It would be cool to improve it in the future though, to allow, for example, assigning a priority to increase the resolution of given lights. If you can, open an issue for this. This could be used for getting extra resolution on lights closer to the camera, and lower on those farther away.

RiscadoA (changes requested):
Thank you for working on this! :tada: The PR is looking great. Added some nitpicks

RiscadoA:
> > On another note, is there a way to test this functionality?
> 
> I'm not sure how the testing framework works, but I figure the `shadowAtlasPlugin` should be testable. The rest probably has the risk of slightly different behavior on different GPUs.

The CI can't handle GPU stuff properly, so yeah, it's not worth the hassle to write tests for these kind of stuff. Do ensure that we have good samples for these features though!
If could add a new render sample focused on shadows, it would be great.

engine/include/cubos/engine/render/shadow_atlas/shadow_atlas.hpp
```diff
@@ -0,0 +1,55 @@
...
+    /// @ingroup render-shadow-atlas-plugin
+    struct CUBOS_ENGINE_API ShadowAtlas
+    {
+        CUBOS_REFLECT;
+
+        /// @brief Size of the shadow atlas texture, in pixels.
+        glm::uvec2 size = {0, 0};
+
+        /// @brief Configured size of the shadow atlas texture, in pixels.
+        glm::uvec2 configSize = {4096, 4096};
```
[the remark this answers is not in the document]
tomas7770: `size` is the actual size of the atlas, used for drawing to it.
`configSize` is the intended size which may be changed by other plugins.
They may not match if the texture hasn't been recreated yet after a change to `configSize`.

Maybe I should change `size` to a private variable and add a getter?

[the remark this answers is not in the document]
RiscadoA: Yeah that seems the way to go. In that case, do not expose `size` on the reflection, as it isn't meant to be changed directly anyway.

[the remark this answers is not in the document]
tomas7770: Just realized that the `shadowAtlasPlugin` needs to set `size` when resizing the texture, so it's probably necessary to have a `resize()` method.

engine/src/render/shadow_atlas/plugin.cpp
tomas7770: Will open after this is merged!

RiscadoA (approved):
LGTM! Thank you for working on this!

fallenatlas (approved):
I see some familiar things but I don't know much about this, so can't say much about the algorithm, but the sample works and other than that it LGTM! 👍

diogomsmiranda (approved):
LGTM

PR #1289, opened 2024-07-21
refactor(reflection): no longer fallback to anonymous types automatically

RiscadoA:
# Description

This change was necessary due to the previous system being highly error prone. Before, if reflect<T> was called for a type T without defined reflection included, a placeholder definition would be used which returned an anonymous type. Sadly, due to the C++ One Definition Rule, this meant that other source files using reflect<T>, could return an anonymous type, even if the type's reflection definition was included!

Basically, previously a simple mistake such as forgetting to include an external reflection header could lead to a random reflection failure in the core of the library, such as a missing trait error. Now, the users must define reflection for all types, as there is no longer a fallback. To simplify this for the previous 'anonymous types', a new CUBOS_ANONYMOUS_REFLECT macro was added.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] ~~Write new samples.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Move Opt class to memory namespace (#1212, **@roby2014**).
 - Moved from XPBD to TGS Soft for physics solving (#1269, **@fallenatlas**).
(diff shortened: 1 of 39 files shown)


Scarface1809 (approved):

diogomsmiranda (approved):
LGTM

luishfonseca (approved):

PR #1294, opened 2024-07-25
Physics Materials

fallenatlas:
# Description

Add PhysicsMaterial component which allows to customize the behaviour of physics objects by changing their friction coefficient and bounciness. How the values of each are mixed to determine the value to use in the collision resolution can also be customized by choosing between 4 methods, which are some of the most common methods found online. This is inspired by Unity and Unreal which also do this. What the best is will depend on the use case, so I believe that this is better than only letting 1 method.

What some use (restitution only): https://stackoverflow.com/questions/1678908/applying-coefficient-of-restitution-in-a-collision-resolution-method
Unity: https://docs.unity3d.com/Manual/collider-surfaces-combine.html
Unreal: https://dev.epicgames.com/documentation/en-us/unreal-engine/physical-materials-reference-for-unreal-engine

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/src/physics/plugin.cpp
+++ b/engine/src/physics/plugin.cpp
@@ -46,10 +46,9 @@ CUBOS_REFLECT_IMPL(Impulse)
         .build();
 }
 
-CUBOS_REFLECT_EXTERNAL_DECL(CUBOS_EMPTY, PhysicsMaterial::MixProperty);
 CUBOS_REFLECT_EXTERNAL_IMPL(PhysicsMaterial::MixProperty)
 {
-    return cubos::core::reflection::Type::create("PhysicsMaterial::MixProperty")
+    return cubos::core::reflection::Type::create("cubos::engine::PhysicsMaterial::MixProperty")
         .with(cubos::core::reflection::EnumTrait{}
                   .withVariant<PhysicsMaterial::MixProperty::Maximum>("Maximum")
                   .withVariant<PhysicsMaterial::MixProperty::Multiply>("Multiply")
--- a/engine/src/physics/plugin.cpp
+++ b/engine/src/physics/plugin.cpp
@@ -2,6 +2,9 @@
 
 #include <glm/glm.hpp>
 
+#include <cubos/core/reflection/traits/enum.hpp>
+#include <cubos/core/reflection/type.hpp>
+
 #include <cubos/engine/physics/plugin.hpp>
 #include <cubos/engine/physics/solver/plugin.hpp>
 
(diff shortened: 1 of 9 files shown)


DiogoMendonc-a (approved):
lgtm

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -266,10 +290,12 @@ void cubos::engine::penetrationConstraintPlugin(Cubos& cubos)
                 float frictionMass = kFriction > 0.0F ? 1.0F / kFriction : 0.0F;
 
                 // determine friction (set to predefined value for now)
```
tomas7770, line 292: `(set to predefined value for now)`

Is this part of the comment still supposed to be here, now that friction/restitution are customizable in the material?

tomas7770 (approved):
LGTM! Just a minor observation...

engine/src/physics/solver/penetration_constraint/plugin.cpp
fallenatlas: No it isn't, nice catch

engine/src/physics/plugin.cpp
```diff
@@ -43,6 +46,27 @@ CUBOS_REFLECT_IMPL(Impulse)
         .build();
 }
 
+CUBOS_REFLECT_EXTERNAL_DECL(CUBOS_EMPTY, PhysicsMaterial::MixProperty);
```
RiscadoA, line 49: Reflection declarations should stay in header files - otherwise, users of this enum won't be able to reflect it if necessary. You can move it below the cubos::engine namespace in the header. Only the definition should be here.

engine/src/physics/plugin.cpp
```diff
@@ -43,6 +46,27 @@ CUBOS_REFLECT_IMPL(Impulse)
...
+CUBOS_REFLECT_EXTERNAL_IMPL(PhysicsMaterial::MixProperty)
+{
+    return cubos::core::reflection::Type::create("PhysicsMaterial::MixProperty")
```
RiscadoA, line 52:
```suggestion
    return cubos::core::reflection::Type::create("cubos::engine::PhysicsMaterial::MixProperty")
```

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -33,6 +33,29 @@ void getPlaneSpace(const glm::vec3& n, glm::vec3& tangent1, glm::vec3& tangent2)
     }
 }
 
+PhysicsMaterial::MixProperty getMixProperty(PhysicsMaterial::MixProperty mixProperty1,
+                                            PhysicsMaterial::MixProperty mixProperty2)
+{
+    return mixProperty1 > mixProperty2 ? mixProperty2 : mixProperty1;
+}
+
+float mixValues(float value1, float value2, PhysicsMaterial::MixProperty mixProperty)
```
RiscadoA, line 42:
```suggestion
static float mixValues(float value1, float value2, PhysicsMaterial::MixProperty mixProperty)
```

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -33,6 +33,29 @@ void getPlaneSpace(const glm::vec3& n, glm::vec3& tangent1, glm::vec3& tangent2)
     }
 }
 
+PhysicsMaterial::MixProperty getMixProperty(PhysicsMaterial::MixProperty mixProperty1,
```
RiscadoA, line 36:
```suggestion
static PhysicsMaterial::MixProperty getMixProperty(PhysicsMaterial::MixProperty mixProperty1,
```

engine/src/physics/plugin.cpp
```diff
@@ -89,6 +114,7 @@ void cubos::engine::physicsPlugin(Cubos& cubos)
                 cmds.add(ent, force);
                 cmds.add(ent, impulse);
                 cmds.add(ent, AccumulatedCorrection{});
+                cmds.add(ent, PhysicsMaterial{});
```
RiscadoA, line 117: Maybe you should add a `PhysicsMaterial material` field to the `PhysicsBundle` component, and then add it here?
That way the user can configure the material more easily.

RiscadoA (approved):
Overall LGTM! Pointed out some nitpicks, feel free to merge it after resolving those

PR #1297, opened 2024-07-29
Add components to support rigid body rotation

fallenatlas:
# Description

Add components for torque, angular impulse, inertia, angular velocity and center of mass.
Add ability to add force or impulse on a point of the object, for this we changed the Force and Impulse components, which now also hold a torque and angular impulse, respectively.

Reference: https://research.ncl.ac.uk/game/mastersdegree/gametechnologies/physicstutorials/3angularmotion/Physics%20-%20Angular%20Motion.pdf

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/include/cubos/engine/physics/components/center_of_mass.hpp
+++ b/engine/include/cubos/engine/physics/components/center_of_mass.hpp
@@ -1,0 +1,23 @@
+/// @file
+/// @brief Component @ref cubos::engine::CenterOfMass.
+/// @ingroup physics-plugin
+
+#pragma once
+
+#include <glm/vec3.hpp>
+
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component which defines the center of mass of a body.
+    /// @ingroup physics-plugin
+    struct CUBOS_ENGINE_API CenterOfMass
+    {
+        CUBOS_REFLECT;
+
+        glm::vec3 vec = {0.0F, 0.0F, 0.0F};
+    };
+} // namespace cubos::engine
--- a/engine/include/cubos/engine/physics/components/center_of_mass.hpp
+++ b/engine/include/cubos/engine/physics/components/center_of_mass.hpp
@@ -18,6 +18,6 @@ namespace cubos::engine
     {
         CUBOS_REFLECT;
 
-        glm::vec3 vec = {0.0F, 0.0F, 0.0F};
+        glm::vec3 vec = {0.0F, 0.0F, 0.0F}; ///< Position of the center in local space.
     };
 } // namespace cubos::engine
--- a/engine/include/cubos/engine/physics/physics_bundle.hpp
+++ b/engine/include/cubos/engine/physics/physics_bundle.hpp
@@ -28,6 +28,7 @@ namespace cubos::engine
         glm::vec3 impulse = {0.0F, 0.0F, 0.0F};
         glm::vec3 angularImpulse = {0.0F, 0.0F, 0.0F};
         PhysicsMaterial material = PhysicsMaterial{};
-        glm::mat3 inertiaTensor = glm::mat3(0.0F);
+        glm::mat3 inertiaTensor = glm::mat3(0.0F); // Temporary value. We use this value to check if the inertia tensor
+                                                   // is to be calculated automatically.
     };
 } // namespace cubos::engine
--- a/engine/include/cubos/engine/physics/physics_bundle.hpp
+++ b/engine/include/cubos/engine/physics/physics_bundle.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <glm/mat3x3.hpp>
 #include <glm/vec3.hpp>
 
 #include <cubos/core/reflection/reflect.hpp>
@@ -20,7 +21,6 @@ namespace cubos::engine
         CUBOS_REFLECT;
 
         float mass = 1.0F;
-        glm::vec3 colliderDimensions = {1.0F, 1.0F, 1.0F};
         glm::vec3 velocity = {0.0F, 0.0F, 0.0F};
         glm::vec3 angularVelocity = {0.0F, 0.0F, 0.0F};
         glm::vec3 force = {0.0F, 0.0F, 0.0F};
@@ -28,5 +28,6 @@ namespace cubos::engine
         glm::vec3 impulse = {0.0F, 0.0F, 0.0F};
         glm::vec3 angularImpulse = {0.0F, 0.0F, 0.0F};
         PhysicsMaterial material = PhysicsMaterial{};
+        glm::mat3 inertiaTensor = glm::mat3(0.0F);
     };
 } // namespace cubos::engine
--- a/engine/include/cubos/engine/physics/physics_bundle.hpp
+++ b/engine/include/cubos/engine/physics/physics_bundle.hpp
@@ -20,9 +20,13 @@ namespace cubos::engine
         CUBOS_REFLECT;
 
         float mass = 1.0F;
+        glm::vec3 colliderDimensions = {1.0F, 1.0F, 1.0F};
         glm::vec3 velocity = {0.0F, 0.0F, 0.0F};
+        glm::vec3 angularVelocity = {0.0F, 0.0F, 0.0F};
         glm::vec3 force = {0.0F, 0.0F, 0.0F};
+        glm::vec3 torque = {0.0F, 0.0F, 0.0F};
         glm::vec3 impulse = {0.0F, 0.0F, 0.0F};
+        glm::vec3 angularImpulse = {0.0F, 0.0F, 0.0F};
         PhysicsMaterial material = PhysicsMaterial{};
     };
 } // namespace cubos::engine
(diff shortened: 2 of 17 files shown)


engine/include/cubos/engine/physics/physics_bundle.hpp
```diff
@@ -20,7 +21,11 @@ namespace cubos::engine
 
         float mass = 1.0F;
         glm::vec3 velocity = {0.0F, 0.0F, 0.0F};
+        glm::vec3 angularVelocity = {0.0F, 0.0F, 0.0F};
         glm::vec3 force = {0.0F, 0.0F, 0.0F};
+        glm::vec3 torque = {0.0F, 0.0F, 0.0F};
         glm::vec3 impulse = {0.0F, 0.0F, 0.0F};
+        glm::vec3 angularImpulse = {0.0F, 0.0F, 0.0F};
+        glm::mat3 inertiaTensor = glm::mat3(0.0F);
```
RiscadoA, line 29: Btw `glm::mat3(0.0F)` isn't an identity matrix, it is a matrix filled with zeroes. `glm::mat3(1.0F)` is an identity matrix. I don't understand the math though, so you might have written this on purpose. Pointing this out just to be sure

engine/src/physics/plugin.cpp
```diff
@@ -21,6 +22,22 @@ CUBOS_REFLECT_IMPL(Mass)
...
+{
+    return cubos::core::ecs::TypeBuilder<CenterOfMass>("cubos::engine::CenterOfMass")
+        .withField("vec", &CenterOfMass::vec)
+        .build();
+}
+
+CUBOS_REFLECT_IMPL(Inertia)
+{
+    return cubos::core::ecs::TypeBuilder<Inertia>("cubos::engine::Inertia")
+        .withField("Inertia", &Inertia::inertia)
```
RiscadoA, line 39:
```suggestion
        .withField("inertia", &Inertia::inertia)
```

engine/src/physics/plugin.cpp
```diff
@@ -31,15 +48,40 @@ CUBOS_REFLECT_IMPL(Velocity)
...
+        .withField("vec", &AngularVelocity::vec)
+        .build();
+}
+
 CUBOS_REFLECT_IMPL(Force)
 {
-    return cubos::core::ecs::TypeBuilder<Force>("cubos::engine::Force").withField("mForce", &Force::mForce).build();
+    return cubos::core::ecs::TypeBuilder<Force>("cubos::engine::Force")
+        .withField("mForce", &Force::mForce)
+        .withField("mTorque", &Force::mTorque)
```
RiscadoA, line 62: Don't expose private fields through reflection - they're not meant to be modified directly. You can just reflect an empty type.

engine/src/physics/plugin.cpp
```diff
@@ -31,15 +48,40 @@ CUBOS_REFLECT_IMPL(Velocity)
...
+        .build();
+}
+
+CUBOS_REFLECT_IMPL(Torque)
+{
+    return cubos::core::ecs::TypeBuilder<Torque>("cubos::engine::Torque")
+        .withField("mTorque", &Torque::mTorque)
```
RiscadoA, line 69: Ditto

engine/src/physics/plugin.cpp
```diff
@@ -31,15 +48,40 @@ CUBOS_REFLECT_IMPL(Velocity)
...
     return cubos::core::ecs::TypeBuilder<Impulse>("cubos::engine::Impulse")
         .withField("mImpulse", &Impulse::mImpulse)
+        .withField("mAngularImpulse", &Impulse::mAngularImpulse)
+        .build();
+}
+
+CUBOS_REFLECT_IMPL(AngularImpulse)
+{
+    return cubos::core::ecs::TypeBuilder<AngularImpulse>("cubos::engine::AngularImpulse")
+        .withField("mAngularImpulse", &AngularImpulse::mAngularImpulse)
```
RiscadoA, line 84: Ditto

engine/src/physics/plugin.cpp
```diff
@@ -31,15 +48,40 @@ CUBOS_REFLECT_IMPL(Velocity)
...
+{
+    return cubos::core::ecs::TypeBuilder<Torque>("cubos::engine::Torque")
+        .withField("mTorque", &Torque::mTorque)
+        .build();
 }
 
 CUBOS_REFLECT_IMPL(Impulse)
 {
     return cubos::core::ecs::TypeBuilder<Impulse>("cubos::engine::Impulse")
         .withField("mImpulse", &Impulse::mImpulse)
```
RiscadoA, line 76: Ditto

engine/src/physics/plugin.cpp
```diff
@@ -58,16 +104,32 @@ CUBOS_REFLECT_IMPL(Damping)
     return core::ecs::TypeBuilder<Damping>("cubos::engine::Damping").build();
 }
 
+// Compute Inertia Tensor for box shape
+glm::mat3 boxInertiaTensor(float mass, glm::vec3 dimensions)
```
RiscadoA, line 108:
```suggestion
static glm::mat3 boxInertiaTensor(float mass, glm::vec3 dimensions)
```

engine/src/physics/plugin.cpp
```diff
@@ -81,16 +143,65 @@ void cubos::engine::physicsPlugin(Cubos& cubos)
...
+                cmds.add(ent, AngularVelocity{.vec = bundle.angularVelocity});
                 cmds.add(ent, force);
+                cmds.add(ent, torque);
                 cmds.add(ent, impulse);
+                cmds.add(ent, angularImpulse);
                 cmds.add(ent, AccumulatedCorrection{});
             }
         });
 
+    // Should this be here?
```
RiscadoA, line 172: Good question. I think you can keep it here for now, but maybe in the future we should keep those in a separate sub plugin?

engine/src/physics/plugin.cpp
```diff
@@ -81,16 +143,65 @@ void cubos::engine::physicsPlugin(Cubos& cubos)
...
+                }
+
+                mass.changed = false;
+                shape.changed = false;
+
+                // Object has infinite mass
+                if (mass.inverseMass == 0.0F)
+                {
+                    inertia.inertia = glm::mat3(0.0F);
+                    inertia.inverseInertia = glm::mat3(0.0F);
```
RiscadoA, line 195: Once again, is this supposed to be an identity matrix or a zero matrix?

engine/include/cubos/engine/physics/physics_bundle.hpp
fallenatlas: Yup, I know, this is essencially just a default to know when we want to automatically calculate the inertia tensor, bc we're not supposed to have objects with mass 0 and inertia 0, which is what a matrix of all 0s means in this case, when I unpack the physicsBundle I check to see if it's still 0 or it has been replaced, bc if the dev put a specific value then we probably don't want to replace it. Maybe we should make it more explicit, so there isn't any confusion?

engine/src/physics/plugin.cpp
fallenatlas: I think so yeah. Should it be maybe a plugin with all the systems to update some component automatically like this? I initially thought just for Inertia, bc we have one more that will be needed eventually, but the idea seemed weird to me.

engine/src/physics/plugin.cpp
fallenatlas: Here it's another case, the inverseInertia is correct, the inertia itself I'm still not sure what to put bc it should be infinite, which is the same issue we currently have with mass. This doesn't affect anything for now bc we don't use it directly, but we should have a better way to identify these edge cases probably. I still haven't looked into it bc I'm pretty sure that when we introduce proper static bodies and such, this will naturally come into discussion again, but I'll create an issue so I don't forget.

engine/include/cubos/engine/physics/components/center_of_mass.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace cubos::engine
+{
+    /// @brief Component which defines the center of mass of a body.
+    /// @ingroup physics-plugin
+    struct CUBOS_ENGINE_API CenterOfMass
+    {
+        CUBOS_REFLECT;
+
+        glm::vec3 vec = {0.0F, 0.0F, 0.0F};
```
DiogoMendonc-a, line 21: You should probably document in what space this vector is in

DiogoMendonc-a (approved):

joaomanita (approved):

engine/src/physics/plugin.cpp
RiscadoA: Yeah I'm not sure tbh. Probably best to keep everything together for now, and when we get the full 'essential' feature set implemented we split it into multiple plugins to make it more readable

engine/include/cubos/engine/physics/physics_bundle.hpp
RiscadoA: I would just add a comment explaining that it is only a temporary value, otherwise LGTM!

RiscadoA (approved):
Other than the still unresolved conversations, LGTM!
Sorry for taking so long to review this :pensive:

PR #1313, opened 2024-09-14
Fix spot light range cutoff when using shadows

tomas7770:
# Description

Fixes spot lights being "cut" before the maximum range is reached, when shadows are used on a spot light with global scale lower than 1.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


joaomanita (approved):

RiscadoA (approved):
Good catch! LGTM!

Dacops (approved):

PR #1315, opened 2024-09-19
Add rotation on constraint solving

fallenatlas:
# Description

Adds angular velocity computation in penetration constraint solving.

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/src/physics/solver/penetration_constraint/plugin.cpp
+++ b/engine/src/physics/solver/penetration_constraint/plugin.cpp
@@ -2,7 +2,7 @@
 
 #include <glm/glm.hpp>
 
-#include <cubos/engine/collisions/colliding_with.hpp>
+#include <cubos/engine/collisions/contact_manifold.hpp>
 #include <cubos/engine/collisions/plugin.hpp>
 #include <cubos/engine/fixed_step/plugin.hpp>
 #include <cubos/engine/physics/plugin.hpp>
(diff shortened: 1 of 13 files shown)


joaomanita (approved):

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -33,6 +33,27 @@ void getPlaneSpace(const glm::vec3& n, glm::vec3& tangent1, glm::vec3& tangent2)
     }
 }
 
+void getTangents(const glm::vec3& velocity1, const glm::vec3& velocity2, const glm::vec3& n, glm::vec3& tangent1,
```
RiscadoA, line 36:
```suggestion
static void getTangents(const glm::vec3& velocity1, const glm::vec3& velocity2, const glm::vec3& n, glm::vec3& tangent1,
```

engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -56,52 +77,58 @@ static float mixValues(float value1, float value2, PhysicsMaterial::MixProperty
     }
 }
 
-void solvePenetrationConstraint(Query<Entity, const Mass&, AccumulatedCorrection&, Velocity&, PenetrationConstraint&,
-                                      Entity, const Mass&, AccumulatedCorrection&, Velocity&>
-                                    query,
-                                const FixedDeltaTime& fixedDeltaTime, const Substeps& substeps, const bool useBias)
+void solvePenetrationConstraint(
```
RiscadoA, line 80:
```suggestion
static void solvePenetrationConstraint(
```

RiscadoA (approved):
As usual, I don't really understand what's going on here, so I'll just trust you on the math.
The sample also runs fine on my machine, but I noticed the movement of the boxes is sometimes weird/unnatural? Is it still missing step or something? What's next?

RiscadoA:
Also, don't forget to add an entry to the changelog!

fallenatlas:
> Thanks for working on this 🥳 Very cool to see rotation being relevant. As usual, I don't really understand what's going on here, so I'll just trust you on the math. The sample also runs fine on my machine, but I noticed the movement of the boxes is sometimes weird/unnatural? Is it still missing step or something? What's next?

Sometimes it still looks weird yeah, I'm not entirely sure of what the problem is, it could be because of the friction and bounciness settings in the sample which could produce something we don't naturally expect, or my guess is that it has to do with the softness parameters, which might be too strong (they limit the velocities for the solving thus making it feel less crispy and more sluggish, but it's important for stability), but I don't fully understand this part yet so it's something that I'll have to look into in the future.

RiscadoA:
@fallenatlas Just noticed your commit messages are not very correct:
- when working on samples, don't use `feat(samples)`, instead use something like `docs(physics)`
- don't use `penetration_constraint` as a scope, use `physics` (the rule is to stick to the B-label names). If you want to mention it's related to that, put it in the message itself.

If you can, change this before merging!

Dageus (approved):

PR #1320, opened 2024-09-24
Add initial communication between game and editor

RiscadoA:
# Description

Adds remote debugging to Cubos. Basically, the Cubos class now parses its incoming command line arguments.
If `--debug <port>` is passed, then, instead of immediately starting and running the game, it waits for a TCP connection on that port. When a connection is accepted, it listens for commands, such as:
- `run` - makes the application resume running.
- `pause` - makes the application stop and wait for further commands.
- `update <N>` - makes the application run its update systems `N` times.
- `close` - closes the application.
- `disconnect` - closes the connection. The application starts listening again for another connection.

Also adds a simple UI tool to test this functionality inside Tesseratos.

## Checklist

- [x] Self-review changes.
- [x] ~~Ensure test coverage.~~ (too troublesome to test right now, we don't have time for that in 0.4 :pensive:)
- [x] ~~Write new samples.~~ (works with any existing sample)

### DIFF
--- a/core/include/cubos/core/ecs/cubos.hpp
+++ b/core/include/cubos/core/ecs/cubos.hpp
@@ -16,6 +16,7 @@
 #include <cubos/core/ecs/system/tag.hpp>
 #include <cubos/core/ecs/world.hpp>
 #include <cubos/core/memory/opt.hpp>
+#include <cubos/core/reflection/type_server.hpp>
 
 namespace cubos::core::ecs
 {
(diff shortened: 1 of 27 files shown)


kuukitenshi (approved):
I didn't do an in-depth review because I don't wanna delay it, but after skimming through, it looks good to me.

core/include/cubos/core/ecs/cubos.hpp
```diff
@@ -327,6 +343,12 @@ namespace cubos::core::ecs
 
         /// @brief Maps registered tags to their data.
         std::unordered_map<std::string, TagInfo> mTags;
+
+        /// @brief Port to listen on, in case debugging is enabled.
+        memory::Opt<uint16_t> mDebugPort;
+
+        /// @brief Type server used in case debugging is enabled.
+        reflection::TypeServer mTypeServer;
```
mkuritsu, line 351: I'm not sure about this but since the mDebugPort is an opt because is only used when debugging is enabled shouldn't mTypeServer also be an opt?

RiscadoA: Good point! Will change it to an Opt :relieved:

mkuritsu (approved):

GalaxyCrush (approved):

PR #1329, opened 2024-09-27
Allow opening projects through Tesseratos

RiscadoA:
# Description

Adds the project plugin, which adds support for the `ProjectManager` system argument, along with an ImGui window for opening and closing projects + launching them and terminating them.
Also adds an utility class Process to the core to handle multiple processes in a cross-platform way.

There are also some other random fixes or simple methods that I needed to add to other areas in order to make this possible. Looking at the commit history probably gives a better view of what I changed.

**I need people with Windows and MacOS to check if this also works in other operating systems**

## Checklist

- [x] Self-review changes.
- [x] ~~Ensure test coverage.~~ The only part that we could in theory could (easily) test of this is the cross-platform multi process utilities I added, but sadly doctest doesn't handle multi process stuff well at all.
- [x] ~~Write new samples.~~ You can test it by opening tesseratos, opening the Project window, setting a project folder and a binary/executable path, and launching it. Then you can use the debugger tool to step, run or terminate the game.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Global telemetry level (for tracing/logging) (#1265, **@roby2014**).
 - Spans for tracing and profiling (#1265, **@roby2014**).
 - Simple cross-platform multi-process utilities (**@RiscadoA**).
+- Project opening and closing on Tesseratos (#1218, **@RiscadoA**).
 
 ### Changed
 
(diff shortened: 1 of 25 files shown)


Scarface1809 (approved):

kuukitenshi (approved):
It all works on Windows ✨
Aside from that, I didn't find any issues or bugs. The code LGTM!

diogomsmiranda (approved):
Manually Tested for macOS and works!

Why the choice to call it "Update" and not "Step"?

We should later add a more interactive way to add the project path and binary path. Like a File Explorer type thing.

Code LGTM.

RiscadoA:
> Manually Tested for macOS and works!
> 
> Why the choice to call it "Update" and not "Step"?
> 
> We should later add a more interactive way to add the project path and binary path. Like a File Explorer type thing.
> 
> Code LGTM.

I used 'update' in order to be consistent with 'update' systems, but yeah, probably changing it to 'step' is more readable.
Also, @Scarface1809 has almost finished the file selector tool, so we'll be able to do that on the next version :eyes:

PR #1333, opened 2024-09-27
Show stacktrace on CUBOS_FAIL

RiscadoA:
# Description

Adds yet another dependency to the core lib, *cpptrace*.
This one is really necessary - turns out implementing this ourselves is a giant hassle, as there a lot of platform and compiler quirks.
In the future it might no longer be necessary, as C++23 comes with a `<stacktrace>` include - not all compilers support it yet.

To use this, I simply call our own `abort` function, which first prints the stacktrace using `cpptrace`, and then calls the usual `std::abort` function.

The result looks like this:
image

## Checklist

- [x] Self-review changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "engine/lib/implot"]
 	path = engine/lib/implot
 	url = https://github.com/epezent/implot.git
+[submodule "core/lib/cpptrace"]
+	path = core/lib/cpptrace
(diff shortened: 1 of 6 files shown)


mkuritsu (approved):
Stacktraces look good 👍

kuukitenshi (approved):
Looks nice to me 😸

fallenatlas (approved):
LGTM!

PR #1335, opened 2024-09-28
feat(rendering): Add OrthographicCamera component

mkuritsu:
# Description

Added OrthographicCamera component to update the Camera with a orthographic projetion.

One thing I'm not sure if its the best approach is the size parameter added, I put it as a multiplier to the viewport the camera is drawing to. Would it be better to leave it like that or change it to be values to represent the plane size, something like a vec2{800, 600}?

## Checklist

- [x] Self-review changes.
- [x] Update render main sample.


kuukitenshi (approved):
LGTM 👍🏻

tomas7770:
> One thing I'm not sure if its the best approach is the size parameter added, I put it as a multiplier to the viewport the camera is drawing to. Would it be better to leave it like that or change it to be values to represent the plane size, something like a vec2{800, 600}?

With the multiplier approach, does the visible area become smaller/larger depending on the viewport size? It doesn't seem right that different players see more or less of the game world just because their display resolution is different, atleast not by default. There is a lot of variance in display resolution/size combinations, ranging from things like 800x600 (usually windowed) all the way to 4K/8K, and small laptop displays to large TVs, which could result in very different experiences.

mkuritsu:
> > One thing I'm not sure if its the best approach is the size parameter added, I put it as a multiplier to the viewport the camera is drawing to. Would it be better to leave it like that or change it to be values to represent the plane size, something like a vec2{800, 600}?
> 
> With the multiplier approach, does the visible area become smaller/larger depending on the viewport size? It doesn't seem right that different players see more or less of the game world just because their display resolution is different, atleast not by default. There is a lot of variance in display resolution/size combinations, ranging from things like 800x600 (usually windowed) all the way to 4K/8K, and small laptop displays to large TVs, which could result in very different experiences.

Ok yes true, wasn't thinking in that way, my mistake I will change and commit with the other approach.

RiscadoA:
> > > One thing I'm not sure if its the best approach is the size parameter added, I put it as a multiplier to the viewport the camera is drawing to. Would it be better to leave it like that or change it to be values to represent the plane size, something like a vec2{800, 600}?
> > 
> > 
> > With the multiplier approach, does the visible area become smaller/larger depending on the viewport size? It doesn't seem right that different players see more or less of the game world just because their display resolution is different, atleast not by default. There is a lot of variance in display resolution/size combinations, ranging from things like 800x600 (usually windowed) all the way to 4K/8K, and small laptop displays to large TVs, which could result in very different experiences.
> 
> Ok yes true, wasn't thinking in that way, my mistake I will change and commit with the other approach.

I guess we have four options here:
a) User specifies the final ortographic projection size - we simply don't take into account the target and viewport in the projection matrix math; leads to stretching but the user might want this in some cases I guess? I'm not sure
b) Multiply user specified size by the target*viewport size (what you have right now); AR is correct among all possible viewport configs, but users will be able to see a larger region of the world (in all directions) when rendering at higher resolutions.
c) User specifies the width of the projection matrix, and the height is calculated by multiplying the width by the aspect ratio of the target+viewport. AR is also correct on all possible configs, but users with higher display resolutions won't be able to view more of the world. Instead, users with weird aspect ratios might get to see more or less than what they should, but only in the vertical direction.
d) Same as above, but user specifies the height of the projection matrix instead of the width. I think this is more fitting than c), as in the previous options, users with extra wide displays would have their 'vertical view' very reduced. By specifying the height, we get the opposite. Users with extra wide displays get a larger 'horizontal view' but the 'vertical view' remains constant.

In my opinion, options c) and d) are the most useful, and we should probably allow the user to opt between them.
Maybe keep a `float size;` + an enum specifying which axis is being specified.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Generic Camera component to hold projection matrix (#1331, **@mkuritsu**)
 - Initial application debugging through Tesseratos (#1303, **@RiscadoA**).
 - Print stacktrace with *cpptrace* on calls to CUBOS_FAIL (#1172, **@RiscadoA**).
+- OrthographicCamera component (#1182, **@mkuritsu**)
```
RiscadoA, line 22:
```suggestion
- Orthographic Camera component (#1182, **@mkuritsu**).
```

tomas7770:
> In my opinion, options c) and d) are the most useful, and we should probably allow the user to opt between them.
> Maybe keep a `float size;` + an enum or boolean specifying which axis is being specified. What do you think?

Looks good. I think the perspective camera basically follows d)? Alternatively we could have a) and let a separate plugin manage the size, but that might be adding unneeded complexity.

mkuritsu:
> In my opinion, options c) and d) are the most useful, and we should probably allow the user to opt between them.
> Maybe keep a `float size;` + an enum or boolean specifying which axis is being specified. What do you think?

Yes seems good, I was already making the d) option but I will add the enum to also allow for c) then.

engine/include/cubos/engine/render/camera/orthographic_camera.hpp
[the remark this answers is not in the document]
mkuritsu: Regarding this perspective camera is written with the _camera suffix it will seem weird having one with it and one without it, should I rename perspective_camera.hpp to be only perspective.hpp (will need to replace includes), keep the suffix or let only the orthographic without suffix?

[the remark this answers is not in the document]
RiscadoA: I meant renaming both, my bad. So yeah, perspective_camera.hpp should also become perspective.hpp

RiscadoA:
Just noticed this, don't use uppercase in the first letter of the commit messages; it should be: `feat(scope): hey`, not `feat(scope): Hey`

kuukitenshi (approved):

engine/include/cubos/engine/render/camera/orthographic_camera.hpp
```diff
@@ -0,0 +1,41 @@
...
+    /// @ingroup render-camera-plugin
+    struct CUBOS_ENGINE_API OrthographicCamera
+    {
+        CUBOS_REFLECT;
+
+        /// @brief The axis that can be fixed for this projection.
+        enum class Axis
+        {
+            HORIZONTAL,
+            VERTICAL
```
RiscadoA, line 24:
```suggestion
            Horizontal,
            Vertical
```
I dunno why clang-tidy didn't catch this, but we use PascalCase for enum variants.

engine/include/cubos/engine/render/camera/orthographic_camera.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+        /// @brief Near clipping plane.
+        float zNear{0.1F};
+
+        /// @brief Far clipping plane.
+        float zFar{1000.0F};
+    };
+} // namespace cubos::engine
+
+CUBOS_REFLECT_EXTERNAL_DECL(CUBOS_EMPTY, cubos::engine::OrthographicCamera::Axis);
```
RiscadoA, line 41:
```suggestion
CUBOS_REFLECT_EXTERNAL_DECL(CUBOS_ENGINE_API, cubos::engine::OrthographicCamera::Axis);
```

RiscadoA (approved):
Added my last two comments, after resolving my comments it should be good to go.
Don't forget to remove the `_camera` suffix from the files!

mkuritsu:
> Added my last two comments, after resolving my comments it should be good to go. Don't forget to remove the `_camera` suffix from the files!

ok I think should all be fixed sorry for the small mistakes

RiscadoA:
> > Added my last two comments, after resolving my comments it should be good to go. Don't forget to remove the `_camera` suffix from the files!
> 
> ok I think all should be fixed sorry for the small mistakes

Dw about it, sorry for being so picky!

tomas7770 (approved):
LGTM!

PR #1338, opened 2024-09-30
Add cascaded shadow maps

tomas7770:
# Description

Adds cascaded shadow maps, to allow for directional lights to cast shadows.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.


engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
 
+        static constexpr int MaxSplits = 10;
+
         cubos::engine::ShadowCaster baseSettings;
 
-        int numSplits;                     ///< Number of splits for PSSM/CSM.
-        std::vector<float> splitDistances; ///< Distance of each split from 0 to 1.
-        float maxDistance = 0;             ///< Max distance (if 0, uses camera zFar).
+        int configNumSplits = 3; ///< Target number of splits for PSSM/CSM.
+        int numSplits = 0;       ///< Current number of splits for PSSM/CSM.
```
RiscadoA, line 31: I would swap these names, and make the 'current number of splits' a private variable.
You could then have a method which compares these two, and returns whether they're different. If they're different, the current is set to the target.

engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
+
+        std::vector<float> configSplitDistances; ///< Target distance of each split's near plane from 0 to 1 (relative
+                                                 ///< to maxDistance). If the amount of distances doesn't match the
+                                                 ///< number of splits, default values will be used.
+        std::vector<float> splitDistances;       ///< Distance of each split's near plane from 0 to 1.
```
RiscadoA, line 36: Same as above.

engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
+            /// @brief Recreates the shadow map textures.
+            /// @param rd Render device used to create the textures.
+            /// @param configSize Max size of the shadow map textures.
+            /// @param numSplits Number of shadow map cascades.
+            void resize(cubos::core::gl::RenderDevice& rd, glm::uvec2 configSize, int numSplits);
+
+        private:
+            /// @brief Size of the shadow map textures, in pixels.
+            glm::uvec2 mSize = {0, 0};
+        };
```
RiscadoA, line 62: For more 'complex' classes like this, to avoid cluttering the component code, we usually just declare them within the component, at the top, and then define them outside.

For example:

```cpp
class DirectionalCaster
{
public:
   /// @brief Directional shadow map data.
   class ShadowMap;

  ... other fields
};

class DirectionalShadowCaster::ShadowMap
{
  // actual class definition
};
```

engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
+
+        std::unordered_map<core::ecs::Entity, ShadowMap, core::ecs::EntityHash>
+            shadowMaps; ///< Caster's shadow maps for each camera.
+
+        /// @brief Target size of the shadow map textures, in pixels.
+        /// Use this to change the resolution of the shadow maps. Note that the
+        /// real resolution may not match this, as it may be dynamically adjusted.
+        glm::uvec2 configSize = {4096, 4096};
+
+        glm::uvec2 size = {0, 0}; ///< Used to check if configSize has been changed since the latest frame.
```
RiscadoA, line 72: Same as above.

engine/src/render/cascaded_shadow_maps/plugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+            for (auto [caster] : query)
+            {
+                // Remove shadow maps for cameras that no longer exist
+                std::vector<Entity> removedCameras;
+                for (auto& [cameraEntity, shadowMap] : caster.shadowMaps)
+                {
+                    if (!cameras.at(cameraEntity))
+                    {
+                        removedCameras.push_back(cameraEntity);
+                    }
```
RiscadoA, line 37: Maybe here we should also discard inactive cameras. What do you think?

engine/src/render/cascaded_shadow_maps/plugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                // Create shadow maps for new cameras
+                for (auto [entity, camera] : cameras)
+                {
+                    if (!caster.shadowMaps.contains(entity))
+                    {
+                        caster.shadowMaps[entity] = DirectionalShadowCaster::ShadowMap{};
+                        caster.shadowMaps[entity].resize(window->renderDevice(), caster.size, caster.numSplits);
+                    }
```
RiscadoA, line 78: Same as above, should we discard inactive cameras here?

engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
+        class ShadowMap
+        {
+        public:
+            core::gl::Texture2DArray cascades;         ///< Cascades of the shadow map.
+            core::gl::Texture2DArray previousCascades; ///< Used to check if the framebuffer needs to be recreated.
+            core::gl::Framebuffer framebuffer;         ///< Framebuffer used by the shadow rasterizer.
+
+            /// @brief Gets the size of the shadow map textures.
+            /// @return Size of the shadow map textures, in pixels.
+            glm::uvec2 getSize() const;
```
RiscadoA, line 51:
```suggestion
            glm::uvec2 size() const;
```
For consistency with the C++ STL

engine/include/cubos/engine/render/shadows/directional_caster.hpp
```diff
@@ -19,10 +23,52 @@ namespace cubos::engine
...
-        std::vector<float> splitDistances; ///< Distance of each split from 0 to 1.
-        float maxDistance = 0;             ///< Max distance (if 0, uses camera zFar).
+        int configNumSplits = 3; ///< Target number of splits for PSSM/CSM.
+        int numSplits = 0;       ///< Current number of splits for PSSM/CSM.
+
+        std::vector<float> configSplitDistances; ///< Target distance of each split's near plane from 0 to 1 (relative
+                                                 ///< to maxDistance). If the amount of distances doesn't match the
+                                                 ///< number of splits, default values will be used.
+        std::vector<float> splitDistances;       ///< Distance of each split's near plane from 0 to 1.
+
```
RiscadoA, line 37: Also, isn't the `numSplits` variable redundant? Isn't it always equal to the size of splitDistances vector?

engine/src/render/cascaded_shadow_maps/plugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+                else
+                {
+                    // Set default split distances
+                    caster.splitDistances.clear();
+                    for (int i = 0; i < caster.numSplits; i++)
+                    {
+                        caster.splitDistances.push_back(float(i) / float(caster.numSplits));
+                    }
+                }
+
```
RiscadoA, line 70: You could check here if the `numSplits`/`splitDistances.size()` is > MaxSplits, and in that case, issue a warning and clamp/crop it down to the allowed number.

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
```diff
@@ -0,0 +1,314 @@
...
+using cubos::core::io::Window;
+using cubos::engine::RenderMeshVertex;
+
+namespace
+{
+    // Holds the data sent per scene to the GPU.
+    struct PerScene
+    {
+        glm::mat4 lightViewProj[cubos::engine::DirectionalShadowCaster::MaxSplits];
+        int numSplits;
```
RiscadoA, line 39: I think you might need to include padding here (`int padding[3];`), otherwise scary UB might happen, due to the struct on the GPU having 12 bytes of padding.

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
```diff
@@ -0,0 +1,314 @@
...
+        for (int y = -1; y <= 1; y += 2)
+        {
+            for (int z = -1; z <= 1; z += 2)
+            {
+                auto corner = glm::inverse(viewProj) * glm::vec4(x, y, z, 1.0F);
+                corners.push_back(corner / corner.w);
+            }
+        }
+    }
+}
```
RiscadoA, line 116: This information will be useful in other regions of the code (i.e. frustum culling), maybe you can make it a static method of the new `Camera` component? Another option would be a global function in a `core/geom/frustum.hpp` or something (might make more sense in this case).

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+                    float cameraZFar = 1000.0F;
+                    if (perspectiveCamera.contains())
+                    {
+                        cameraZFar = perspectiveCamera.value().zFar;
+                    }
+                    else if (orthographicCamera.contains())
+                    {
+                        cameraZFar = orthographicCamera.value().zFar;
+                    }
```
RiscadoA, line 207: Huh, I guess the `zFar` and `zNear` properties should belong to the base `Camera`. Could you move it there?
Then you would be able to abstract this over camera type.

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
```diff
@@ -0,0 +1,314 @@
...
+                            }
+                            else
+                            {
+                                cameraProj = glm::ortho(-ortho.size, ortho.size, -ortho.size / aspect,
+                                                        ortho.size / aspect, near, far);
+                            }
+                        }
+
+                        std::vector<glm::vec4> cameraFrustumCorners;
+                        getCameraFrustumCorners(cameraProj * cameraView, cameraFrustumCorners);
```
RiscadoA, line 246:
```suggestion
                        getCameraFrustumCorners(camera.projection * cameraView, cameraFrustumCorners);
```
You don't need to calculate the projection yourself right? Why not use the field in the Camera component?
Then this whole code wouldn't need references to specific camera types, right? I might be missing something in the math

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
```diff
@@ -0,0 +1,314 @@
...
+                        };
+                        state.perMeshCB->fill(&perMesh, sizeof(perMesh));
+
+                        // Iterate over the buckets of the mesh (it may be split over many of them).
+                        for (auto bucket = mesh.firstBucketId; bucket != RenderMeshPool::BucketId::Invalid;
+                             bucket = pool.next(bucket))
+                        {
+                            rd.drawTriangles(pool.bucketSize() * bucket.inner, pool.vertexCount(bucket));
+                        }
+                    }
```
RiscadoA, line 310: Not something to fix right now, but worth pointing out I guess.
The fact that we have this logic duplicated over at least three files as of now smells of code smell to me.
We should think about this further; will open an issue.

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -214,17 +270,138 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
...
+                        {
+                            float cameraZFar = 1000.0F;
+                            if (perspectiveCamera.contains())
+                            {
+                                cameraZFar = perspectiveCamera.value().zFar;
+                            }
+                            else if (orthographicCamera.contains())
+                            {
+                                cameraZFar = orthographicCamera.value().zFar;
+                            }
```
RiscadoA, line 298: If you do make the change I mentioned before, make sure to make this Camera type agnostic again

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -214,17 +270,138 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
...
+                                    {
+                                        cameraProj = glm::ortho(-ortho.size * aspect, ortho.size * aspect, -ortho.size,
+                                                                ortho.size, near, far);
+                                    }
+                                    else
+                                    {
+                                        cameraProj = glm::ortho(-ortho.size, ortho.size, -ortho.size / aspect,
+                                                                ortho.size / aspect, near, far);
+                                    }
+                                }
```
RiscadoA, line 336: Same as above

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -214,17 +270,138 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
...
+                                    break;
+                                case 2:
+                                    perLight.shadowFarSplitDistances[i / 4].z = far;
+                                    break;
+                                case 3:
+                                    perLight.shadowFarSplitDistances[i / 4].w = far;
+                                    break;
+                                default:
+                                    break;
+                                }
```
RiscadoA, line 394:
```suggestion
                                    perLight.shadowFarSplitDistances[i / 4][i % 4] = far;
```

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -293,6 +470,9 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
                     state.albedoBP->bind(gBuffer.albedo);
                     state.ssaoBP->bind(ssao.blurTexture);
                     state.shadowAtlasBP->bind(shadowAtlas.atlas);
+                    // directionalShadowMap needs to be binded even if it's null, or else errors may occur on some GPUs
```
RiscadoA, line 473:
```suggestion
                    // directionalShadowMap needs to be bound even if it's null, or else errors may occur on some GPUs
```

engine/assets/render/deferred_shading.fs
```diff
@@ -120,9 +130,73 @@ vec3 spotLightCalc(vec3 fragPos, vec3 fragNormal, SpotLight light)
...
+            float bias = light.shadowBias / positionLightSpace.w; // make the bias not depend on near/far planes
+            // PCF
+            if (light.shadowBlurRadius <= 0.001f)
+            {
+                float pcfDepth = texture(directionalShadowMap, vec3(uv.xy, split)).r;
+                shadow += currentDepth - bias > pcfDepth ? 1.0 : 0.0;
+            }
+            else
+            {
+                vec2 texelSize = vec2(1.0 / 1024.0); // largely arbitrary value, affects blur size
```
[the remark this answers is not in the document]
tomas7770: It can be, but I'm not sure it makes much sense, since it only scales the blur radius by a constant factor, and the radius can already be configured.

engine/src/render/cascaded_shadow_maps/plugin.cpp
tomas7770: Sounds good!

engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp
tomas7770: The problem is that the way the camera frustum is split for each cascade is by changing the near/far planes in the projection.

On a second thought, perhaps `getCameraFrustumCorners` can be modified to convert different Z coordinates rather than just the corners. This way the frustum could be split without recalculating the projection matrix.

joaomanita (approved):

RiscadoA (approved):
Other than the comments still to be resolved, LGTM!

core/src/gl/ogl_render_device.cpp
```diff
@@ -2009,12 +2009,14 @@ Texture2DArray OGLRenderDevice::createTexture2DArray(const Texture2DArrayDesc& d
     GLuint id;
     glGenTextures(1, &id);
     glBindTexture(GL_TEXTURE_2D_ARRAY, id);
-    glTexStorage3D(GL_TEXTURE_2D_ARRAY, static_cast<GLsizei>(desc.mipLevelCount), internalFormat,
-                   static_cast<GLsizei>(desc.width), static_cast<GLsizei>(desc.height),
-                   static_cast<GLsizei>(desc.size));
+    for (std::size_t j = 0; j < desc.mipLevelCount; ++j)
+    {
+        glTexImage3D(GL_TEXTURE_2D_ARRAY, static_cast<GLint>(j), internalFormat, static_cast<GLsizei>(desc.width),
```
mkuritsu, line 2014:
```suggestion
        glTexImage3D(GL_TEXTURE_2D_ARRAY, static_cast<GLint>(j), static_cast<GLint>(internalFormat), static_cast<GLsizei>(desc.width),
```
On clang is not compiling without this static_cast

mkuritsu (approved):
Apart from the change I mention LGTM!

Issue #1004, opened 2024-02-24
Add audio support to the engine


RiscadoA:
# Checklist
- [x] #1005
- [ ] #231
- [ ] #1041
- [ ] #1042
- [ ] #206

roby2014:
![giphy](https://github.com/GameDevTecnico/cubos/assets/47612455/8456f125-b4e3-4205-8ce8-dfe2db4e8728)

PR #1339, opened 2024-10-03
Replace OpenAL audio device for Miniaudio backend

### LINKED ISSUE #1005: Replace OpenAL audio device by another backend

diogomsmiranda:
# Description

This PR is one of many in #1004, it replaces the OpenAL audio backend with the miniaudio.h backend.

Closes #1005.

## Checklist

- [x] Self-review changes.


core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+        /// @brief Enumerates the available devices.
+        /// @param[out] devices Vector to fill with the available devices.
+        virtual void enumerateDevices(std::vector<std::string>& devices) = 0;
+
+        /// @brief Creates a new audio device
+        /// @param specifier The specifier of the audio device, used to identify it
+        /// @param listenerCount The number of audio listener to be created by the device's engine.
+        /// @return Handle of the new device
+        virtual AudioDevice createDevice(unsigned int listenerCount,
+                                         const std::string& specifier = "MiniaudioDevice") = 0;
```
RiscadoA, line 69: There shouldn't be a reference to Miniaudio here, this is supposed to be abstract.
You could use an empty specifier as a default value, and then handle it on the miniaudio device as necessary.

core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+        /// @return AudioContext, or nullptr on failure.
+        static std::shared_ptr<AudioContext> create();
+
+        /// @brief Enumerates the available devices.
+        /// @param[out] devices Vector to fill with the available devices.
+        virtual void enumerateDevices(std::vector<std::string>& devices) = 0;
+
+        /// @brief Creates a new audio device
+        /// @param specifier The specifier of the audio device, used to identify it
+        /// @param listenerCount The number of audio listener to be created by the device's engine.
```
RiscadoA, line 66: Docs are swapped

RiscadoA, line 66:
```suggestion
        /// @param specifier Identifier of the audio device.
        /// @param listenerCount Number of audio listener to be supported by the device.
```

core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+    public:
+        AudioContext() = default;
+        virtual ~AudioContext() = default;
+
+        /// @brief Creates an audio context.
+        /// @return AudioContext, or nullptr on failure.
+        static std::shared_ptr<AudioContext> create();
+
+        /// @brief Enumerates the available devices.
+        /// @param[out] devices Vector to fill with the available devices.
```
RiscadoA, line 61:
```suggestion
        /// @param[out] devices Vector to fill with the available devices' specifiers
```

core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+        /// @brief Creates a new audio device
+        /// @param specifier The specifier of the audio device, used to identify it
+        /// @param listenerCount The number of audio listener to be created by the device's engine.
+        /// @return Handle of the new device
+        virtual AudioDevice createDevice(unsigned int listenerCount,
+                                         const std::string& specifier = "MiniaudioDevice") = 0;
+
+        /// @brief Creates a new audio buffer.
+        /// @param data Data to be written to the buffer, either .wav, .mp3 or .flac.
+        /// @param datSize Size of the data to be written.
```
RiscadoA, line 73:
```suggestion
        /// @param dataSize Size of the data to be written.
```

core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+    {
+        /// @brief Abstract audio buffer.
+        class CUBOS_CORE_API Buffer
+        {
+        public:
+            virtual ~Buffer() = default;
+
+            /// @brief  Gets the length in seconds of the audio buffer.
+            /// @return Length in seconds of the audio buffer.
+            virtual size_t length() = 0;
```
RiscadoA, line 89: Should this be a float then, perhaps? Or do we only have a per-second resolution?

core/src/al/miniaudio_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+    ma_decoder decoder;
+
+    MiniaudioBuffer(const void* data, size_t dataSize)
+    {
+        if (ma_decoder_init_memory(data, dataSize, nullptr, &decoder) != MA_SUCCESS)
+        {
+            CUBOS_ERROR("Failed to initialize Decoder from data");
+        }
+
+        mValid = true;
```
RiscadoA, line 20: It's always gonna be valid, you forgot an else

core/src/al/miniaudio_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+        ma_result result = ma_decoder_get_length_in_pcm_frames(&decoder, &lengthInPCMFrames);
+
+        if (result != MA_SUCCESS)
+        {
+            CUBOS_ERROR("Failed to get the length of audio in PCM frames.");
+            return 0;
+        }
+
+        // Calculate the length in seconds: Length in PCM frames divided by the sample rate.
+        return static_cast<size_t>(lengthInPCMFrames) / decoder.outputSampleRate;
```
RiscadoA, line 40: Better cast to float, otherwise all sub-second sounds are going to have a length of 0.

core/src/al/miniaudio_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+    void setBuffer(Buffer buffer) override
+    {
+        // Try to dynamically cast the Buffer to a MiniaudioBuffer.
+        auto miniaudioBuffer = std::dynamic_pointer_cast<MiniaudioBuffer>(buffer);
+
+        if (miniaudioBuffer == nullptr)
+        {
+            CUBOS_FAIL("Buffer is not of type MiniaudioBuffer.");
+            return;
+        }
```
RiscadoA, line 74:
```suggestion
        // Try to dynamically cast the Buffer to a MiniaudioBuffer.
        auto miniaudioBuffer = std::static_pointer_cast<MiniaudioBuffer>(buffer);
```
I think you can safely assume there won't be other buffer - we probably don't need to cover a scenario where there are more than one audio context implementations coexisting in the same application.

core/include/cubos/core/al/audio_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+            virtual void setPitch(float pitch) = 0;
+
+            /// @brief Sets whether the source plays in a loop.
+            /// @param looping Looping flag.
+            virtual void setLooping(bool looping) = 0;
+
+            /// @brief Sets whether the source position and velocity is relative to the listener or
+            /// not.
+            /// @param relative Relative flag.
+            virtual void setRelative(bool relative) = 0;
```
RiscadoA, line 129: For this to make sense, the source should be pinned to a specific listener right?
Maybe we can pass it here as an argument, and then use `ma_sound_set_pinned_listener_index` in the implementation?

core/src/al/miniaudio_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+std::string MiniaudioContext::getDefaultDevice()
+{
+    // Create a temporary context
+    ma_context context;
+    if (ma_context_init(nullptr, 0, nullptr, &context) != MA_SUCCESS)
+    {
+        CUBOS_ERROR("Failed to initialize audio context for default device lookup");
+        return "";
+    }
```
RiscadoA, line 294: This method should not be static, so that you don't need to create a temporary context.
You can make it a normal method and then use `mContext`.

core/src/al/miniaudio_context.cpp
```diff
@@ -0,0 +1,369 @@
...
+        CUBOS_WARN("No audio playback devices found");
+    }
+}
+
+Buffer MiniaudioContext::createBuffer(const void* data, size_t dataSize)
+{
+    auto buffer = std::make_shared<MiniaudioBuffer>(data, dataSize);
+    if (!buffer->isValid())
+    {
+        CUBOS_ERROR("Failed to create MiniaudioBuffer.");
```
RiscadoA, line 352:
```suggestion
        CUBOS_ERROR("Failed to create MiniaudioBuffer");
```

RiscadoA (approved):
Thank you both for working on this! LGTM after resolving the comments I've just added :relieved:

GalaxyCrush (approved):
LGTM!

Dageus (approved):

PR #1343, opened 2024-10-08
feat(assets): Allowed identifying assets in code from their path

GalaxyCrush:
# Description

Changed some logic to the assets system to identify assets by their path.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/assets/asset.hpp
+++ b/engine/include/cubos/engine/assets/asset.hpp
@@ -30,7 +30,7 @@ namespace cubos::engine
 
         /// @brief Avoid using this field, use @ref getId() instead.
         /// @todo This was added as a dirty fix for #692, should be removed once the issue is fixed.
-        uuids::uuid reflectedId;
+        std::string pathOrId;
 
         ~AnyAsset();
 
@@ -71,9 +71,13 @@ namespace cubos::engine
         /// @return Asset version.
         int getVersion() const;
 
-        /// @brief Gets the UUID of the asset.
+        /// @brief Gets the Path or UUID of the asset.
+        /// @return Asset Path or UUID.
+        std::string getIdString() const;
+
+        /// @brief Gets the UUID of the asset if it is the case of having one.
         /// @return Asset UUID.
-        uuids::uuid getId() const;
+        std::optional<uuids::uuid> getId() const;
 
         /// @brief Checks if the handle is null.
         /// @return Whether the handle is null.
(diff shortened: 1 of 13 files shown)


engine/include/cubos/engine/assets/assets.hpp
```diff
@@ -356,6 +356,11 @@ namespace cubos::engine
         /// @return Lock guard.
         std::unique_lock<std::shared_mutex> lockWrite(const AnyAsset& handle) const;
 
+        /// @brief Checks if the given path is a valid path to an asset.
+        /// @param path Path to check.
+        /// @return A string that says if the given string is a path, uuid or not a valid format string
+        static std::string isPath(const std::string& path);
```
RiscadoA, line 362: Maybe this should be part of the `Asset` handle class, no?
But in any case, I would not use a `std::string` as a return type here: in these cases, either stick to booleans or create your own enums, otherwise it becomes very error prone.

engine/src/assets/asset.cpp
```diff
@@ -40,17 +40,21 @@ AnyAsset::AnyAsset(std::string_view str)
 {
     if (auto id = uuids::uuid::from_string(str))
     {
-        reflectedId = id.value();
+        pathOrId = str;
         mId = id.value();
     }
+    else if (str.find('/') != std::string_view::npos || str.find('\\') != std::string_view::npos)
```
RiscadoA, line 46:
```suggestion
    else if (!str.empty() && str[0] == '/')
```
Paths in the virtual filesystem always use `/`, and they must also start with `/` in this case, so you can do something like this ^

RiscadoA (changes requested):
Thank you very much for working on this! Pointed out some nitpicks :relieved:

Docas95 (approved):

engine/include/cubos/engine/assets/asset.hpp
```diff
@@ -12,6 +12,14 @@
 
 namespace cubos::engine
 {
+    /// @brief Type of the asset identifier.
+    enum class IdType
+    {
+        UUID,
+        Path,
+        Invalid
+    };
```
RiscadoA, line 21: To avoid polluting the whole `cubos::engine` namespace, we usually prefix these types with the relevant keywords (e.g. `Asset`) or move them within the class itself, so that it becomes something like `AnyAsset::IdType`.

RiscadoA (approved):
Added one last comment, other than that, LGTM :) feel free to merge after resolving it

kuukitenshi (approved):
LGTM

PR #1324, opened 2024-09-24
refactor(tools): move most tools to the engine

RiscadoA:
# Description

So this PR looks big and scary but it's nothing ground breaking - I didn't implement any features here.
I just moved most of the tools in Tesseratos to the engine, under the directory `tools`.

Previously, Tesseratos was a library, which meant we could just use its utilities in games (such as Scraps vs Zombies) by linking Tesseratos as a dependency.
With 0.3, Tesseratos became an actual executable, which meant its tools became inaccessible.
The goal here is to simply move the tools which can be useful even in a deployed game to the engine library itself, while keeping actual editor stuff, such as the asset explorer and scene editor, on the editor itself.

## Checklist

- [ ] Self-review changes.
- [x] Evaluate impact on the documentation. (I had to change a lot of stuff!)
- [x] ~~Ensure test coverage.~~ No new features.
- [x] ~~Write new samples.~~ No new features, although I added the tools to the CuboSurfers plugin.
- [x] Add entry to the changelog's unreleased section.

PR #1344, opened 2024-10-09
fix(engine): add missing CUBOS_ENGINE_API to new tool functions

RiscadoA:
# Description

Turns out in #1324 I forgot to add the CUBOS_ENGINE_API macro to most of the tool functions.
This leads to weird crashes when building with shared libraries enabled (e.g. plugin missing error)
This PR just fixes that.

## Checklist

- [x] Self-review changes.


RiscadoA:
Sorry por the pings guys! I'll just go YOLO on this one, since there's really much to review here and we have few time.

PR #1347, opened 2024-10-14
fix: two small fixes to crash causing bugs

SrGesus:
# Description

- Play Pause plugin was missing a ImGui::End() call when ImGui::Begin() returns false.

- ECS would crash when an onRemove observer made a change to the entity

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


SrGesus:
noticed identical problem in World::destroy

Issue #1351, opened 2024-10-21
-O3 flag causes Compiler Error on ECS QueryFilter


SrGesus:
# Description

Sometimes the compiler doesn't seem to be able to infer that a QueryFilter's mNodeCount is lesser or equal than MaxCursorCount, this results in a compiler warning, and therefore a compiler error.

# Reproduce
1. Clone the repo
2. `cmake -S. -Bbuild -D CMAKE_CXX_COMPILER=c++`
3. `cmake --build build -j 12`

# Expected behaviour
No compiler error.

# Screenshots
```
[ 26%] Building CXX object core/CMakeFiles/cubos-core.dir/src/ecs/query/filter.cpp.o
[...]

PR #1352, opened 2024-10-22
-O3 flag causes compiler error on ECS QueryFilter

SrGesus:
# Description

Added an assert that mNodeCount is smaller than the array size.
Had to also add some kind of "aliasing" to mNodeCount because the compiler thinks it might be mutated.
```cpp
 138   │     int mNodeCount = this->mNodeCount;
 139   │     CUBOS_ASSERT(mNodeCount <= QueryNode::MaxCursorCount);
 140   │     for (int i = 1; i < mNodeCount; ++i)
 141   │     {
 142   │         mNodePins[i] = mNodePins[i - 1] | mNodes[i - 1]->pins();
 143   │     }
```

## Context
- #1351

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.

### DIFF
--- a/core/src/ecs/query/filter.cpp
+++ b/core/src/ecs/query/filter.cpp
@@ -134,7 +134,10 @@ void QueryFilter::update()
 
     // Update the pin masks with which each node will run.
     mNodePins[0] = 0;
-    for (int i = 1; i < mNodeCount; ++i)
+    // Copy mNodeCount since the compiler thinks it will change. See issue #1351
+    int nodeCount = this->mNodeCount;
+    CUBOS_ASSERT(nodeCount <= QueryNode::MaxCursorCount);
+    for (int i = 1; i < nodeCount; ++i)
     {
         mNodePins[i] = mNodePins[i - 1] | mNodes[i - 1]->pins();
     }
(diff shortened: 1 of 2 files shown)


core/src/ecs/query/filter.cpp
```diff
@@ -134,6 +134,9 @@ void QueryFilter::update()
 
     // Update the pin masks with which each node will run.
     mNodePins[0] = 0;
+    // Copy mNodeCount since the compiler thinks it will change.
+    int mNodeCount = this->mNodeCount;
+    CUBOS_ASSERT(mNodeCount <= QueryNode::MaxCursorCount);
     for (int i = 1; i < mNodeCount; ++i)
```
RiscadoA, line 140:
```suggestion
    // Copy mNodeCount since the compiler thinks it will change. See issue TODO add number here
    int nodeCount = this->mNodeCount;
    CUBOS_ASSERT(nodeCount <= QueryNode::MaxCursorCount);
    for (int i = 1; i < nodeCount; ++i)
```
I would just change the naming of the variable (since we reserve the m prefix for member variables), and maybe link the issue you created in the comment above.

RiscadoA (approved):
Ty very much for chasing these bugs :pray: LGTM other than the comment

RiscadoA (approved):

joaomanita (approved):

diogomsmiranda (approved):
LGTM

PR #1357, opened 2024-10-29
1330 add active component for all kinds of disabling purposes

### LINKED ISSUE #1330: Add Active component for all kinds of disabling purposes
# Problem
We currently have the `active` field in the camera component. In the future, it would also be nice to have a similar field for light types, for example. This would lead to duplicating this field all over the place though.
It wouldn't also be possible to treat the 'disableable' entities in a generic way.

# Solution
Instead of duplicating this field all over the place, we could move it to its own component, `Active`.
We could add it to a new `activePlugin` in the engine which does nothing but add it.
The camera plugin should be changed to use this plugin.
It would also be nice to use this opportunity to add toggleable lights, by skipping shading for lights whose `Active` component is set to false.

In my opinion, `Active` components should be optional. If they're not present, the entity should be considered to be active.

RodrigoVintem:
# Description

add active component for all kinds of disabling purposes

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


engine/include/cubos/engine/render/active_component/active_component.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <cubos/core/reflection/reflect.hpp>
+
+namespace cubos::engine
+{
+    struct ActiveComponent
```
RiscadoA, line 7:
```suggestion
    struct Active
```

RiscadoA, line 7: Missing /// documentation - check out other files.

engine/include/cubos/engine/render/active_component/active_component.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
RiscadoA, line 1: Missing file-level documentation

engine/CMakeLists.txt
```diff
@@ -169,6 +169,8 @@ set(CUBOS_ENGINE_SOURCE
 	"src/render/shadow_atlas_rasterizer/shadow_atlas_rasterizer.cpp"
 	"src/render/cascaded_shadow_maps/plugin.cpp"
 	"src/render/cascaded_shadow_maps_rasterizer/plugin.cpp"
+	"src/render/active_component/plugin.cpp"
+	"src/render/active_component/active_component.cpp"
```
RiscadoA, line 173:
```suggestion
	"src/render/active/plugin.cpp"
	"src/render/active/active.cpp"
```
You should rename these files to this, which is more in line with the rest of the code base.
Don't forget to do the same for the include files!

engine/src/render/active_component/active_component.cpp
```diff
@@ -0,0 +1,11 @@
+#include <cubos/core/ecs/reflection.hpp>
+#include <cubos/core/reflection/external/primitives.hpp>
+
+#include <cubos/engine/render/active_component/active_component.hpp>
+
+CUBOS_REFLECT_IMPL(cubos::engine::ActiveComponent)
+{
+    return core::ecs::TypeBuilder<ActiveComponent>("cubos::engine::Active_component")
```
RiscadoA, line 8: The reflection type name should match the C++ type exactly

engine/src/render/active_component/plugin.cpp
```diff
@@ -0,0 +1,11 @@
+#include <cubos/engine/render/active_component/active_component.hpp>
+#include <cubos/engine/render/active_component/plugin.hpp>
+
+// Environment not included
```
RiscadoA, line 4:
```suggestion
```

engine/src/render/active_component/plugin.cpp
```diff
@@ -0,0 +1,11 @@
...
+
+using namespace cubos::engine;
```
RiscadoA, line 6:
```suggestion
```
Unused

engine/src/render/camera/camera.cpp
```diff
@@ -7,7 +7,7 @@
 CUBOS_REFLECT_IMPL(cubos::engine::Camera)
 {
     return core::ecs::TypeBuilder<Camera>("cubos::engine::Camera")
-        .withField("active", &Camera::active)
+        //.withField("active", &Camera::active)
```
RiscadoA, line 10: You should remove it entirely

engine/src/render/lights/plugin.cpp
```diff
@@ -7,8 +9,39 @@
...
+
+    cubos.observer("add active component on add SpotLight")
+        .onAdd<SpotLight>()
+        .without<ActiveComponent>()
+        .call([](Commands cmds, Query<Entity> query) {
+            for (auto [ent] : query)
+            {
+                cmds.add(ent, ActiveComponent{});
+            }
+        });
```
RiscadoA, line 46: Btw if you add this you will also need to add an if to skip over these lights where they're used.
Otherwise, it won't do anything. It's a good idea to ask the graphics team about it.

tomas7770 (commented):
What's already implemented LGTM! Here's what should need to be done for the lights. (GitHub doesn't let me add comments to unchanged lines of code so I'll try to do this manually)

**engine/src/render/deferred_shading/plugin.cpp**

In each of the loops that initialize lights in the shader (there is one for each type of light: directional, point, spot), check if the light is active, and `continue` if it's not.
```
for (auto [lightLocalToWorld, light, lightActive, caster] : directionalLights)
{
    if (!lightActive.active)
    {
        continue;
    }
    auto& perLight = perScene.directionalLights[perScene.numDirectionalLights++];
```

**engine/src/render/shadow_atlas/plugin.cpp**

No need to reserve space in the shadow atlas for lights that aren't active.
```
for (auto [caster, active] : casters)
{
    if (!active.active)
    {
        continue;
    }
    bool foundSlot = false;
```

The entity that has the caster is the same one that has the light, so adding the `Active` component to the caster query allows checking if the light is active.
```
.call([](ShadowAtlas& atlas, Query<SpotShadowCaster&, const Active&> casters) {
```

**engine/src/render/cascaded_shadow_maps/plugin.cpp**

Same as above.
```
for (auto [caster, active] : query)
{
    if (!active.active)
    {
        continue;
    }
    // Remove shadow maps for cameras that no longer exist
```

**engine/src/render/shadow_atlas_rasterizer/plugin.cpp**

No need to draw shadows for lights that aren't active.
```
for (auto [caster, light, localToWorld, active] : lights)
{
    if (!active.active)
    {
        continue;
    }
    // Get light viewport
```

**engine/src/render/cascaded_shadow_maps_rasterizer/plugin.cpp**

Same as above.
```
for (auto [caster, light, localToWorld, lightActive] : lights)
{
    if (!lightActive.active)
    {
        continue;
    }
    auto& shadowMap = caster.shadowMaps.at(cameraEntity);
```
------
Little sidenote unrelated to the PR.

**engine/src/render/ssao/plugin.cpp** has several includes that don't seem to be used for anything, such as lights and HDR. Maybe these can be removed?

RodrigoVintem:
Added Tomas' changes to the files related to light, that should use the Active component. Also removed the unnecessary includes mentioned by Tomas.

engine/include/cubos/engine/render/camera/camera.hpp
```diff
@@ -20,7 +20,6 @@ namespace cubos::engine
         CUBOS_REFLECT;
 
         /// @brief Whether the camera is drawing to a target.
```
SrGesus, line 22: Should remove this comment too if the field was removed

SrGesus (changes requested):
Otherwise LGTM!

engine/src/tools/debug_camera/plugin.cpp
```diff
@@ -51,7 +52,7 @@ void cubos::engine::debugCameraPlugin(Cubos& cubos)
     cubos.startupSystem("create Debug Camera").call([](Commands commands, State& debugCamera) {
         debugCamera.entity = commands.create()
                                  .named("debug-camera")
-                                 .add(Camera{.active = false})
+                                 .add(Camera{})
```
mkuritsu, line 55:
```suggestion
                                
```
This .add(Camera{}) can be removed then just to simplify things since PerspectiveCamera already adds it

mkuritsu (changes requested):
Overall looks good to me just very small changes to maintain consistency.

Aside from that I'm just concern about some implications of this implementation.

engine/src/render/shadow_atlas/plugin.cpp
```diff
@@ -38,16 +39,17 @@ void cubos::engine::shadowAtlasPlugin(Cubos& cubos)
...
             atlas.slotsMap.clear();
             atlas.slots.push_back(
                 std::make_shared<ShadowAtlas::Slot>(glm::vec2(1.0F, 1.0F), glm::vec2(0.0F, 0.0F), -1));
 
             int id = 1;
 
-            for (auto [caster] : casters)
+            for (auto [caster, active] : casters)
             {
+
```
tomas7770, line 52: Still missing the active check.
```suggestion
if (!active.active)
{
    continue;
}
```

tomas7770 (commented):
There are still some small things that should be changed, but otherwise looks good!

RodrigoVintem:
All the new changes proposed by the reviewers have been made.

tomas7770 (approved):

RiscadoA (changes requested):
Mostly LGTM, pointed out some details that we should sort out before merging.
Ty for working on this :relieved:

RiscadoA:
Since this has been abandoned I will pick it up on the next milestone.

PR #1358, opened 2024-11-05
fix(memory): prevent double destruction in AnyVector::swapMove

RiscadoA:
# Description

This happened only when the last element of an AnyVector was removed in a swapMove.
The destructor of the last element was called in the swapMove itself, and once again on pop.
So, simply moved the extra destructor call to within the if itself.

## Checklist

- [x] Self-review changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Collision detection with VoxelCollisionShapes (#994, **@joaomanita**).
-
-### Added
-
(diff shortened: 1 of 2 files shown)


diogomsmiranda:
Although one thing I notice now is why do we have assertions in the middle of srccode? Shouldn't we have tests with those instead?

RiscadoA:
> Although one thing I notice now is why do we have assertions in the middle of srccode? Shouldn't we have tests with those instead?

They're just sanity checks, it's good practice to write all assumptions of the code in assertions :relieved:

RodrigoVintem (approved):

Dageus (approved):

PR #1362, opened 2024-11-08
feat(settings): add method that saves settings to file

SrGesus:
# Description

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/settings/settings.hpp
+++ b/engine/include/cubos/engine/settings/settings.hpp
@@ -101,6 +101,12 @@ namespace cubos::engine
         /// @param settingsToMerge Settings to be merged to this instance.
         void merge(const Settings& settingsToMerge);
 
+        /// @brief Save settings as json to specified path.
+        /// @param path Virtual `cubos::core::data::FileSystem` path to write settings in.
+        /// @param indent Indent of the Json, -1 for selects the most compact representation.
+        /// @return Whether call was successful.
+        bool save(std::string_view path = "/settings.json", int indent = 2) const;
+
         /// @return Underlying `std::unordered_map` with the settings.
         const std::unordered_map<std::string, std::string>& getValues() const;
 
--- a/engine/include/cubos/engine/settings/settings.hpp
+++ b/engine/include/cubos/engine/settings/settings.hpp
@@ -103,9 +103,14 @@ namespace cubos::engine
 
         /// @brief Save settings as json to specified path.
         /// @param path Virtual `cubos::core::data::FileSystem` path to write settings in.
-        /// @param indent Indent of the Json, -1 for selects the most compact representation.
+        /// @param indent Indent of the Json, -1 for no whitespace.
         /// @return Whether call was successful.
         bool save(std::string_view path = "/settings.json", int indent = 2) const;
+
+        /// @brief Load settings from json in specified path.
+        /// @param path Virtual `cubos::core::data::FileSystem` path to read settings from.
+        /// @return Loaded settings object.
+        static Settings load(std::string_view path = "/settings.json");
 
         /// @return Underlying `std::unordered_map` with the settings.
         const std::unordered_map<std::string, std::string>& getValues() const;
--- a/engine/samples/settings/main.cpp
+++ b/engine/samples/settings/main.cpp
@@ -14,7 +14,7 @@ int main(int argc, char** argv)
 
     cubos.startupSystem("print setting value").after(settingsTag).call([](Settings& settings) {
         CUBOS_INFO("{}", settings.getString("greeting", "Hello!"));
-        settings.save("/settings.json", 4);
+        settings.save();
     });
 
     cubos.startupSystem("print number setting value").after(settingsTag).call([](Settings& settings) {
@@ -22,7 +22,7 @@ int main(int argc, char** argv)
         CUBOS_INFO("{}", settings.getDouble("double", 1e-07));
         CUBOS_INFO("{}", settings.getBool("boolean", false));
         CUBOS_INFO("{}", settings.getString("emptyString", ""));
-        settings.save("/settings.json", 4);
+        settings.save();
     });
 
     cubos.run();
--- a/engine/samples/settings/main.cpp
+++ b/engine/samples/settings/main.cpp
@@ -14,6 +14,15 @@ int main(int argc, char** argv)
 
     cubos.startupSystem("print setting value").after(settingsTag).call([](Settings& settings) {
         CUBOS_INFO("{}", settings.getString("greeting", "Hello!"));
+        settings.save("/settings.json", 4);
+    });
+
+    cubos.startupSystem("print number setting value").after(settingsTag).call([](Settings& settings) {
+        CUBOS_INFO("{}", settings.getInteger("number", 1));
+        CUBOS_INFO("{}", settings.getDouble("double", 1e-07));
+        CUBOS_INFO("{}", settings.getBool("boolean", false));
+        CUBOS_INFO("{}", settings.getString("emptyString", ""));
+        settings.save("/settings.json", 4);
     });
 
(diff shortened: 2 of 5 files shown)


RiscadoA:
If we do end up reimplementing the settings resource as a system argument, this would no longer be necessary :thinking: 
You could just have a `Settings::save` method, which ig would be even better because then you don't need to mess with the weird event system we have and it is also synchronous. Opinions?

SrGesus:
I think we wanted events to be able to do something if a setting changes, e.g reload graphics if you change a graphics setting or something and there we will need to use something similar if not events (like an observer idk).
But yes for saving i don't need events, in fact i can do it without events without reimplementing settings

engine/src/settings/settings.cpp
```diff
@@ -96,6 +103,54 @@ void Settings::merge(const Settings& settingsToMerge)
...
+        if (pEnd != value.second.c_str() + value.second.length() && (value.second != "true" && value.second != "false"))
+        {
+            stream.printf("\n\t\"{}\": \"{}\"", value.first.c_str(), value.second.c_str());
+        }
+        else
+        {
+            stream.printf("\n\t\"{}\": {}", value.first.c_str(), value.second.c_str());
+        }
+    }
+    stream.print("\n}\n");
```
RiscadoA, line 131: I think it is probably a better a idea to turn this into a `nlohmann::json`, and then serialize that value with `json.dump()`. Less code duplication.

engine/src/settings/settings.cpp
```diff
@@ -96,6 +103,54 @@ void Settings::merge(const Settings& settingsToMerge)
...
+}
+
+bool Settings::save()
+{
```
RiscadoA, line 135:
```suggestion
bool Settings::save(std::string_view path)
{
```
I would change this to receive the virtual file system path where to save to, and by default set it to `/settings.json`.
This gives the caller more flexibility. Internally then, you wouldn't mount an archive, you would only keep the last three lines. Since the settings file is already mounted at startup anyway the logic in the beginning is unnecessary

RiscadoA (changes requested):
Thanks for working on this! I would just simplify this a bit before merging.
Also, don't forget the changelog entry :relieved:

RiscadoA (changes requested):
Accidentally dismissed my own review, my bad

engine/src/settings/settings.cpp
SrGesus: I avoided doing that because:
- Serializing the map into a nlohmann::json sounds like an expensive process with a lot of allocations just to then Serialize it again into a string, compared to just iterating the values
- That will add quotation marks to every value :/

diogomsmiranda:
What is the status on this? Is this ready for review?

engine/src/settings/settings.cpp
RiscadoA: Regarding the first point, performance isn't really an issue here. We won't be saving the settings every frame thousands of times! So, IMO readability/maintainability is the priority here.

Regarding the second point, not necessarily; if you do something like `nlohmann::json json = 32`, it dumps you an integer, without the quotation marks. It will only put quotation marks if you do `json = "32"`, so it's just a matter of doing it like you're doing right now - checking if it is a number of a boolean

SrGesus: But we don't have integers we just have a map of string -> string so every variable is a string

SrGesus: Ah you're saying making a new json, iterating the map and adding each element individually, converting it into a double or a boolean?

RiscadoA: Sure, but if it is a number you can just use the return value of `strtod` no? And for the booleans, you can do `json = value == "true";`

SrGesus:
> ready
The settings part should be done now. I just intend on adding a few tests and the changelog.

engine/include/cubos/engine/settings/settings.hpp
```diff
@@ -101,6 +101,17 @@ namespace cubos::engine
         /// @param settingsToMerge Settings to be merged to this instance.
         void merge(const Settings& settingsToMerge);
 
+        /// @brief Save settings as json to specified path.
+        /// @param path Virtual `cubos::core::data::FileSystem` path to write settings in.
+        /// @param indent Indent of the Json, -1 for selects the most compact representation.
```
RiscadoA, line 106:
```suggestion
        /// @param indent Indent of the Json, -1 for no whitespace.
```

engine/include/cubos/engine/settings/settings.hpp
```diff
@@ -101,6 +101,17 @@ namespace cubos::engine
...
+        /// @return Whether call was successful.
+        bool save(std::string_view path = "/settings.json", int indent = 2) const;
+
+        /// @brief Load settings from json in specified path.
+        /// @param path Virtual `cubos::core::data::FileSystem` path to read settings from.
+        /// @return Whether call was successful.
```
RiscadoA, line 112:
```suggestion
        /// @return Loaded settings object.
```

engine/samples/settings/main.cpp
```diff
@@ -14,6 +14,15 @@ int main(int argc, char** argv)
 
     cubos.startupSystem("print setting value").after(settingsTag).call([](Settings& settings) {
         CUBOS_INFO("{}", settings.getString("greeting", "Hello!"));
+        settings.save("/settings.json", 4);
```
RiscadoA, line 17:
```suggestion
        settings.save();
```
Maybe omit the arguments here? :thinking: or do you want to showcase them?

engine/samples/settings/main.cpp
```diff
@@ -14,6 +14,15 @@ int main(int argc, char** argv)
...
+    });
+
+    cubos.startupSystem("print number setting value").after(settingsTag).call([](Settings& settings) {
+        CUBOS_INFO("{}", settings.getInteger("number", 1));
+        CUBOS_INFO("{}", settings.getDouble("double", 1e-07));
+        CUBOS_INFO("{}", settings.getBool("boolean", false));
+        CUBOS_INFO("{}", settings.getString("emptyString", ""));
+        settings.save("/settings.json", 4);
```
RiscadoA, line 25: Same as above

engine/src/settings/settings.cpp
```diff
@@ -96,6 +108,120 @@ void Settings::merge(const Settings& settingsToMerge)
...
+    }
+}
+
+Settings Settings::load(std::string_view path)
+{
+    // If the settings file is not mounted.
+    if (FileSystem::find(path) == nullptr)
+    {
+        return {};
+    }
```
RiscadoA, line 188: I think you shoud just check if the `stream` returned by `open` is `nullptr`.
E.g., the path may point to a directory instead of a file. In that case, this will succeed, but open will return nullptr

engine/src/settings/settings.cpp
```diff
@@ -96,6 +108,120 @@ void Settings::merge(const Settings& settingsToMerge)
...
+
+bool Settings::save(std::string_view path, int indent) const
+{
+    // If the settings file is not mounted.
+    if (FileSystem::find(path) == nullptr)
+    {
+        return false;
+    }
+
+    auto stream = FileSystem::open(path, File::OpenMode::Write);
```
RiscadoA, line 143: I think you shoud just check if the `stream` returned by `open` is `nullptr`.
E.g., the path may point to a directory instead of a file. In that case, this will succeed, but open will return nullptr

RiscadoA (approved):
Other than the comments I pointed out, LGTM!

RiscadoA:
Also don't forget to update the CHANGELOG!

RodrigoVintem (approved):

PR #1366, opened 2024-11-16
Add point light shadows

tomas7770:
# Description

Implements shadows for point lights.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [x] Write new samples.


engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -337,13 +344,81 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
...
+                                                                glm::vec3(1.0F / lightLocalToWorld.worldScale())));
+                                    break;
+                                default:
+                                    lightView = glm::inverse(glm::scale(
+                                        lightLocalToWorld.mat, glm::vec3(1.0F / lightLocalToWorld.worldScale())));
+                                    break;
+                                }
+
+                                perLight.matrices[i] = lightProj * lightView;
+                            }
```
RiscadoA, line 412: I think you can simplify this section of code a bit - the for barely does anything, why not unfold it and remove the switch case entirely? One other option (probably better) would be to move the values which change for each case into an array. E.g.
```cpp
static const lightViewAngle[6] = { 0.0F, 180.0F, ...};
static const glm::vec3 lightViewAxis[6] = { ... };
... // use these within the for loop, avoid the switch

File: engine/src/render/shadow_atlas_rasterizer/plugin.cpp [hunk 2]
```diff
@@ -32,6 +34,10 @@ namespace
     {
         glm::mat4 lightViewProj;
     };
+    struct PerSceneCube
```
RiscadoA, line 37:
```suggestion
    };

    struct PerSceneCube
```

engine/src/render/shadow_atlas_rasterizer/plugin.cpp
```diff
@@ -208,5 +236,105 @@ void cubos::engine::shadowAtlasRasterizerPlugin(Cubos& cubos)
...
+                            glm::scale(glm::rotate(localToWorld.mat, glm::radians(-90.0F), glm::vec3(1.0F, 0.0F, 0.0F)),
+                                       glm::vec3(1.0F / localToWorld.worldScale())));
+                        break;
+                    default:
+                        view = glm::inverse(glm::scale(localToWorld.mat, glm::vec3(1.0F / localToWorld.worldScale())));
+                        break;
+                    }
+
+                    perScene.lightViewProj[i] = proj * view;
+                }
```
RiscadoA, line 313: This is the same exact math as before right? In that case, it would probably be a good idea to move it to a common place, so that you don't define it twice

RiscadoA (changes requested):
Very good work on this :rocket: :eyes: that was quick
Other than the comments I added, I feel like it would be a good idea to split the shadow atlas resource and plugin into two if you implement it like this.
The code is already quite complex as it is, and with this you're essentially merging two different rendering phases into a single one. It would be simpler to read and understand if they were separate, I think.
Maybe a `spot_shadow_atlas` and a `point_shadow_atlas`? Or should we keep it as is?

Docas95 (approved):

Scarface1809 (approved):

core/src/geom/utils.cpp
```diff
@@ -191,3 +192,16 @@ void cubos::core::geom::getCameraFrustumCorners(const glm::mat4& view, const glm
...
+{
+    static const float ViewAngle[6] = {0.0F, 180.0F, 90.0F, -90.0F, 90.0F, -90.0F};
+    static const glm::vec3 ViewAxis[6] = {glm::vec3(0.0F, 1.0F, 0.0F), glm::vec3(0.0F, 1.0F, 0.0F),
+                                          glm::vec3(0.0F, 1.0F, 0.0F), glm::vec3(0.0F, 1.0F, 0.0F),
+                                          glm::vec3(1.0F, 0.0F, 0.0F), glm::vec3(1.0F, 0.0, 0.0F)};
+
+    for (int i = 0; i < 6; i++)
+    {
+        cubeViewMatrices.push_back(glm::inverse(glm::rotate(inverseView, glm::radians(ViewAngle[i]), ViewAxis[i])));
+    }
```
RiscadoA, line 206:
```suggestion
    
    cubeViewMatrices.resize(6);
    for (int i = 0; i < 6; i++)
    {
        cubeViewMatrices[i] = glm::inverse(glm::rotate(inverseView, glm::radians(ViewAngle[i]), ViewAxis[i]));
    }
```
Just to avoid some memory allocations :relieved:

RiscadoA (approved):
LGTM!

PR #1370, opened 2024-11-18
feat(rendering): Add Shadow Normal Offset Bias in the deferred shading

GalaxyCrush:
# Description

I added the Shadow Normal Offset Bias to the deferred shading and gave the user the option to use or not use this bias.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/assets/render/deferred_shading.fs
+++ b/engine/assets/render/deferred_shading.fs
@@ -25,6 +25,7 @@ struct DirectionalLight
     vec4 shadowFarSplitDistances[3]; // intended to be a float array, but std140 layout aligns array elements
                                      // to vec4 size; number of vec4s = ceiling(MaxCascades / 4 components)
     int numCascades;
+    float normalOffsetScale;
 };
 
 struct PointLight
@@ -49,6 +50,7 @@ struct SpotLight
     vec2 shadowMapSize;
     float shadowBias;
     float shadowBlurRadius;
+    float normalOffsetScale;
 };
 
 layout(std140) uniform PerScene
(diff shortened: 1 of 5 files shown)


RodrigoVintem (approved):
LGTM

engine/assets/render/deferred_shading.fs
```diff
@@ -80,13 +82,28 @@ float remap(float value, float min1, float max1, float min2, float max2)
     return max2 + (value - min1) * (max2 - min2) / (max1 - min1);
 }
 
+vec3 ApplyNormalOffset(vec3 normal, vec3 lightDir, vec3 position) {
```
RiscadoA, line 85:
```suggestion
vec3 applyNormalOffset(vec3 normal, vec3 lightDir, vec3 position) {
```

Just to be consistent with the rest of the shader

engine/assets/render/deferred_shading.fs
```diff
@@ -80,13 +82,28 @@ float remap(float value, float min1, float max1, float min2, float max2)
...
+
+    float prod = dot(normal, lightDir);
+    prod = max(0.0, prod); 
+    
+    float SlopeScale = 0.02;
+    float NormalOffsetScale = 0.007;
```
RiscadoA, line 91:
```suggestion
    float slopeScale = 0.02;
    float normalOffsetScale = 0.007;
```

engine/include/cubos/engine/render/shadows/caster.hpp
```diff
@@ -18,8 +18,9 @@ namespace cubos::engine
     {
         CUBOS_REFLECT;
 
-        float bias = 0.002F;     ///< Shadow bias.
-        float blurRadius = 1.0F; ///< Shadow blur radius.
-        int id = -1;             ///< Caster id for the shadow atlas.
+        float bias = 0.001F;               ///< Shadow bias.
+        float blurRadius = 0.005F;         ///< Shadow blur radius.
```
RiscadoA, line 22: What's reason behind changing this? Does this give better results?

RiscadoA (approved):
Thanks for working on this! LGTM

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -60,7 +60,8 @@ namespace
                                           4]; // intended to be a float array, but std140 layout aligns array elements
                                               // to vec4 size; number of vec4s = ceiling(MaxCascades / 4 components)
         int numCascades;
-        int padding2[3];
+        int useShadowNormalOffsetBias;
+        int padding2[2]; // padding to 16 bytes
```
mkuritsu, line 64:
```suggestion
        bool useShadowNormalOffsetBias;
        char padding2[11]; // padding to 16 bytes
```
Maybe in here you could use a `bool` instead and have the padding array be of `char` type to allow the same padding but making the variable type better suited.

engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -86,7 +87,8 @@ namespace
         glm::vec2 shadowMapSize;
         float shadowBias;
         float shadowBlurRadius;
-        float padding[2];
+        int useShadowNormalOffsetBias;
+        float padding[1];
```
mkuritsu, line 91: same thing as above here

mkuritsu (approved):
LGTM! Just a small change in the variable types

engine/src/render/deferred_shading/plugin.cpp
RiscadoA: In GLSL, [bools declared in uniform blocks](http://www.opengl.org/registry/doc/glspec45.core.pdf#page=157) are represented as unsigned integers (32-bit wide). So the correct type would actually be `unsigned int`. Although `int` should be fine

tomas7770 (approved):
LGTM! Just tested some of the demos from the cubos-demo repo and noticed significant improvements!

engine/include/cubos/engine/render/shadows/caster.hpp
```diff
@@ -18,8 +18,9 @@ namespace cubos::engine
...
+        int id = -1;                       ///< Caster id for the shadow atlas.
+        bool useShadowNormalOffset = true; ///< Whether to use the shadow normal offset.
```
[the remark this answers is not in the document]
RiscadoA: That would be great - it would also mean avoiding one more branch, which is always a performance hog on gpus

engine/include/cubos/engine/render/shadows/caster.hpp
GalaxyCrush: I was testing different values of bias to see the results, and I think this bias gives better results. The blurRadius was only for me to see the shadows without much blur to look for shadow acne artifacts.

tomas7770 (approved):

PR #1372, opened 2024-11-19
fix: Fixed delta time not used in some physics systems

joaomanita:
# Description

## Checklist

- [X] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


RiscadoA (approved):
Looks good to me! Don't forget to also cover what @fallenatlas talked about in Discord.
Also, add a CHANGELOG entry!

Dageus (approved):
lgtm!

RiscadoA:
Btw please change the commit msg before merging to something like `fix(physics): do fixed updates correctly`

engine/src/fixed_step/plugin.cpp
```diff
@@ -26,7 +26,9 @@ void cubos::engine::fixedStepPlugin(Cubos& cubos)
         return false;
     });
 
-    cubos.system("accumulate time resource").before(fixedStepTag).call([](AccumulatedTime& timer, const DeltaTime& dt) {
-        timer.value += dt.value();
-    });
+    cubos.system("accumulate time resource")
+        .before(fixedStepTag)
+        .call([](AccumulatedTime& timer, const DeltaTime& dt, const FixedDeltaTime& step) {
+            timer.value = std::min(timer.value + dt.value(), 2 * step.value);
```
fallenatlas, line 32: This is actually a really neat way to do it. I actually only thought this because it was put in this way, maybe we should always keep the previous timer.value. I'm thinking that since this is to prevent explosive behavior coming from DeltaTime being too big on that frame, we only need to limit that time and add it to the accumulator, this way on the next frame that doesn't have a huge delta time, we accumulate on that previous value instead of 0. I have no idea if it makes a difference. What do you think?
```suggestion
            timer.value = += std::min(dt.value(), 2 * step.value);
```

fallenatlas (approved):
LGTM! 👍

GCeSilva (approved):
lgtm!

PR #1376, opened 2024-11-21
Fix crash when using ImGui across DLL boundaries on Windows

RiscadoA:
# Description

You can read the documentation I included in the PR changes to understand what I'm fixing here.
If you're running Windows, previously with `CUBOS_ENGINE_SHARED` set to `ON` on CMake you probably got a crash on both Tesseratos and on the ImGui sample. With this change, that should no longer happen, so **please test that before approving**.

## Checklist

- [x] Self-review changes.
- [x] Updated imgui plugin documentation.
- [x] Update sample.


diogomsmiranda:
If @SrGesus and @jdbaracho are Windows users then I don't mind looking at the code and review it, else I think we should assign Windows users as reviewers no? I have no machine to test this on.

@RiscadoA thoughts?

RiscadoA:
> Is this about `CUBOS_ENGINE_SHARED` or `BUILD_ENGINE_SHARED` @RiscadoA ?

The first one, the second one no longer exists, it was the old name of the same option which was renamed in a previous PR

jdbaracho (approved):
Tested on a Windows machine, at least the ImGUI sample and cubosurfers seem to run without any problems.
LGTM!

SrGesus (approved):

Issue #537, opened 2023-09-12
Allow querying for changed components


RiscadoA:
Related to #273.

It would be useful to be able to filter the results of a query such that only components which were modified are seen.
In the transform plugin, for example, this could be used to calculate only the transform matrix when necessary.

The syntax for this is still unclear, maybe something like:
```cpp
static system(Query<Write<LocalToWorld>, Read<Transform>, Changed<Transform>> query)
{
    for (auto [entity, localToWorld, transform, changed] : query)
    {
         // 'changed' isn't really used at all, maybe we should skip returning it in the tuple somehow?
         // ...
    }
}
```

PR #1378, opened 2024-11-21
feat(ui): add UI text rendering

mkuritsu:
# Description

Added initial support for text in the UI system including font loading, atlas generation and text rendering.

# Missing features

- Currently only ASCII characters are supported, Unicode with dynamic font atlas creation needs to be implemented.
- Currently the wrapping functionality is not implemented so the text will overflow out of the `UIElement` bounds which can cause some miss alignment in the intended position.

# Known Issues

- Related to #537 currently its not supported to change the text from the component, in the mean time in order to have the changes reflected you could either remove and then add the `UIText` component with the updated text or call `.reset()` in `UiText.va` to force the recreation of the buffer.

## Checklist

- [x] Self-review changes.
- [x] Updated existing sample.


engine/include/cubos/engine/ui/text/text.hpp
```diff
@@ -0,0 +1,48 @@
...
+        float fontSize{24.0F};
+
+        /// @brief The font to be used.
+        Asset<Font> font{AnyAsset("93cbe82e-9c9b-4c25-aa55-5105c1afd0cc")};
+
+        core::gl::VertexArray va{nullptr};
+
+        std::shared_ptr<FontAtlas> atlas{nullptr};
+
+        size_t vertexCount{0};
```
RiscadoA, line 46: If the user isn't meant to mess with those these should be private (and the struct made a class) I think :thinking: otherwise it gets kinda confusing. You can add methods so that you can interact with them from the plugin itself

engine/include/cubos/engine/ui/text/text_stretch.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component which makes a UI element fit the text it has.
+    /// @ingroup ui-text-plugin
+    struct CUBOS_ENGINE_API UITextStretch
+    {
+        CUBOS_REFLECT;
+    };
```
RiscadoA, line 18: @DiogoMendonc-a is this what you envisioned?

engine/samples/ui/main.cpp
```diff
@@ -61,7 +67,7 @@ int main(int argc, char** argv)
         /// [Set up Canvas]
         /// [Set up Background]
         auto elementBg = commands.create()
-                             .add(UIElement{})
+                             .add(UIElement{.layer = -2})
```
RiscadoA, line 70: Probably easier to understand if we use positive layers in the sample

engine/src/font/atlas.cpp
```diff
@@ -0,0 +1,57 @@
...
+    int width = 0;
+    int height = 0;
+    packer.getDimensions(width, height);
+
+    msdf_atlas::ImmediateAtlasGenerator<float, 3, &msdf_atlas::msdfGenerator,
+                                        msdf_atlas::BitmapAtlasStorage<msdf_atlas::byte, 3>>
+        generator(width, height);
+    msdf_atlas::GeneratorAttributes attributes;
+    generator.setAttributes(attributes);
+    generator.setThreadCount(4);
```
RiscadoA, line 43: Lots of magic numbers here! Some of those should probably be in constants, and others should be configurable by the user when creating a font atlas. If we do go through that 'FontAtlasBridge' I mentioned in the other comment, I guess we could fetch those options from the .meta

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow identifying assets in code from their path (#1177. **@GalaxyCrush**).
 - Added an Audio asset (#230, **@Dageus**, **@diogomsmiranda**).
 - Hold function for debug_camera (#1030, **@jdbaracho**).
+- Added UIText component (#1300, **@mkuritsu**)
```
RiscadoA, line 16:
```suggestion
- UI text element using MSDF text rendering (#1300, **@mkuritsu**)
```
This makes it seem like a small feature!

engine/src/font/font.cpp
```diff
@@ -0,0 +1,55 @@
...
+    Font::Font(msdfgen::FreetypeHandle* ft, core::memory::Stream& stream)
+    {
+        if (ft == nullptr)
+        {
+            CUBOS_ERROR("Couldn't load font: Invalid Freetype handle");
+            return;
+        }
+        stream.seek(0, cubos::core::memory::SeekOrigin::End);
+        size_t streamSize = stream.tell();
+        auto* contents = static_cast<msdfgen::byte*>(operator new(sizeof(msdfgen::byte) * streamSize));
```
RiscadoA, line 25: You can just use a vector here: `std::vector<msdfgen::byte> contents{streamSize};`
Safer and less scary

engine/src/font/font.cpp
```diff
@@ -0,0 +1,55 @@
...
+        stream.seek(0, cubos::core::memory::SeekOrigin::Begin);
+        if (stream.read(contents, streamSize) < streamSize)
+        {
+            CUBOS_ERROR("Couldn't load font: Read less than stream length");
+            fontHandle = nullptr;
+            operator delete(contents);
+            return;
+        }
+        fontHandle = msdfgen::loadFontData(ft, contents, static_cast<int>(streamSize));
```
RiscadoA, line 34: If you change it to a vec
```suggestion
        fontHandle = msdfgen::loadFontData(ft, contents.data(), static_cast<int>(streamSize));
```

engine/src/font/font.cpp
```diff
@@ -0,0 +1,55 @@
...
+        if (ft == nullptr)
+        {
+            CUBOS_ERROR("Couldn't load font: Invalid Freetype handle");
+            return;
+        }
+        stream.seek(0, cubos::core::memory::SeekOrigin::End);
+        size_t streamSize = stream.tell();
+        auto* contents = static_cast<msdfgen::byte*>(operator new(sizeof(msdfgen::byte) * streamSize));
+        stream.seek(0, cubos::core::memory::SeekOrigin::Begin);
+        if (stream.read(contents, streamSize) < streamSize)
```
RiscadoA, line 27: If you change it to a vec
```suggestion
        if (stream.read(contents.data(), streamSize) < streamSize)
```

CHANGELOG.md
RiscadoA: Also, you can add an entry to the fixes section with something like `- Stopped undeterministic behavior when drawing transparent UI elements by drawing layer by layer (**@mkuritsu**)`

engine/src/ui/canvas/plugin.cpp
```diff
@@ -274,7 +285,7 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
                  Query<Entity, UICanvas&, RenderTarget&> canvasQuery, const core::io::Window& window,
                  const State& state) {
             // For each render target with a UI canvas.
-            window->renderDevice().setDepthStencilState(state.dss);
```
RiscadoA, line 277: You still need to set the depth stencil state! Otherwise it will use the last pass' state. Set it one where depth testing is disabled.

engine/src/ui/canvas/plugin.cpp
```diff
@@ -139,7 +151,7 @@ void cubos::engine::uiCanvasPlugin(Cubos& cubos)
                     canvas.virtualSize = canvas.referenceSize;
                 }
 
-                canvas.mat = glm::ortho<float>(0, canvas.virtualSize.x, 0, canvas.virtualSize.y);
+                canvas.mat = glm::ortho<float>(0, canvas.virtualSize.x, 0, canvas.virtualSize.y, -1000.0F, 1000.0F);
```
RiscadoA, line 154: ? Why limit it?

RiscadoA (changes requested):
This was just a first pass - it's a very big PR! Great job on this, tyvm for working on it :partying_face:

engine/CMakeLists.txt
```diff
@@ -237,6 +248,25 @@ add_library(stb_image INTERFACE)
...
+FetchContent_MakeAvailable(freetype)
+
+set(MSDF_ATLAS_USE_VCPKG OFF)
+set(MSDF_ATLAS_USE_SKIA OFF)
+FetchContent_Declare(msdf-atlas-gen
+    GIT_REPOSITORY "https://github.com/Chlumsky/msdf-atlas-gen.git"
+)
+FetchContent_MakeAvailable(msdf-atlas-gen)
+set_property(TARGET msdf-atlas-gen PROPERTY POSITION_INDEPENDENT_CODE ON)
+set_property(TARGET msdfgen-core PROPERTY POSITION_INDEPENDENT_CODE ON)
```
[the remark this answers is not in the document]
mkuritsu: I added it because it was throwing an error when compiling cubos as a shared lib about realocations, I will add a comment stating that then

engine/assets/font/Roboto-Regular.ttf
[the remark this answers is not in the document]
mkuritsu: I'm not to knowledge either but was checking on google fonts website and Roboto seems to use apache 2.0 (https://fonts.google.com/specimen/Roboto/license) I think we would only need to add the license in the repo with it if I'm not mistaken.

Regarding licenses another thing to consider is MSDF, it uses the FreeType library internally (and I needed to add it in the CMakeLists because Windows doesn't have it in the system like other OS'es) and that library also as some different licensing, it has a dual license that we can choose one or the other, GPL v2 or their own license (http://freetype.org/license.html)

engine/src/ui/canvas/plugin.cpp
mkuritsu: Without the values set only elements in the layer 0 where being shown, I think it is probably due to the default value for zNear and zFar

mkuritsu:
Most of the requests handled, just kept the ones related to the FontAtlas to solve since probably that whole class will suffer some changes.

engine/include/cubos/engine/ui/text/text_stretch.hpp
mkuritsu: I added this component because sometimes we could want to have a text displayed on screen and the size of the UIElement to be the whole text, it would be annoying to need to set manually the correct size and could be prone to errors specially if the font change since different fonts would have different spacing.

This component is used in the sample, the title with "cubosengine.org" has this component because the size needs to be perfectly fit to the text to allow proper positions calculations with the anchor and pivot since these depend on the size of the element.

engine/include/cubos/engine/ui/text/plugin.hpp
tomas7770, line 1: Plugin dependencies should be documented here.

engine/assets/ui/text_element.vs
tomas7770, line 1: Since this file and `text.fs` are used together for text rendering, is there any reason why they have different names (apart from the extension)? When reviewing I initially tried to look for a `text_element.fs` and `text.vs` until I realized they didn't exist.

tomas7770 (approved):
LGTM! Just some nitpicking...

GalaxyCrush (approved):
After all the reviews LGTM!! You could use the asset path instead of the ID, so maybe it would be easier to identify the asset. But it's nothing important.

engine/include/cubos/engine/font/atlas/bridge.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <cubos/core/gl/render_device.hpp>
+
+#include <cubos/engine/assets/bridge.hpp>
+#include <cubos/engine/assets/bridges/file.hpp>
+#include <cubos/engine/font/atlas/atlas.hpp>
+#include <cubos/engine/font/font.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref FontAtlas assets.
```
RiscadoA, line 16:
```suggestion
    /// @brief Bridge which loads @ref FontAtlas assets.
```

engine/include/cubos/engine/font/atlas/glyph.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Struct that holds glyph data such as texcoords, position offsets and advance.
+    ///
+    /// @ingroup font-plugin
+    struct CUBOS_ENGINE_API Glyph
```
RiscadoA, line 18: I would maybe rename this to `FontGlyph`, to avoid naming collisions. The file can keep the same name though :relieved:

engine/include/cubos/engine/font/bridge.hpp
```diff
@@ -0,0 +1,29 @@
...
+/// @ingroup font-plugin
+
+#pragma once
+
+#include <cubos/engine/assets/bridges/file.hpp>
+#include <cubos/engine/font/font.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Bridge which loads and saves @ref Font assets.
```
RiscadoA, line 12:
```suggestion
    /// @brief Bridge which loads @ref Font assets.
```

engine/include/cubos/engine/font/font.hpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// @brief Asset containing font data.
+    ///
+    /// @ingroup font-plugin
+    class CUBOS_ENGINE_API Font
+    {
+    public:
+        CUBOS_REFLECT;
+
+        Font(void* freetypeHandle, core::memory::Stream& stream);
```
RiscadoA, line 24: Missing documentation

engine/include/cubos/engine/font/font.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <vector>
+
+#include <cubos/core/memory/stream.hpp>
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Asset containing font data.
```
RiscadoA, line 16: I would explain a little bit better what this asset represents, so that people can understand the difference from the `FontAtlas` asset.

engine/include/cubos/engine/ui/text/text.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+        /// @brief The font size to draw the characters.
+        float fontSize{24.0F};
+
+        /// @brief The font atlas to be used.
+        Asset<FontAtlas> fontAtlas{AnyAsset("ab32bedd-8ea5-49f0-bca6-0b319feb1207")};
+
+        core::gl::VertexArray vertexArray{nullptr};
+
+        size_t vertexCount{0};
```
RiscadoA, line 43: One interesting approach to hide these internal implementation data which is being used in the new audio plugin, is to have a separate, private (header in the `src/` directory) component called `UITextImpl` which contains these private fields.
Then you add observers which automatically add and remove that component when the main component is removed, and you no longer have the problem of the user being able to mess with them!
I think it is also a good fit here, what do you think?

engine/src/font/atlas/bridge.cpp
```diff
@@ -0,0 +1,52 @@
...
+            miterLimit = std::stod(optPixelRange.value());
+        }
+        std::optional<std::string> optFont = meta->get("font");
+        if (optFont.has_value())
+        {
+            fontAsset = AnyAsset{optFont.value()};
+        }
+        if (fontAsset.isNull())
+        {
+            CUBOS_ERROR("Could not load FontAtlas! Missing font.");
```
RiscadoA, line 36:
```suggestion
            CUBOS_ERROR("Could not load FontAtlas: missing font identifier in .meta");
```

engine/src/font/atlas/bridge.cpp
```diff
@@ -0,0 +1,52 @@
...
+    double pixelRange = 4.0;
+    double miterLimit = 1.0;
+    AnyAsset fontAsset{};
+    {
+        auto meta = assets.readMeta(handle);
+        std::optional<std::string> optMinimumScale = meta->get("minimumScale");
+        if (optMinimumScale.has_value())
+        {
+            minimumScale = std::stod(optMinimumScale.value());
+        }
```
RiscadoA, line 18:
```suggestion
        if (auto optMinimumScale = meta->get("minimumScale"))
        {
            minimumScale = std::stod(*optMinimumScale);
        }
```
Cooler syntax :relieved:

RiscadoA (approved):
Other than what I have pointed out, LGTM!

engine/include/cubos/engine/ui/text/text.hpp
mkuritsu: yes ok seems good to me will implement that change

PR #1379, opened 2024-11-21
feat(engine): added resource to easily configure magic numbers in solver

GCeSilva:
# Description

Added a resource to easily configure magic numbers in solver.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/src/physics/solver/integration/plugin.cpp
+++ b/engine/src/physics/solver/integration/plugin.cpp
@@ -54,13 +54,14 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
         .call([](Query<Velocity&, AngularVelocity&, const Force&, const Torque&, const Mass&, const Inertia&,
                        const Rotation&>
                      query,
-                 const Damping& damping, const FixedDeltaTime& fixedDeltaTime, const Substeps& substeps) {
+                 const Damping& damping, const FixedDeltaTime& fixedDeltaTime, const Substeps& substeps,
+                 const SolverConstants& solverConstants) {
             float subDeltaTime = fixedDeltaTime.value / (float)substeps.value;
 
             for (auto [velocity, angVelocity, force, torque, mass, inertia, rotation] : query)
             {
                 // Linear velocity
-                if (mass.inverseMass <= 0.0F)
+                if (mass.inverseMass <= solverConstants.minInvMass)
                 {
                     continue;
                 }
@@ -74,7 +75,7 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
                 velocity.vec += deltaLinearVelocity;
 
                 // Angular velocity
-                if (inertia.inverseInertia == glm::mat3(0.0F))
+                if (inertia.inverseInertia == solverConstants.minInvInertia)
                 {
                     continue;
                 }
@@ -97,13 +98,14 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
         .call([](Query<AccumulatedCorrection&, Rotation&, const Velocity&, const AngularVelocity&, const Mass&,
                        const Inertia&>
                      query,
-                 const FixedDeltaTime& fixedDeltaTime, const Substeps& substeps) {
+                 const FixedDeltaTime& fixedDeltaTime, const Substeps& substeps,
+                 const SolverConstants& solverConstants) {
             float subDeltaTime = fixedDeltaTime.value / (float)substeps.value;
 
             for (auto [correction, rotation, velocity, angVelocity, mass, inertia] : query)
             {
                 // Position
-                if (mass.inverseMass <= 0.0F)
+                if (mass.inverseMass <= solverConstants.minInvMass)
                 {
                     continue;
                 }
@@ -111,7 +113,7 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
                 correction.position += velocity.vec * subDeltaTime;
 
                 // Rotation
-                if (inertia.inverseInertia == glm::mat3(0.0F))
+                if (inertia.inverseInertia == solverConstants.minInvInertia)
                 {
                     continue;
                 }
@@ -124,10 +126,11 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
 
     cubos.system("finalize position")
         .tagged(physicsFinalizePositionTag)
-        .call([](Query<Position&, AccumulatedCorrection&, const Mass&> query) {
+        .call([](Query<Position&, AccumulatedCorrection&, const Mass&> query,
+                 const SolverConstants& solverConstants) {
             for (auto [position, correction, mass] : query)
             {
-                if (mass.inverseMass <= 0.0F)
+                if (mass.inverseMass <= solverConstants.minInvMass)
                 {
                     continue;
                 }
--- a/engine/src/physics/solver/penetration_constraint/plugin.cpp
+++ b/engine/src/physics/solver/penetration_constraint/plugin.cpp
@@ -34,14 +34,14 @@ static void getPlaneSpace(const glm::vec3& n, glm::vec3& tangent1, glm::vec3& ta
 }
 
 static void getTangents(const glm::vec3& velocity1, const glm::vec3& velocity2, const glm::vec3& n, glm::vec3& tangent1,
-                        glm::vec3& tangent2)
+                        glm::vec3& tangent2, const SolverConstants& solverConstants)
 {
     // Use linear relative velocity for determining tangents.
     glm::vec3 linearVr = velocity2 - velocity1;
 
     tangent1 = linearVr - glm::dot(linearVr, n) * n;
     float tangentLenSq = glm::length2(tangent1);
-    if (tangentLenSq > 1e-6 * 1e-6) /// TODO: check this
+    if (tangentLenSq > solverConstants.minTangentLenSq) /// TODO: check this
     {
         tangent1 = glm::normalize(tangent1);
         tangent2 = glm::cross(n, tangent1);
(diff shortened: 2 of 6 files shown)


engine/src/physics/solver/penetration_constraint/plugin.cpp
```diff
@@ -120,7 +121,7 @@ static void solvePenetrationConstraint(
             }
             else if (useBias)
             {
-                bias = glm::max(constraint.biasCoefficient * separation, -4.0F);
+                bias = glm::max(constraint.biasCoefficient * separation, physicsConstants.biasMax);
```
joaomanita, line 124: I have no  knowledge about this code, but if we're taking the max of two quantities, shouldn't biasMax be called biasMin?

GCeSilva: It might be wrong, I don't really have knowledge about some aspects of the code or its magic numbers context.
I based it off of its "position in the code", I read "bias", I read "max", so I called it "biasMax", but that might be wrong yes.
Tiago might be able to clarify this as I actually need someone who knows the code to check if the names are all right.

fallenatlas: Yeah, it's a bit of a weird case because it is the maximum bias, but the bias happens to be negative so it's treated as the minimum. I think the best would be to maybe keep biasMax but put a comment on the resource saying the bias is negative and so it is the maximum negative value it can have, just so there is no confusion in the future.

engine/src/physics/solver/integration/physics_constants_integration.hpp
```diff
@@ -0,0 +1,24 @@
...
+    /// @brief Resource which allows configuration over magic numbers in the solver plugin.
+    /// @ingroup physics-plugin
+    struct PhysicsConstantsIntegration
+    {
+        CUBOS_REFLECT;
+
+        float cmpInvMass = 0.0F;
+
+        float cmpInvInertia = 0.0F;
+    };
```
fallenatlas, line 22: Since these are both the minimum values, instead of `cmp` we should have `min` in the name, it's a bit clearer.

engine/src/physics/solver/penetration_constraint/physics_constants_pc.hpp
```diff
@@ -0,0 +1,36 @@
+/// @file
+/// @brief Resource @ref cubos::engine::MagiConfig.
```
fallenatlas, line 2: Don't forget to also change the resource name in this line.

engine/src/physics/solver/penetration_constraint/physics_constants_pc.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma once
+
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Resource which allows configuration over magic numbers in the solver plugin.
```
fallenatlas, line 13: We should avoid using magic numbers to describe something haha. Could be like this
```suggestion
    /// @brief Resource which allows configuration over constants in the penetration constraint plugin.
```

engine/src/physics/solver/penetration_constraint/physics_constants_pc.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+        float kFriction = 0.0F;
+
+        float cmpTangentLenSq = 1e-6F * 1e-6F;
+
+        float cmpRestitution = 0.0F;
+
+        float cmpNormalSpeed = -0.01F;
+
+        float cmpNormalImpulse = 0.0F;
```
fallenatlas, line 33: Same thing as before here, they are all minimum values. The kNormal and KFriction should also be specified as min.

fallenatlas (changes requested):
Thanks for working on this! 👍 Overall looks good to me, just have a couple of details that need sorting out.

joaomanita (approved):

engine/src/physics/solver/integration/physics_constants_integration.hpp
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// @brief Resource which allows configuration over constants in the integration plugin.
+    /// @ingroup physics-plugin
+    struct PhysicsConstantsIntegration
+    {
+        CUBOS_REFLECT;
+
+        float minInvMass = 0.0F;
+
+        float minInvInertia = 0.0F;
```
fallenatlas, line 21: Ok, one last thing that I didn't notice last time, the minInvInertia should be the matrix itself. This avoids creating a new matrix every time we do a comparison with it.
```suggestion
        float minInvInertia = glm::mat3(0.0F);
```

engine/src/physics/solver/integration/plugin.cpp
```diff
@@ -74,7 +77,7 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
                 velocity.vec += deltaLinearVelocity;
 
                 // Angular velocity
-                if (inertia.inverseInertia == glm::mat3(0.0F))
+                if (inertia.inverseInertia == glm::mat3(physicsConstants.minInvInertia))
```
fallenatlas, line 80:
```suggestion
                if (inertia.inverseInertia == physicsConstants.minInvInertia)
```

engine/src/physics/solver/integration/plugin.cpp
```diff
@@ -97,21 +100,22 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
...
+                if (mass.inverseMass <= physicsConstants.minInvMass)
                 {
                     continue;
                 }
 
                 correction.position += velocity.vec * subDeltaTime;
 
                 // Rotation
-                if (inertia.inverseInertia == glm::mat3(0.0F))
+                if (inertia.inverseInertia == glm::mat3(physicsConstants.minInvInertia))
```
fallenatlas, line 118:
```suggestion
                if (inertia.inverseInertia == physicsConstants.minInvInertia)
```

fallenatlas (changes requested):

engine/src/physics/solver/penetration_constraint/physics_constants_pc.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma once
+
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Resource which allows configuration over magic numbers in the solver plugin.
```
fallenatlas, line 13: Don't forget to change this line to not have the "magic numbers" too.

engine/src/physics/solver/integration/plugin.cpp
RiscadoA, line 80: I'm not sure what this means, but is this really a constant that can be something other than zero?
Since you're using it only in `==` comparisons, it's not really a 'minimum inverse inertia', right? :thinking: 
But once again, I don't understand the math, so if it makes sense to be like this leave it as is

engine/src/physics/solver/integration/physics_constants_integration.hpp
RiscadoA, line 21: Would be cool to document the meaning of these constants, maybe @fallenatlas can help on that

engine/src/physics/solver/penetration_constraint/physics_constants_pc.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+        float minKFriction = 0.0F;
+
+        float minTangentLenSq = 1e-6F * 1e-6F;
+
+        float minRestitution = 0.0F;
+
+        float minNormalSpeed = -0.01F;
+
+        float minNormalImpulse = 0.0F;
```
RiscadoA, line 33: Same as before, it would be good to have these things documented! Otherwise they won't get picked up by doxygen when generating our doc pages.

engine/src/physics/solver/integration/physics_constants_integration.hpp
RiscadoA, line 1: I just noticed that these resources are defined in the private section of the source code (i.e. `src/` directory).
This makes it impossible for a game dev to configure these things if needed, without recompiling the engine itself.
I would put these in the `include/` dir. Also, do we really need to have two separate resources? :thinking: what's the reasoning behind it?

RiscadoA (approved):
Other than what I pointed out LGTM! Ty for working on this!
If you could open a follow up issue to add a tool for configuring these values at runtime it would be cool :eyes: might be a good first issue for someone interesting in playing with tools, or for one of you in the physics team if you want to be able to properly debug stuff

RiscadoA:
Also, don't forget to update the CHANGELOG.md!

engine/src/physics/solver/integration/physics_constants_integration.hpp
fallenatlas: I also thought about having it together or not, honestly idk what's best, on one hand they are both technically just solver constants but ones are only used in the integrator and the others in the constraint. The biggest player to putting them together would be if they are again used in another constraint but since we don't have any other constraints yet it's hard to know if it's best to keep them all together or not. I don't really see any downsides to putting all them in the same tho.

engine/src/physics/solver/integration/plugin.cpp
fallenatlas: Inertia has always been weird because the idea is that it is supposed to be checked in the same way as the normal mass which is just a float, but it's weird to do that with a matrix. Like do we set the minimum matrix and then do we want to check if the other has all elements smaller? Or just 1 element is enough? I'm not sure and that's why it's still this way. (reminder that it was like this because of the floating point "issues" which I'm not even sure if they're that relevant rn since we never really had any problems until now). If you have any idea for this, we'll take it.

fallenatlas (approved):
LGTM! 👍 Don't forget to update the CHANGELOG

PR #1385, opened 2024-11-23
feat(anti-aliasing): add anti-aliasing by FXAA

kuukitenshi:
# Add anti-aliasing by FXAA

Added anti-aliasing using the FXAA technique.

Before
image

After
image

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/assets/render/tone_mapping.fs
+++ b/engine/assets/render/tone_mapping.fs
@@ -6,11 +6,172 @@ uniform sampler2D hdrTexture;
 uniform float gamma;
 uniform float exposure;
 
+uniform uvec2 screenSize;
+uniform bool fxaaEnabled;
+
+layout(std140) uniform FxaaConfig
+{
+    float edgeThresholdMin;
+    float edgeThresholdMax;
+    float subpixelQuality;
+    int iterations;
+};
+
 layout(location = 0) out vec4 color;
+
+// Convert RGB to luma using the formula: L = 0.299 * R + 0.587 * G + 0.114 * B
+float rgb2luma(vec3 rgb){
+    return sqrt(dot(rgb, vec3(0.299, 0.587, 0.114)));
+}
+
+float quality(int i) {
+    return (i < 5) ? 1.0 : 1.5 + (i - 5) * 0.5; //increase progressively the quality
+}
+
+vec3 fxaa(sampler2D screenTexture, vec2 fragUv, vec2 inverseScreenSize){
+    
+    vec3 colorCenter = texture(screenTexture, fragUv).rgb;
+    float lumaCenter = rgb2luma(colorCenter);
+   
+    // direct neighbours of the current fragment
+    float lumaDown = rgb2luma(textureOffset(screenTexture, fragUv, ivec2(0,-1)).rgb);
+    float lumaUp = rgb2luma(textureOffset(screenTexture, fragUv, ivec2(0,1)).rgb);
+    float lumaLeft = rgb2luma(textureOffset(screenTexture, fragUv, ivec2(-1,0)).rgb);
+    float lumaRight = rgb2luma(textureOffset(screenTexture, fragUv, ivec2(1,0)).rgb);
+
+    float lumaMin = min(lumaCenter,min(min(lumaDown,lumaUp),min(lumaLeft,lumaRight)));
+    float lumaMax = max(lumaCenter,max(max(lumaDown,lumaUp),max(lumaLeft,lumaRight)));
+    float lumaRange = lumaMax - lumaMin;
+    
+    // when luma variation is lower that a threshold (or if we are in a dark area), we aren't on an edge, so don't do AA
+    if(lumaRange < max(edgeThresholdMin,lumaMax*edgeThresholdMax)){
+        return colorCenter;
+    }
+
+    // corners of the current fragment
+    float lumaDownLeft = rgb2luma(textureOffset(screenTexture, fragUv, ivec2(-1,-1)).rgb);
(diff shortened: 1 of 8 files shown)


engine/assets/render/tone_mapping.fs
```diff
@@ -5,12 +5,162 @@ in vec2 fragUv;
 uniform sampler2D hdrTexture;
 uniform float gamma;
 uniform float exposure;
+uniform uvec2 screenSize;
+
+const float EDGE_THRESHOLD_MIN = 0.0312;
+const float EDGE_THRESHOLD_MAX = 0.125;
+const float SUBPIXEL_QUALITY = 0.75;
+const int ITERATIONS = 12;
```
RiscadoA, line 13: If these parameters (or some of them) are things you need to tinker with to get right, I think you should pass them as an uniform, so that you can have these values come from a resource or component on the CPU side!

Then we could have a tool to configure that, which would be cool :eyes:
In this case, ig it would go into the `ToneMapping` component? Since there are so many uniforms already, I would put them all in a single uniform block, as seen in other shaders

One other extra that I think would be really cool would be to have a separate `FXAA` component, managed by the tone mapping plugin.
If the component is present, then `FXAA` is enabled, and the settings are fetched from it. Otherwise, we just have plain old tonemapping

RiscadoA (approved):
Tyvm for working on this! Always cool to see graphics improvements :eyes:
LGTM, I just added some suggestions which I think would be nice to have and make this more flexible.

engine/assets/render/tone_mapping.fs
```diff
@@ -6,11 +6,165 @@ uniform sampler2D hdrTexture;
...
 uniform float exposure;
 
+uniform uvec2 screenSize;
+
+layout(std140) uniform FxaaConfig
+{
+    float edgeThresholdMin;
+    float edgeThresholdMax;
+    float subpixelQuality;
+    int iterarions;
```
tomas7770, line 16:
```suggestion
    int iterations;
```

engine/src/render/tone_mapping/plugin.cpp
```diff
@@ -29,17 +40,25 @@ namespace
...
             gammaBP = pipeline->getBindingPoint("gamma");
             exposureBP = pipeline->getBindingPoint("exposure");
-            CUBOS_ASSERT(hdrBP && gammaBP && exposureBP, "hdrTexture, gamma and exposure binding points must exist");
-
+            screenSizeBP = pipeline->getBindingPoint("screenSize");
+            fxaaConfigBP = pipeline->getBindingPoint("FxaaConfig");
+            CUBOS_ASSERT(hdrBP && gammaBP && exposureBP && screenSizeBP && fxaaConfigBP,
+                         "hdrTexture, gamma, exposure, screeSize and fxaaConfig binding points must exist");
             generateScreenQuad(renderDevice, pipeline, screenQuad);
+            fxaaConfigCB = renderDevice.createConstantBuffer(sizeof(FxaaConfig), nullptr, Usage::Default);
```
tomas7770, line 61: When I tested on my machine, even though it worked, it complained about using `glBufferSubData` to update a static buffer. Changing it to dynamic fixed the error.
```suggestion
            fxaaConfigCB = renderDevice.createConstantBuffer(sizeof(FxaaConfig), nullptr, Usage::Dynamic);
```

engine/src/render/tone_mapping/plugin.cpp
```diff
@@ -74,11 +94,21 @@ void cubos::engine::toneMappingPlugin(Cubos& cubos)
...
             {
+                if (fxaa.contains())
+                {
+                    FxaaConfig fxaaConfig{};
+                    fxaaConfig.edgeThresholdMin = fxaa->edgeThresholdMin;
+                    fxaaConfig.edgeThresholdMax = fxaa->edgeThresholdMax;
+                    fxaaConfig.subpixelQuality = fxaa->subpixelQuality;
+                    fxaaConfig.iterations = fxaa->iterations;
+                    state.fxaaConfigCB->fill(&fxaaConfig, sizeof(FxaaConfig));
+                }
```
tomas7770, line 111: I may be missing something, but this seems to leave the buffer completely uninitialized if the FXAA component isn't present, which in that case could lead to unexpected behavior.

tomas7770 (commented):
Mostly LGTM! I left some more comments about potential things to improve/fix.

engine/src/render/tone_mapping/plugin.cpp
```diff
@@ -29,17 +40,28 @@ namespace
...
             hdrBP = pipeline->getBindingPoint("hdrTexture");
             gammaBP = pipeline->getBindingPoint("gamma");
             exposureBP = pipeline->getBindingPoint("exposure");
-            CUBOS_ASSERT(hdrBP && gammaBP && exposureBP, "hdrTexture, gamma and exposure binding points must exist");
-
+            screenSizeBP = pipeline->getBindingPoint("screenSize");
+            fxaaConfigBP = pipeline->getBindingPoint("FxaaConfig");
+            fxaaEnabledBP = pipeline->getBindingPoint("fxaaEnabled");
+            CUBOS_ASSERT(
+                hdrBP && gammaBP && exposureBP && screenSizeBP && fxaaConfigBP,
```
tomas7770, line 61:
```suggestion
                hdrBP && gammaBP && exposureBP && screenSizeBP && fxaaConfigBP && fxaaEnabledBP,
```

tomas7770 (approved):
Just one final detail you forgot 😌

engine/src/render/tone_mapping/plugin.cpp
kuukitenshi: I'm blind ty XD

Docas95 (approved):

PR #1425, opened 2025-01-04
feat(data): Allow wrapped fields to be read as normal object with a single field

SrGesus:
# Description

Previously in a scene if you had an object with a single fields you would be forced to type it as a value like so:
```cpp
"object": "value1"
```
But that can be inconvenient so now you can also type it like any other object:
```cpp
"object": {
    "field1": "value1"
}
```

If we want to preserve the old behavior for a given type, we use the new WrapperTrait instead of a FieldsTrait.

## Checklist
- [x] Self-review changes.
- [x] Ensure test coverage.


RiscadoA:
This is an important change and since it was pretty much finished I gave it the final touch ups it needed so that we can include it in this version @SrGesus.

RiscadoA (approved):
I reviewed all of the code, fixed some bugs, and added what was missing :relieved:

PR #1433, opened 2025-01-26
Add conditional shader compilation

tomas7770:
# Description

Adds the ability to configure compile-time shader parameters using `#define` macros. This allows changing e.g. the number of CSM splits in deferred shading, textures drawn to in the G-buffer rasterizer (to disable/enable the render picker).

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/render/shader/shader.hpp
+++ b/engine/include/cubos/engine/render/shader/shader.hpp
@@ -4,10 +4,13 @@
 
 #pragma once
 
+#include <utility>
+
 #include <cubos/core/gl/render_device.hpp>
 #include <cubos/core/reflection/reflect.hpp>
 
 #include <cubos/engine/api.hpp>
+#include <cubos/engine/render/shader/shader_builder.hpp>
 
 namespace cubos::engine
 {
@@ -21,15 +24,25 @@ namespace cubos::engine
         ~Shader() = default;
 
         /// @brief Constructs a shader from code.
-        /// @param shaderStage Shader stage created from GLSL code.
-        Shader(cubos::core::gl::ShaderStage shaderStage)
-            : mShaderStage(std::move(shaderStage)){};
+        /// @param stage Shader stage to create.
+        /// @param renderDevice Render device used to create the shader.
+        /// @param contents Shader source code.
+        Shader(const cubos::core::gl::Stage stage, cubos::core::gl::RenderDevice& renderDevice, std::string contents)
+            : mStage(stage)
+            , mRenderDevice(renderDevice)
+            , mContents(std::move(contents)){};
 
         /// @brief Returns the asset's shader stage.
         /// @return Shader stage.
         cubos::core::gl::ShaderStage shaderStage() const;
 
+        /// @brief Returns a shader builder to configure compile-time shader parameters.
+        /// @return Shader builder.
+        ShaderBuilder builder() const;
+
     private:
-        cubos::core::gl::ShaderStage mShaderStage;
+        const cubos::core::gl::Stage mStage;          ///< Shader stage type to create.
+        cubos::core::gl::RenderDevice& mRenderDevice; ///< Render device used to create the shader.
+        const std::string mContents;                  ///< Shader source code.
     };
 } // namespace cubos::engine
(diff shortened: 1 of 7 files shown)


Docas95 (approved):

engine/src/render/shader/shader.cpp
```diff
@@ -19,3 +19,8 @@ cubos::core::gl::ShaderStage Shader::shaderStage() const
 {
     return mShaderStage;
 }
+
+const ShaderBuilder& Shader::builder() const
+{
+    return mBuilder;
+}
```
mkuritsu, line 26:
```suggestion
ShaderBuilder Shader::builder() const
{
    return mBuilder; // this or maybe call the constructor directly an avoid storing builder as member variable
}
```
I think maybe would be better to make the functions in ShaderBuilder mutate the class and here return a new instance of the ShaderBuilder

engine/include/cubos/engine/render/shader/shader.hpp
```diff
@@ -22,14 +23,24 @@ namespace cubos::engine
...
         /// @param shaderStage Shader stage created from GLSL code.
-        Shader(cubos::core::gl::ShaderStage shaderStage)
-            : mShaderStage(std::move(shaderStage)){};
+        /// @param renderDevice Render device used to create the shader.
+        /// @param stage Shader stage to create.
+        /// @param contents Shader source code.
+        Shader(cubos::core::gl::ShaderStage shaderStage, cubos::core::gl::RenderDevice& renderDevice,
+               cubos::core::gl::Stage stage, const std::string& contents)
+            : mShaderStage(std::move(shaderStage))
+            , mBuilder(renderDevice, stage, contents){};
```
mkuritsu, line 32:
```suggestion
        Shader(cubos::core::gl::ShaderStage shaderStage, cubos::core::gl::RenderDevice& renderDevice,
               const std::string& contents)
            : mShaderStage(std::move(shaderStage))
            , mBuilder(renderDevice, mShaderStage->getType(), contents){};
```

engine/src/render/shader/shader_builder.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+ShaderBuilder ShaderBuilder::with(const std::string& defineName) const
+{
+    return ShaderBuilder(*this, "#define " + defineName + "\n" + mContents);
+}
+
+ShaderBuilder ShaderBuilder::with(const std::string& defineName, const std::string& defineValue) const
+{
+    return ShaderBuilder(*this, "#define " + defineName + " " + defineValue + "\n" + mContents);
+}
```
mkuritsu, line 19: If you end up doing the change above these function would not need to create a new instance of ShaderBuilder an copy the string everytime `.with` is called, 

With this maybe u could also change ShaderBuilder to store a map of the defines and values and only add the `#define` to the string in the build function

mkuritsu (changes requested):
In general LGTM just pointing some changes that could be made/discussed

engine/src/render/shader/shader.cpp
tomas7770: Makes sense, indeed after some quick research I realized builders following this kind of pattern are usually mutable.

mkuritsu (approved):

engine/include/cubos/engine/render/shader/shader.hpp
```diff
@@ -22,14 +23,25 @@ namespace cubos::engine
...
         /// @brief Returns the asset's shader stage.
         /// @return Shader stage.
         cubos::core::gl::ShaderStage shaderStage() const;
 
+        /// @brief Returns a shader builder to configure compile-time shader parameters.
+        /// @return Shader builder.
+        ShaderBuilder builder() const;
+
     private:
         cubos::core::gl::ShaderStage mShaderStage;
```
RiscadoA, line 43: Do we still need this? Can't we just change shaderStage to return `builder().build()`?

RiscadoA (approved):
Ty for working on this, and sorry for the delay :pensive: LGTM, I just don't get yet why would we want to keep the 'normal' shader stage in the asset. Any reason for it? If you want to merge this before the hotfixes PR, make sure to cherry pick the CI fix I pushed (last commit on that PR)

engine/include/cubos/engine/render/shader/shader.hpp
tomas7770: > Can't we just change shaderStage to return builder().build()?
Yeah that works.

PR #1457, opened 2025-02-06
feat(engine): allow multiple contacts between two entities

joaomanita:
# Description

Remove everything from CollidingWith except entity, add a vector of manifolds.
ContactManifold is no longer a relation.
Same thing for PenetrationConstraint, add a PenetrationtConstraints relation that holds a vector of PenetrationConstraint.

Change narrow phase systems to buildup the vector of manifolds in a collidingWith relation.
Change peneration constraint plugin to create the vector of PenetrationConstraint for each manifold.
Change penetraton constraint solver to buildup the effects of all constraints in PenetrationConstraints

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/include/cubos/engine/collisions/colliding_with.hpp
+++ b/engine/include/cubos/engine/collisions/colliding_with.hpp
@@ -21,7 +21,7 @@ namespace cubos::engine
     {
         CUBOS_REFLECT;
 
-        cubos::core::ecs::Entity entity; ///< Entity to which the normal is relative to.
+        cubos::core::ecs::Entity entity; ///< Entity to which the normals in the manifolds are relative to..
 
         std::vector<ContactManifold> manifolds; ///< All contact interfaces between the entities
     };
--- a/engine/include/cubos/engine/collisions/colliding_with.hpp
+++ b/engine/include/cubos/engine/collisions/colliding_with.hpp
@@ -7,9 +7,11 @@
 #include <glm/vec3.hpp>
 
 #include <cubos/core/ecs/entity/entity.hpp>
+#include <cubos/core/reflection/external/vector.hpp>
 #include <cubos/core/reflection/reflect.hpp>
 
 #include <cubos/engine/api.hpp>
+#include <cubos/engine/collisions/contact_manifold.hpp>
 
 namespace cubos::engine
 {
@@ -20,10 +22,7 @@ namespace cubos::engine
         CUBOS_REFLECT;
 
         cubos::core::ecs::Entity entity; ///< Entity to which the normal is relative to.
-        glm::vec3 entity1OriginalPosition;
-        glm::vec3 entity2OriginalPosition;
-        float penetration;  ///< Penetration depth of the collision.
-        glm::vec3 position; ///< Position of contact on the surface of the entity.
-        glm::vec3 normal;   ///< Normal of contact on the surface of the entity.
+
+        std::vector<ContactManifold> manifolds; ///< All contact interfaces between the entities
     };
 } // namespace cubos::engine
--- a/engine/src/collisions/narrow_phase/plugin.cpp
+++ b/engine/src/collisions/narrow_phase/plugin.cpp
@@ -3,6 +3,7 @@
 
 #include <glm/glm.hpp>
 
+#include <cubos/core/ecs/entity/entity.hpp>
 #include <cubos/core/geom/intersections.hpp>
 #include <cubos/core/geom/plane.hpp>
 #include <cubos/core/geom/utils.hpp>
@@ -23,6 +24,7 @@ CUBOS_DEFINE_TAG(cubos::engine::collisionsNarrowCleanTag);
 CUBOS_DEFINE_TAG(cubos::engine::collisionsNarrowTag);
 CUBOS_DEFINE_TAG(cubos::engine::collisionsManifoldTag);
 
+using cubos::core::ecs::Entity;
 using cubos::core::geom::PolygonalFeature;
 using cubos::engine::CollidingWith;
 using cubos::engine::ContactFeatureId;
@@ -33,8 +35,7 @@ std::vector<ContactPointData> computeContactPoints(const cubos::core::geom::Box*
                                                    const LocalToWorld* matchedLocalToWorld1,
                                                    const cubos::core::geom::Box* matchedShape2,
                                                    const LocalToWorld* matchedLocalToWorld2,
-                                                   cubos::core::geom::Intersection& intersectionInfo,
-                                                   const CollidingWith& collidingWith)
+                                                   cubos::core::geom::Intersection& intersectionInfo, const Entity ent)
 {
     // Calculate incident and reference face
     PolygonalFeature face1;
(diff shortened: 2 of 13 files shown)


fallenatlas (changes requested):
Thank you for working on this 🙏 Overall, looks pretty good to me, I just have a few comments, mostly regarding contact caching in this new way of doing things, and a few optimizations.

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -363,131 +378,88 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
-                            points.insert(points.end(), newPoints.begin(), newPoints.end());
+                            const LocalToWorld* matchedLocalToWorld1 = &shiftedLocalToWorld1;
+                            const LocalToWorld* matchedLocalToWorld2 = &shiftedLocalToWorld2;
+
+                            if (ent1 != collidingWith.entity)
+                            {
+                                std::swap(matchedShape1, matchedShape2);
+                                std::swap(matchedLocalToWorld1, matchedLocalToWorld2);
+                                intersectionInfo.normal *= -1;
+                            }
```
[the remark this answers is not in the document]
joaomanita: Is swapping needed? If match already ensures that ent1 from potentiallyCollidingWith is the same as ent1 from collidingWith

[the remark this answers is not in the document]
fallenatlas: That was something that I thought we could do too but if you see, we don't do the match with the collidingWith.entity because we don't know if it exists yet, so we still need to swap them afterwards. One alternative would be to do a match again after already knowing we have the collidingWith, but I don't think we gain much from that.

engine/include/cubos/engine/collisions/colliding_with.hpp
```diff
@@ -20,10 +22,7 @@ namespace cubos::engine
         CUBOS_REFLECT;
 
         cubos::core::ecs::Entity entity; ///< Entity to which the normal is relative to.
```
RiscadoA, line 24:
```suggestion
        cubos::core::ecs::Entity entity; ///< Entity to which the normals in the manifolds are relative to.
```

engine/src/physics/constraints/penetration_constraint.hpp
```diff
@@ -62,4 +61,15 @@ namespace cubos::engine
...
         float massCoefficient;
     };
+
+    /// @brief Holds multiple PenetrationConstraints
+    /// @ingroup physics-solver-plugin
+    struct PenetrationConstraints
+    {
+        CUBOS_REFLECT;
+
+        cubos::core::ecs::Entity entity; ///< Entity to which the normal is relative to.
```
RiscadoA, line 71:
```suggestion
        cubos::core::ecs::Entity entity; ///< Entity to which the normals are relative to.
```

engine/src/physics/constraints/penetration_constraint.hpp
```diff
@@ -62,4 +61,15 @@ namespace cubos::engine
         float impulseCoefficient;
         float massCoefficient;
     };
+
+    /// @brief Holds multiple PenetrationConstraints
```
RiscadoA, line 65:
```suggestion
    /// @brief Holds multiple PenetrationConstraints.
    ///
    /// Necessary due to the possibility of two entities having more than once collision contact.
    ///
```

RiscadoA (approved):
Other than what I pointed out, LGTM! Ty for working on this, will be really nice to have this working :eyes:

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -224,124 +246,144 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
+                        {
+                            continue;
+                        }
+
+                        auto points = computeContactPoints(&boxShape.box, &localToWorld1, &box.box,
+                                                           &shiftedLocalToWorld, intersectionInfo, ent1);
+
+                        newManifolds.push_back(ContactManifold{
+                            .normal = intersectionInfo.normal, .points = points, .boxId1 = 1, .boxId2 = box.boxId});
+                        break;
```
fallenatlas, line 354: We don't need this `break` here.

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -224,124 +246,144 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
+                        anyIntersects = true;
+                    }
 
-                        // Create a translation matrix for the shift
-                        glm::mat4 shiftMatrix = glm::translate(glm::mat4(1.0F), -box.shift);
+                    // If penetration not bigger than 0 continue
+                    if (intersectionInfo.penetration < 0)
+                    {
+                        continue;
+                    }
```
fallenatlas, line 303: Same comment as before regarding the penetration and intersects.

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -363,131 +405,100 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
-                // If penetration not bigger than 0 continue
-                if (collidingWith.penetration < 0)
-                {
-                    continue;
-                }
+                        // If penetration not bigger than 0 continue
+                        if (intersectionInfo.penetration < 0)
+                        {
+                            continue;
+                        }
```
fallenatlas, line 422: Here too.

engine/src/collisions/narrow_phase/plugin.cpp
```diff
@@ -224,124 +246,144 @@ void cubos::engine::narrowPhaseCollisionsPlugin(Cubos& cubos)
...
+
+                        bool existed = false;
+                        for (auto& manifold : collidingWith.manifolds)
+                        {
+                            if (manifold.boxId2 == box.boxId)
+                            {
+                                manifold.normal = intersectionInfo.normal;
+                                matchContactPoints(points, manifold.points);
+                                manifold.points = points;
+                                existed = true;
```
fallenatlas, line 332: Small thing but you can also add a `break` here so we stop as soon as we find it.

fallenatlas (changes requested):
Just a couple more things to point out. Also, since I already merged the warm-starting PR, you can rebase and the new things to this, it should be just a couple changes in the penetration constraint plugin.

engine/src/collisions/narrow_phase/plugin.cpp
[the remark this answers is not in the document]
joaomanita: Ok, but about this issue, I don't understand what you mean here because matchedShape, matchedLocalToWorld and the normal depend on what iteration of boxes we're on. At most I could create a variable outside the loops for ent1 != collidingWith.entity.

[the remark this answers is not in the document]
fallenatlas: What I meant was that the matchedShape would be the voxel shape, not the boxes inside the shape, I forgot to say that. But leave it as it is now, I think its ok.

fallenatlas (approved):
Alright, LGTM! Good work 👍

GCeSilva (approved):
LGTM!

PR #1460, opened 2025-02-10
1437 change inputpressed axis etc to use stdstring_view

kuukitenshi:
# Description

Change the name parameter type in input methods like ::pressed, ::axis, etc., to use std::string_view for more readable usage

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.

### DIFF
--- a/engine/src/input/input.cpp
+++ b/engine/src/input/input.cpp
@@ -216,7 +216,7 @@ int Input::gamepadCount() const
     return static_cast<int>(mGamepadStates.size());
 }
 
-bool Input::pressed(const char* actionName, int player) const
+bool Input::pressed(std::string_view actionName, int player) const
 {
     auto pIt = mPlayerBindings.find(player);
     if (pIt == mPlayerBindings.end())
@@ -228,14 +228,14 @@ bool Input::pressed(const char* actionName, int player) const
     auto aIt = pIt->second.actions().find(actionName);
     if (aIt == pIt->second.actions().end())
     {
-        CUBOS_WARN("Action {} is not bound to any input for player {}", actionName, player);
+        CUBOS_WARN("Action {} is not bound to any input for player {}", std::string(actionName), player);
         return false;
     }
 
     return aIt->second.pressed();
 }
 
-bool Input::justPressed(const char* actionName, int player) const
+bool Input::justPressed(std::string_view actionName, int player) const
 {
     auto pIt = mPlayerBindings.find(player);
     if (pIt == mPlayerBindings.end())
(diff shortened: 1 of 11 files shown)


RiscadoA (approved):
LGTM! Thanks for working on this :pray:

diogomsmiranda (approved):
LGTM!!

engine/src/input/input.cpp
```diff
@@ -262,17 +262,17 @@ bool Input::justReleased(const char* actionName, int player) const
         return false;
     }
 
-    auto aIt = pIt->second.actions().find(actionName);
+    auto aIt = pIt->second.actions().find(std::string(actionName));
```
mkuritsu, line 265: I don't like the need to construct an std::string everytime this function is called to do a lookup, maybe this feature of C++20 could be useful to avoid this: https://www.cppstories.com/2021/heterogeneous-access-cpp20/

mkuritsu (changes requested):
Just some change that can be looked at, everything else LGTM

engine/include/cubos/engine/input/transparent_equal.hpp
RiscadoA, line 1: I would make these part of another moduel, since they're just general utilities. Maybe 'data'?

engine/include/cubos/engine/input/transparent_hash.hpp
RiscadoA, line 1: Same as above

RiscadoA (approved):
Ty for working on this! Other than what I pointed out, LGTM!

mkuritsu (approved):
LGTM

engine/CMakeLists.txt
```diff
@@ -116,6 +116,9 @@ set(CUBOS_ENGINE_SOURCE
 	"src/input/action.cpp"
 	"src/input/axis.cpp"
 
+    "src/data/transparent_hash.cpp"
+    "src/data/transparent_equal.cpp"
```
RiscadoA, line 120: I think this should belong to the core library, as it is just a utility. `core::data` is a good fit!
Also, their name should probably be something like `StringTransparentHash` and `StringTransparentEqual` as they're not really 'general', they only apply to strings

engine/include/cubos/engine/data/transparent_equal.hpp
```diff
@@ -0,0 +1,56 @@
...
+    ///
+    /// This class is used to allow heterogeneous lookup in unordered maps and sets.
+    ///
+    /// @ingroup input-plugin
+    class CUBOS_ENGINE_API DataTransparentEqual final
+    {
+    public:
+        CUBOS_REFLECT;
+        
+        using IsTransparent = void; // Tag to allow heterogeneous lookup
```
RiscadoA, line 24: Will only work if you use `is_transparent`

PR #1461, opened 2025-02-13
Add collision masks and layers

fallenatlas:
# Description

Move collision layers and masks to their own components and create `CollisionGroup` component. Check out the issue for more info.
Currently collisions are detected if either of the bodies detects the other (this functionality will be extended in future PRs).

## Checklist

- [x] Self-review changes.

### DIFF
--- a/engine/src/collisions/broad_phase/collision_group.hpp
+++ b/engine/src/collisions/broad_phase/collision_group.hpp
@@ -1,0 +1,41 @@
+/// @file
+/// @brief Component @ref cubos::engine::CollisionMask.
+/// @ingroup collisions-plugin
+
+#pragma once
+
+#include <cstdint>
+
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component that contains the group of the collider and if it is static.
+    /// @ingroup collisions-plugin
+    struct CollisionGroup
+    {
+        CUBOS_REFLECT;
+
+        /// @brief Group to which the collider belongs.
+        ///
+        /// The groupId is the index of the owner entity of the shape.
+        ///
+        /// Internally, it is used to identity whether the shapes should collide or if they constitute a bigger shape,
+        /// as is the case when being a sub-shape of a MultiCollisionShape. In case they are unique, they will be
+        /// attributed their entity index, otherwise, they will be attributed the index of the entity containing the
+        /// MultiCollisionShape component (their parent).
+        ///
+        /// By default, this is set automatically.
+        uint32_t groupId;
+
+        /// @brief Identities if a collider is static.
+        ///
+        /// Internally, it works as a special groupId. It is used to check whether two bodies should collide. For
+        /// optimization, two static bodies do not collide even when they are intersecting.
+        ///
+        /// By default, this is set automatically.
+        bool isStatic;
+    };
+} // namespace cubos::engine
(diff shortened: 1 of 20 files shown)


joaomanita (approved):

engine/src/collisions/interface/collision_layers.cpp
```diff
@@ -0,0 +1,31 @@
...
+    return core::ecs::TypeBuilder<CollisionLayers>("cubos::engine::CollisionLayers")
+        .withField("value", &CollisionLayers::value)
+        .build();
+}
+
+void cubos::engine::CollisionLayers::setLayerValue(unsigned int layerNumber, bool newValue)
+{
+    if (layerNumber > 31)
+    {
+        return;
```
RiscadoA, line 17: I would either add an assert here or log an error. This kind of thing shouldn't fail silently, I think

engine/src/collisions/interface/collision_layers.cpp
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    value = (value & ~(1 << layerNumber)) | (static_cast<uint32_t>(newValue) << layerNumber);
+}
+
+bool cubos::engine::CollisionLayers::getLayerValue(unsigned int layerNumber) const
+{
+    if (layerNumber > 31)
+    {
+        return false;
```
RiscadoA, line 27: Same here

engine/src/collisions/interface/collision_mask.cpp
```diff
@@ -0,0 +1,31 @@
...
+    return core::ecs::TypeBuilder<CollisionMask>("cubos::engine::CollisionMask")
+        .withField("value", &CollisionMask::value)
+        .build();
+}
+
+void cubos::engine::CollisionMask::setLayerValue(unsigned int layerNumber, bool newValue)
+{
+    if (layerNumber > 31)
+    {
+        return;
```
RiscadoA, line 17: Same as before

engine/src/collisions/interface/collision_mask.cpp
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    value = (value & ~(1 << layerNumber)) | (static_cast<uint32_t>(newValue) << layerNumber);
+}
+
+bool cubos::engine::CollisionMask::getLayerValue(unsigned int layerNumber) const
+{
+    if (layerNumber > 31)
+    {
+        return false;
```
RiscadoA, line 27: Same as before

engine/src/collisions/broad_phase/collision_group.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <cstdint>
+
+#include <cubos/core/reflection/reflect.hpp>
+
+#include <cubos/engine/api.hpp>
+
+namespace cubos::engine
+{
+    /// @brief Component that contains the group of the collider and if it is static.
```
RiscadoA, line 15: A bit confused about this, this isn't being used yet right? Did you include it here by mistake? :thinking:

RiscadoA (approved):
Ty for working on this, looks great :relieved: added some nitpicks

engine/src/collisions/broad_phase/collision_group.hpp
fallenatlas: It was because I changed the order in which I was going to do the PRs. I planned on doing this after the simplify physics issues but then realized I would need this first so I did it first, and this was included. And yeah, it's still not used, it will only be used when we have a static body and for supporting multiple shapes for one object. Would you prefer if I only include it when I do those?

RiscadoA: It's fine, you can leave it at is

GCeSilva (approved):
LGTM!

PR #1469, opened 2025-03-06
fix(input): allow changing mouse state through input plugin

mcanais:
# Description

Currently, to implement, for example, a FPS camera, to 'hide' the mouse and lock it to the center of the window, you need to access the Window class itself.
To fix this behavior, the update mechanic of the mouse state is now handled through the input plugin.

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.


engine/include/cubos/engine/input/input.hpp
```diff
@@ -123,6 +126,23 @@ namespace cubos::engine
...
+
+        /// @brief Sets the mouse state to be updated.
+        /// @param window Window.
+        void updateMouseState(core::io::Window& window);
+
+        /// @brief Gets the update mouse state.
+        /// @return Whether the mouse state should be updated.
+        bool updateMouseState() const;
+
+
```
RiscadoA, line 145:
```suggestion
```

RiscadoA (changes requested):
LGTM! Thanks for working on this :relieved: don't forget to add an entry to the change log!

RiscadoA:
Btw the commit message should be 'feat' instead of 'fix', you're adding a feature, not fixing a bug!

RiscadoA (approved):

PR #1470, opened 2025-03-10
Add voxel shape inertia tensor calculation

fallenatlas:
# Description

Inertia is essentially hell.

This PR is a bit of a mix:
- Improved documentation of Inertia.
- Added inertia tensor calculation for any voxel shape. It's not any known algorithm, it's one I've done because I couldn't find anything that did what I wanted exactly. But I noted any references I had.
- Also added inertia tensor calculation for a capsule, since I was already deep in this research, it was better than having to re-learn it again later.
- Divided each physics component to their own cpp for the reflection, to be more like all the others.

## Checklist

- [x] Self-review changes.


engine/include/cubos/engine/collisions/shapes/voxel.hpp
```diff
@@ -17,6 +17,8 @@ namespace cubos::engine
 {
     /// @brief Component which adds a collision shape corresponding to a given voxel grid to an entity, used with a @ref
     /// Collider component.
+    ///
+    /// Assumes the VoxelGrid to be centered in the entity space.
```
RiscadoA, line 21: Does this disregard the transform matrix present in the collider component?

RiscadoA (approved):
Sorry for taking so long! LGTM! Very nice that you got stuff well documented :relieved: 
Only confused with one thing, which I pointed out in the comment

engine/include/cubos/engine/collisions/shapes/voxel.hpp
fallenatlas: Currently, only the AABB and raycasts consider that transform matrix (bc I forgot it existed when I was implementing stuff). My next PR, which will separate it into it's own component and make the AABB a private component, will fix this, and we will be able to have cases where the VoxelGrid does not need to be centered, since that transform matrix will change the center of mass (which basically tells us the point around which the body rotates).

RiscadoA: If we get collision groups working, maybe a 'collider transform' won't be needed anyway, as we could just put the collider in a child entity

fallenatlas: If we are ok with having the collisions shapes always being in a child entity, we could avoid having that transform component yeah. But if we go this route I think it's best if we set as standard to never have the CollisionShape in the entity itself and always have it be in the child, even if it's just one.

RiscadoA: Hm I don't like that, that would mean a lot of unnecessary overhead. Would it make it simpler in any way?

fallenatlas: It doesn't make it simpler but the issue has to do with the collision shape lining up with whatever we want to line it up with. I think we basically have 3 choices. First is to not have the CollisionTransform component and have the collision shape be put on the entity like it is now, in this case, the collision shape will always be centered and we can't change that, meaning that, if the voxel grid is misaligned for example, we need to change the offset of the voxel grid. I think this is the simplest to the user and the easiest to implement as well. Second, we have the CollisionTransform to change the offset of the collision shape and AABB in the local space of the entity, the difference in relation to the previous is that it's a bit more flexible bc you can change either the "visual side" or the collision to line them up, and internally it has some more complicated calculations when rotating the entity bc we need to account that we aren't rotating around the origin of the entity. The third option is to have the collision shape be always in a child entity, adds that bit of overhead, which is a bit annoying yeah, but would allow to move the collision shape, in the local space, to wherever we want (using the normal tranform), like the previous option, and we don't need those complicated calculations (this is basically how it will work if we have more than one collision shape, we would just be making it work the same way if we have only one). Basically, there's no perfect option. I think it's mainly a question of, if we have only one collision shape, do we need to move it or it is ok to always be centered? In theory, I think having it always centered is ok because we can move the voxel grid, but I'm not sure. What do you think?

PR #1471, opened 2025-03-11
fix(render): crash when adding shadow casters

GalaxyCrush:
# Description

Verify if a caster has a shadowmap at the cameraEntity entrance.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/src/render/deferred_shading/plugin.cpp
+++ b/engine/src/render/deferred_shading/plugin.cpp
@@ -342,7 +342,16 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
                             perLight.numCascades = numCascades;
                             perScene.directionalLightWithShadowsId = directionalLightIndex;
                             perLight.normalOffsetScale = caster.value().baseSettings.normalOffsetScale;
-                            directionalShadowMap = caster.value().shadowMaps.at(cameraEntity)->cascades;
+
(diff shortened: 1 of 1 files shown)


engine/src/render/deferred_shading/plugin.cpp
```diff
@@ -342,7 +342,17 @@ void cubos::engine::deferredShadingPlugin(Cubos& cubos)
                             perLight.numCascades = numCascades;
                             perScene.directionalLightWithShadowsId = directionalLightIndex;
                             perLight.normalOffsetScale = caster.value().baseSettings.normalOffsetScale;
-                            directionalShadowMap = caster.value().shadowMaps.at(cameraEntity)->cascades;
+
+                            if (!caster.value().shadowMaps.contains(cameraEntity))
+                            {
+                                CUBOS_ERROR("Directional shadow caster does not contain a shadow map for camera {}",
```
tomas7770, line 348: I suggest using a warning instead of an error, since this is going to show up every time a shadow caster is added.

tomas7770 (commented):
The approach LGTM! Make sure to also fix spot and point casters! 😌 In addition, in the commit message, `rendering` is usually used instead of `render`.

tomas7770 (approved):

PR #1489, opened 2025-04-13
fix(rendering): deferred shading plugin on web

RiscadoA:
# Description

TL;DR padding was needed, for some reason it only affects web builds

## Checklist

- [x] Self-review changes.
- [ ] Ensure test coverage.
- [ ] Write new samples.

### DIFF
--- a/engine/assets/render/deferred_shading.fs
+++ b/engine/assets/render/deferred_shading.fs
@@ -75,9 +75,9 @@ layout(std140) uniform PerScene
 
     uint numDirectionalLights;
     uint numPointLights;
-    uint numSpotLights;
-    
+    uint numSpotLights;    
     int directionalLightWithShadowsId; // index of directional light that casts shadows, or -1 if none
+
     int useSSAO;
 };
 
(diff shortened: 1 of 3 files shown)


engine/assets/render/deferred_shading.fs
```diff
@@ -75,10 +75,13 @@ layout(std140) uniform PerScene
 
     uint numDirectionalLights;
     uint numPointLights;
-    uint numSpotLights;
-    
+    uint numSpotLights;    
     int directionalLightWithShadowsId; // index of directional light that casts shadows, or -1 if none
+
     int useSSAO;
+    int padding1;
```
tomas7770, line 82: Does this work without the padding here? Padding at the end of the struct is usually not needed in the shader code.

tomas7770 (approved):
LGTM!

engine/assets/render/deferred_shading.fs
RiscadoA: You're right! Will change it, good catch

PR #1490, opened 2025-04-13
feat: improve scene format

RiscadoA:
# Description

This PR got kind of big.
TL;DR, I replaced the old scene loading code with a new bridge which loads a scene hierarchy, based on the format we discussed on Discord.
To make this easier to use, I also added some utility methods to Commands to easily get the root of spawned scenes/blueprints. Previously, the 'root' entity needed to be named, now, there are overrides for easily accessing it.

## Checklist

- [x] Self-review changes.
- [X] Ensure test coverage.
- [X] Adapted existing samples.
- [X] Add entry to the changelog's unreleased section.


tools/tesseratos/src/tesseratos/scene_editor/plugin.cpp
tomas7770, line 1: Is this supposed to be left all commented out like this?

tomas7770 (approved):
LGTM as far as I can tell!

PR #1492, opened 2025-04-20
add distance constraint

joaomanita:
# Description

Not really sure if it working properly when the rope is attached to a point other than the center of the object.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [X] Write new samples.


fallenatlas (changes requested):
Thanks for working on this! I only have a few comments, mostly regarding the weird bug. Besides that, it works pretty well, both with points in the center and not in the center.

engine/src/physics/solver/distance_constraint/plugin.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    cubos.system("solve distance constraints bias")
+        .tagged(distanceConstraintSolveTag)
+        .tagged(physicsSolveConstraintTag)
+        .call([](Query<Entity, const Mass&, const Inertia&, const Rotation&, AccumulatedCorrection&, Velocity&,
+                       AngularVelocity&, DistanceConstraint&, Entity, const Mass&, const Inertia&, const Rotation&,
+                       AccumulatedCorrection&, Velocity&, AngularVelocity&>
+                     query,
+                 const SolverConstants& solverConstants, const FixedDeltaTime& dt,
+                 const Substeps& substeps) { solveDistanceConstraint(query, solverConstants, dt, substeps, false); });
```
fallenatlas, line 158: The bias should be `true` here instead of false.

engine/src/physics/solver/distance_constraint/plugin.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    cubos.system("solve distance constraints no bias")
+        .tagged(physicsSolveRelaxContactTag)
```
fallenatlas, line 161: The tag needs to be `physicsSolveRelaxConstraintTag`, this one is only for contact constraints due to the order they need to run in.

engine/samples/distance_constraint/main.cpp
```diff
@@ -0,0 +1,507 @@
...
+                              .add(LocalToWorld{})
+                              .add(Position{glm::vec3{0.0F, 0.0F, 5.0F}})
+                              .add(Rotation{})
+                              .add(PhysicsBundle{.mass = 500.0F, .velocity = {0.0F, 0.0F, 0.0F}})
+                              .entity();
+
+                commands.relate(state.a, state.b,
+                                DistanceConstraint{.isRigid = false,
+                                                   .minDistance = 0.0F,
+                                                   .maxDistance = 5.0F,
```
fallenatlas, line 368: I can't mark more than this haha, but consider it does until the end of the system. I think we could have all this code for creating the scenario in one function so we don't have to repeat the code here and on the initial spawn. Would make it easier if we want to change something in the future.

fallenatlas (changes requested):
Everything is working now, and I like how you did it to switch between the scenarios in the sample. Only missing a couple more small things I noticed.

engine/include/cubos/engine/physics/constraints/distance_constraint.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace cubos::engine
+{
+    /// @brief Relation that holds the information for a distance constraint between two physics bodies.
+    /// @ingroup physics-solver-plugin
+    struct DistanceConstraint
+    {
+        CUBOS_REFLECT;
+
+        bool isRigid;               ///< Wether the constraint enforces a fixed distance or a range
```
RiscadoA, line 22:
```suggestion
        bool isRigid;               ///< Whether the constraint enforces a fixed distance or a range
```

RiscadoA (approved):
Ty for working on this! Didnt review it in depth, will leave that to @fallenatlas ;)

One thing that would be good to have would be a documentation page for the sample. Check out the page.md file in other samples for examples on how to do it

fallenatlas (approved):
Ship it 🚀

PR #1493, opened 2025-04-20
feat(rendering): change pivot to the center of the grids

kuukitenshi:
# Description

Change pivot to the center of the grids and update samples.

## Checklist

- [x] Self-review changes.


engine/src/render/mesh/plugin.cpp
```diff
@@ -117,7 +117,8 @@ void cubos::engine::renderMeshPlugin(Cubos& cubos)
                 }
                 else if (entry.firstBucketId != RenderMeshPool::BucketId::Invalid)
                 {
-                    cmds.add(ent, RenderMesh{.firstBucketId = entry.firstBucketId}).remove<LoadRenderVoxels>(ent);
+                    auto assetRead = assets.read(entry.asset);
+                    cmds.add(ent, RenderMesh{.firstBucketId = entry.firstBucketId, .baseOffset = assetRead->size() / 2U}).remove<LoadRenderVoxels>(ent);
```
RiscadoA, line 121: Shouldn't the offset be the symetric/negative of this? From the sample scenes, it looks like it should

RiscadoA (changes requested):

engine/include/cubos/engine/render/mesh/mesh.hpp
```diff
@@ -18,5 +18,11 @@ namespace cubos::engine
 
         /// @brief First bucket in the vertex pool which is part of the mesh.
         RenderMeshPool::BucketId firstBucketId{};
+
+        /// @brief Base offset of the mesh.
+        glm::vec3 baseOffset{};
+
+        /// @brief Bounding box of the mesh.
+        glm::vec3 boundingBox{};
```
RiscadoA, line 26: What does 'bounding box' mean here? If it is half the size of the bounding box, I would rename this to 'halfSize' or something like that. If it is the size, maybe just 'size'.

RiscadoA (approved):
Other than what I and Tomás pointed out, LGTM!

tomas7770 (approved):

PR #1494, opened 2025-04-20
feat(ecs): add relation creation and removal detection to observers

### LINKED ISSUE #1416: Add relation creation and removal detection to observers
# Problem
Currently observers can only detect addition and removal of components, not relations.

# Solution
As we have `.onAdd<Foo>` / `.onRemove<Foo>`, we should also have `.onRelate<ChildOf>` and `.onUnrelate<ChildOf>`.

kuukitenshi:
# Description

Added relation creation and removal detection to observers to allow  ```.onRelate<ChildOf>``` and ```.onUnrelate<ChildOf>```.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.


RiscadoA (changes requested):
These observers must not rely on column ids, similarly to what we discussed regarding the entity destruction PR. This is because relations are stored in a different manner from components. We can discuss this later, ping me on discord when you pick this up

PR #1504, opened 2025-05-01
fix(build): compile error in inertia.cpp when using gcc 15

mkuritsu:
# Description

GCC 15 seems to have broken something in the inertia.cpp code. This change seems to fix the compile error just want to be sure it doesn't modify the expected behaviour.

# Additional Information
[compile error.txt](https://github.com/user-attachments/files/20002752/out.txt)

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.


RiscadoA (approved):

RiscadoA:
LGTM to me! @fallenatlas when you can pls approve this

PR #1505, opened 2025-05-01
Make selection gizmos clickable again and properly handle their depth

RiscadoA:
# Description

- The bug in the issue came from considering `ImGui::GetWindowSize()` to return the window size - it doesn't, it returns the current ImGui window's size. I simply replaced it by the correct call.
- I also noticed that when I clicked, depth wasn't properly taken into account, i.e., occluded gizmos were sometimes selected instead of the ones in front. Turns out we forgot to bind the depth texture on the gizmos framebuffer.

## Checklist

- [x] Self-review changes.


tomas7770 (approved):
LGTM!

PR #1509, opened 2025-05-03
1497 add lua bindings plugin

mkuritsu:
# Description

Add lua bindings plugin to be able to execute lua code.

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] Write new samples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reflection support for C arrays (**@RiscadoA**).
 - Reflection for RenderEnvironment (**@RiscadoA**).
 - Configurable directional shadow range for objects outside the frustum (**@tomas7770**).
+- Barebones lua bindings plugin that can execute standard lua code (#1497, **@mkuritsu**).
 
 ### Changed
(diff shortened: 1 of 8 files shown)


RiscadoA:
I'm still not convinced by the 'script = component' model. I definitely think there should be some mechanism to dispatch Lua code from components, but I don't think the whole Lua scripting landscape should come from this.

I feel like we would be better served by treating Lua as almost equivalent to C++, by having it being loaded before any systems run at all. This would allow us to do any of the stuff we're doing in C++ in Lua, such as defining startup systems, that can run before (or in the middle of) the C++ startup systems, or configure the `Assets` plugin from Lua, etc.

Another thing that worries me about this approach is that forcing script files to be assets would make for a quite cumbersome workflow I think, and makes us deviate a bit from the conventions of Lua. I think a more standard Lua project structure, where you have a directory `scripts/` (separate from the assets directory), in which you import other scripts y `require`ing them, would be easier to get into and to expand.

One thing that we should discuss is how using external Lua dependencies would look like. Would they be placed in the scripts directory too? As a submodule perhaps?

RiscadoA (approved):
LGTM! Ty for working on this!

PR #1544, opened 2025-05-17
Show similar or related components close together on the inspector #1525

### LINKED ISSUE #1524: Allow selecting relation source/target entities in the inspector
# Problem
Currently it is possible to see the source/target entities of relations an entity has in the inspector, but not select them. This is not practical at all, as you must resort to finding them in the world inspector.

# Solution
It should be possible.

### LINKED ISSUE #1525: Show similar or related components close together on the inspector
# Problem
Currently the components appear in a random order in the entity inspector. This makes it very hard to work with.

# Solution
At the very least components should be grouped into categories. E.g., 'Transform' components, 'Collision' components, etc. All other components could be placed in a 'Misc' category.

R-Camacho:
# Description

Created a new reflection trait to be able to categorize types. Then used that to group components into categories on the entity inspector.

## Checklist

- [ ] Self-review changes.
- [X] Ensure test coverage.
- [X] Write new samples.


core/include/cubos/core/ecs/reflection.hpp
```diff
@@ -95,6 +96,17 @@ namespace cubos::core::ecs
             return std::move(*this);
         }
 
+        /// @brief Adds a category to a type.
```
RiscadoA, line 99: Might be a good idea to exemplify why this is useful. E.g., used to show related components next to each other on the UI.

core/tests/reflection/traits/categorizable.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+CUBOS_REFLECT_IMPL(AnotherPhysicsType)
+{
+    return Type::create("AnotherPhysicsType").with(CategorizableTrait{"Physics", 1});
+}
+
+TEST_CASE("reflection::CategorizableTrait")
+{
+    // TODO: maybe useless
```
RiscadoA, line 52: Yeah this is covered by the main test so I don't think it is needed - but I think other similar tests also do this, so ig it would be consistent. Feel free to remove it - or not, it would also be ok. Ig its a sanity test

core/tests/reflection/traits/categorizable.cpp
```diff
@@ -0,0 +1,76 @@
...
+using cubos::core::reflection::Type;
+
+namespace
+{
+    struct SimpleType
+    {
+        CUBOS_REFLECT;
+    };
+
+    // TODO: probably renaming these
```
RiscadoA, line 20: I think these names are fine!

core/tests/reflection/traits/categorizable.cpp
R-Camacho: Ok, I'm keeping it!

PR #1547, opened 2025-05-19
feat(data): handle MaskTrait in JSON (de)serializer

RiscadoA:
# Description

Simply adds handlers for the MaskTrait to the JSON serializer and deserializer.
Mask types are now serialized as lists of flags. E.g., with a mask trait with three flags, 'A', 'B' and 'C', it would be serialized as `[]` or `["A", "B"]` etc.

## Checklist

- [ ] Self-review changes.
- [x] Ensure test coverage.


mcanais (approved):
Looks good to me!

PR #1548, opened 2025-05-19
Handle constant types in the reflection

RiscadoA:
# Description

This PR adds a new `ConstantTrait` to the reflection which identifies types which are constant.
E.g., `const T` is reflected as a type with a ConstantTrait, which then points to the type `T`.

I've also added a handler for ConstantTrait to the ImGui inspector, and an example to the imgui sample, so that constant types are always shown as read-only.
image

## Checklist

- [x] Self-review changes.
- [x] Ensure test coverage.
- [x] ~~Write new samples.~~ updated ImGui sample


mkuritsu (approved):
LGTM

Issue #1542, opened 2025-05-16
Use global rotation on physics solving.


fallenatlas:
# Problem
Currently solving uses `Rotation` which contains the rotation of the entity relative to it's parent. This is not an issue until the parent entity possesses a rotation of it's own, leading to weird behaviour.

# Solution
Replace the current rotation with the global rotation of the entity on the solver and on the integrator.

PR #1563, opened 2025-06-21
feat(physics): Interpolation

### LINKED ISSUE #1459: Fix localToWorld update and smooth physics movement.
# Problem
Currently we only do the localToWorld update once per frame, however, when the game is below 30 fps, we run the systems tagged with the fixed step multiple times in one frame. This is particularly troublesome since the Collisions need the updated localToWorld to compute if objects are colliding and their contact points. 

With the game the game itself also running at a higher frame rate, the movement with physics looks very "low fps", due to the position only being updated at 30 fps.

# Proposed solution at the moment

Put `Position`, `Rotation`, etc in a single `Transform` component, to make sure the matrix is always updated.
Ideally we'd have a `Transform`, a `GlobalTransform`, and a `FixedTransform`.

`FixedTransform` would hold the correction calculated by the `solverPlugin` during the fixed step. This means the physics would not directly manipulate the `Transform` but instead the `FixedTransform`.

The `Transform` would be updated by interpolating the `FixedTransform` between fixed updates.
  - Note: When we do the fixed update, we need to always have **exactly** the `FixedTransform` applied to the `Transform`. (so when performing 2 fixed updates in one frame, when doing the second we need to make sure the first is instantly fully applied to the `Transform`)

joaomanita:
things

# Description

Add Interpolated{} to entities to interpolate them.
Example in distance constraint sample.

## Checklist

- [ ] Self-review changes.
- [ ] Ensure test coverage.
- [ x] Write new samples.


RiscadoA (changes requested):
Ty for working on this! Will be very very useful for online multiplayer too :eyes:
I'm not very familiar with the physics code, but I have two doubts regarding this implementation:
- why would `Interpolated` be optional? Shouldn't it always be used? I don't see a reason to use physics without interpolation.
- the whole interpolation mechanism should be independent from the physics plugin. It should be a separate plugin, outside the physics directory. Of course, the physics plugin would depend on it. This would allows us to use it for other stuff such as online multiplayer without necessarily enabling physics.

fallenatlas (changes requested):
Sorry for thanking so long to review and thank you for working on this! 🙏 Overall, I think this works well for what we want, I only have a couple concerns.

fallenatlas:
> Ty for working on this! Will be very very useful for online multiplayer too 👀 I'm not very familiar with the physics code, but I have two doubts regarding this implementation:
> 
> * why would `Interpolated` be optional? Shouldn't it always be used? I don't see a reason to use physics without interpolation.

I think it reduces the necessity of the dev knowing that exists at the start and yeah, ideally we'd always want it I think. I guess it isn't for the same reasons as the solver and stuff also isn't included with the physics which, tbh, I think we should also change those as it doesn't make a lot of sense.

> * the whole interpolation mechanism should be independent from the physics plugin. It should be a separate plugin, outside the physics directory. Of course, the physics plugin would depend on it. This would allows us to use it for other stuff such as online multiplayer without necessarily enabling physics.

Having in account that fixed step is also not part of the physics plugin I think this would make sense yes. I'm worried about how it would work having multiple plugins writing to the same thing and having it work well. But, the biggest problem I see is that, this interpolation is strictly for fixedStep, if we had for example servers running outside of fixedStep I'm not sure if this plugin would be good for that.

engine/src/interpolation/plugin.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <cubos/engine/interpolation/plugin.hpp>
+#include <cubos/engine/physics/plugin.hpp>
+#include <cubos/engine/physics/solver/plugin.hpp>
+#include <cubos/engine/transform/plugin.hpp>
+
+using namespace cubos::engine;
+
+CUBOS_REFLECT_IMPL(cubos::engine::Interpolated)
+{
+    return cubos::core::ecs::TypeBuilder<Interpolated>("cubos::engine::Interpolated").build();
```
RiscadoA, line 17: Any reason to not include reflection for the struct's fields?

engine/src/interpolation/plugin.cpp
```diff
@@ -0,0 +1,72 @@
...
+            for (auto [ent, _, position, rotation, scale] : query)
+            {
+                cmds.add(ent, Interpolated{.previousPosition = position.vec,
+                                           .previousRotation = rotation.quat,
+                                           .previousScale = scale.factor,
+                                           .nextPosition = position.vec,
+                                           .nextRotation = rotation.quat,
+                                           .nextScale = scale.factor});
+            }
+        });
```
RiscadoA, line 42: I think this should be responsibility of the physics plugin. I.e., the physics plugin should depend on interpolation, and add this observer, nto the other way around.

RiscadoA (approved):
LGTM other than what I just pointed out!

engine/src/physics/solver/integration/plugin.cpp
```diff
@@ -118,23 +120,24 @@ void cubos::engine::physicsIntegrationPlugin(Cubos& cubos)
                     continue;
                 }
 
-                rotation.quat =
-                    glm::normalize(rotation.quat +
-                                   (glm::quat(glm::vec4(angVelocity.vec * subDeltaTime * 0.5F, 0.0F)) * rotation.quat));
+                interp.nextRotation = glm::normalize(
+                    interp.nextRotation +
+                    (glm::quat(glm::vec4(angVelocity.vec * subDeltaTime * 0.5F, 0.0F)) * interp.nextRotation));
```
fallenatlas, line 125: This has a big issue that I had forgotten. This can't be this way because we always need the solver to have the most updated rotation, and basically that means that there's no way for you to interpolate it (that I can remember of, if you have any suggestions do say). Idk if you noticed, but I'm pretty sure that's why the collisions sample looks so weird now, it basically can't solve anything because the rotation used is wrong.

This relates to issue #1542 where the plan is to have an independent rotation (and position) for the physics, and then in that case you could get the final rotation from there to use in the interpolation. But rn, we need rotation to always be the most recent.

Solution, I think we can have position interpolation but not rotation for now.

fallenatlas (changes requested):

fallenatlas (changes requested):
Just a couple more things.

Also, I have a doubt. Do we have a guarantee that Position is nextPosition when a fixedStep starts? We need it to make sure that the physics remain independent from the framerate.

engine/include/cubos/engine/interpolation/plugin.hpp
```diff
@@ -0,0 +1,49 @@
...
+        float previousScale; ///< Previous frame scale factor.
+
+        glm::vec3 nextPosition; ///< Next frame position.
+        // glm::quat nextRotation; ///< Next frame rotation.
+        float nextScale; ///< Next frame scale factor.
+    };
+
+    /// @ingroup interpolation-plugin
+    /// @brief Interpolates position, rotation and scale of entities with the @ref Interpolated component in between
+    /// fixed updates.
```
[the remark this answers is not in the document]
joaomanita: What could be different? That there is no rotation interpolation yet?

PR #1570, opened 2025-08-20
Simplify using physics

### LINKED ISSUE #1246: Add `TriggerArea`
Identifies a collider as an area without physical collision that detects collisions with other areas and bodies., and emits signals for these events. 

### Current Design

Components:
- Have a `TriggerArea` component 
  - bool `monitoring` - if true the area is active, otherwise it doesn't detect any collisions
  - bool `monitored` - if true the area is detected by others
- Needs only a `CollisionShape` and a `Collider` to work.
The `TriggerArea` component should be defined in the `CollisionsPlugin`.

Events:
- If a collision object has the `TriggerArea` component the following relations should be created between the area entity and the other entity:
  - OnAreaEntered/Exited, OnBodyEntered/Exited
  - OnArea - for when an object is in the area (equivalent to collidingWith rn, check if this is needed)
- To read an event an entity needs to have an observer for the relation it wants to treat. This observer can technically be anywhere, its up to the developer, however, its probably best to be from the area side since usually its the area that determines the behavior (e.g. have an area handle plugin that implements a certain behavior for an area, and you can have different handle area plugins). Document this.

Other details:
- We want to avoid computing `ContactManifold` for any entity with the `TriggerArea` component.
  - This means these entities only perform the basic collision test (SAT for boxes, ...)

We will want a new sample to demonstrate this.

### Checklist

- [ ] Create `TriggerArea` Component and add check to avoid computing contact manifold
- [ ] Create event relations and create event functionality

### LINKED ISSUE #1426: Simplify using physics in games
# Problem

Currently the Collision and Physics treat all bodies the same way. However, for optimization and ease of use we want to consider different types of Collision Objects, particularly trigger areas and physics bodies, and have an easy way of creating each of these

# Design
We propose this separation:

- [ ] #1246 
- [ ] #1253 

For efficiency we should avoid:
- Computing contact points between areas and every other colliding object.
- Collisions between 2 Physics Bodies that are static.

fallenatlas:
# Description

Create Collider component
  - Supports area behavior and start and end intersection events
  - Holds if a body is active and if is static (for optimizations)
  - Replaces the necessity for collision groups since a collider is always the owner of a collision shape 

Create ColliderBundle: facilitates creation of entities that have collider behavior (intersections)
  - Supports area behavior and start and end intersection events
  - Adds layers and masks
 
Add RigidBodyBundle
Add StaticBodyBundle

Make AccumulatedCorrection private

## Checklist

- [x] Self-review changes.
- [x] Write new samples (collision_areas sample)


joaomanita (approved):
LGTM

JoaoTRoxo (approved):
Looks good to me. The changes are clear and consistent, and I didn't see any issues. Thanks for the thorough implementation.

